PR #3, opened 2015-03-09
Do not open a socket when DogStatsd is instantiated

yannmh:
Forking https://github.com/DataDog/dogstatsd-python/pull/25. 

All credits goes to @benweatherman. Thank you


benweatherman:
Thanks @yannmh!

PR #8, opened 2015-03-23
Don't require `simplejson`

benweatherman:
`simplejson` isn't listed as a dependency, but certain modules were
importing it directly. I switched to using `datadog.compat` since that
does the right thing for py2k and py3k and whether or not you have
`simplejson` installed.

A few other fixes
- Fixed requirements needed for running `python setup.py test`
- Use `pillow` instead of `PIL` for the images needed during testing.
  There are lots of reasons `pillow` is a much better choice, but py3k
  support is a pretty big one.


miketheman:
I am also curious how `test_requires works` with [requirements.txt](https://github.com/DataDog/datadogpy/blob/master/requirements.txt) - I think that having both leads to confusion.

/cc: @yannmh

benweatherman:
You're right, I didn't even notice that. Once we decide on how this project will use `requirements.txt` then I'll update [CONTRIBUTING.md](https://github.com/DataDog/datadogpy/blob/master/CONTRIBUTING.md) accordingly.

yannmh:
Great ! Thanks again @benweatherman.

`requirements.txt` is a leftover: I don't think it makes any sense to keep it. Would you mind removing it please ?

Once, it's done we can merge your contribution and publish a new release :boom:

benweatherman:
OK, I think I've got it all tidied up now. Anything else that I can do to make this ready for primetime?

yannmh:
Great work. Thanks a lot @benweatherman.

I am merging your PR and releasing a new version of `datadogpy`.

PR #30, opened 2015-04-08
Sniff DATADOG_HOST environment variable to determine which API host to use

ddaniels888:


remh:
looks good! Thanks @ddaniels888 !

PR #32, opened 2015-04-09
Add thread safety tests to default test suite

### LINKED ISSUE #29: Allow empty list of tags to be passed in
If an empty list of tags is passed in, a Python error about an unhashable list pops up. This is because the check for tags is faulty here:
https://github.com/DataDog/datadogpy/blob/master/datadog/threadstats/metrics.py#L134

This line should use None if tags fails to bool eval to True, not the passed in value (in this case, []).

yannmh:
And fix #29


LeoCavaille:
The PR title says add thread safety test but it looks like you're deleting a test about thread safety? Is this the right change, I am confused.

yannmh:
There is a discreet additional `__init__.py` file in ` tests/performance/` directory ( which contains thread safety tests): as a result these tests are included in the default test suite ran by `python setup.py test'.

I removed the old test thread safety test you mentionned, as it was redundant.

LeoCavaille:
Looks good, that was a sneaky one indeed.

PR #36, opened 2015-04-14
fix parameter in event api post

### LINKED ISSUE #35: dogshell event post typo
the message doesn't show up when using `dog event post`, i think it's because the api call should have `text=message` instead of `message=message` https://github.com/DataDog/datadogpy/blob/master/datadog/dogshell/event.py#L118

celenechang:
fixes #35


yannmh:
:+1: Thanks @celenechang

Issue #33, opened 2015-04-10
dogwrap doesn't provide any output until wrapped command completes


ddaniels888:
dogwrap buffers the wrapped command's stderr and stdout to a string, and only prints to stdout and stderr upon completion. Without some realtime output, it's hard to use dogwrap on long-running commands.

ddaniels888:
Here's a snippet I've used before to grab output in realtime from subprocess, in case it's useful: https://github.com/mortardata/mortar-luigi/blob/master/mortar/luigi/mortartask.py#L496:L508

You lose differentiation between the wrapped command's stdout and stderr when you do that though, but do get realtime output. It may be possible to still subprocess.PIPE stderr and just not consume it until the process ends.

I think the timeout feature of dogwrap also makes this more comlex, as reading the wrapped command's stdout will block, and hence might require some threading to implement the timeout separately.

yannmh:
Hey @ddaniels888, I agree it would be nice to have a some kind of realtime output.
Thanks for the snippet, I'll see how I can adapt with dogwrap.

alq666:
@yannmh if you don't have a card for this, can you add one so that we can triage?

yannmh:
Sure. 
Just added a card for it.

yannmh:
Fixed in v0.4.0

PR #39, opened 2015-04-23
[stdout stream] Added an option to stream outputs

elafarge:
- By default, dogwrap now forwards the stdout and stderr of the command
  it executes to the terminal launching dogwrap.

- This behaviour can be prevented using the --buffer_outs option
  (dogwrap then acts just as it used to do)

- I also added some comments in the code.

- Finally, I added help messages for the different possible options, as
well as an "USAGE" description and and version number (automatically
enabled --version option for dogwrap).

### DIFF
--- a/datadog/dogshell/wrap.py
+++ b/datadog/dogshell/wrap.py
@@ -19,9 +19,11 @@ dogwrap -n test-job -k $API_KEY --timeout=1 "sleep 3"
 '''
 
 import sys
-import subprocess
 import time
-from optparse import OptionParser
+import optparse
+import threading
+import subprocess
+import pkg_resources as pkg
 
 from datadog import initialize, api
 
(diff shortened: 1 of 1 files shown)


datadog/dogshell/wrap.py
```diff
@@ -86,17 +148,44 @@ def execute(cmd, cmd_timeout, sigterm_timeout, sigkill_timeout,
...
+    parser.add_option('--proc_poll_interval', action='store', type='float', default=0.5,
+                      help="(in seconds). interval at which your command will be polled \
+(default: 500ms)")
+    parser.add_option('--notify_success', action='store', type='string', default='',
+                      help="a message string and @people directives to send notifications in \
+case of success.")
+    parser.add_option('--notify_error', action='store', type='string', default='',
+                      help="a message string and @people directives to send notifications in \
+case of error.")
+    parser.add_option('-f', '--flush_live', action='store_true', dest='flush_live', default=False,
```
yannmh: Should we make it `True` by default ? I think we might want to use this option more often than we won't.

elafarge: Yeah, I was wondering as well but so far the behaviour was to have it set to false and I thought maybe some of our clients use dogwrap in some scripts of their owns. So basically I did that for "backward compatibility" but since dogwrap has been released very recently maybe we can put it to True by default. Shall I do that ?

yannmh: I don't think 'flushing live' versus 'once the command is completed', can introduce backward compatibility.

yannmh:
Did you have a chance to test with Python 2.6 and Python >= 3.3 ?

yannmh:
This fix #33 
Thanks @elafarge !
 
cc @ddaniels888

degemer:
Quick remark about the command line option:
`-f` for most Linux commands mean `--force`, so you might want to choose another letter (`-d` for `--debug` maybe if it's optional?) to avoid confusion.

yannmh:
I can't think about any reason why we would rather have the output provided once the command is completed (versus `real-time`).  Should we just drop it and avoid an extra option ? 
@ddaniels888 any thoughts ?

elafarge:
Yeah, I dont see either to be honest. Maybe we can just enable it by default and add a -b option for (--buffer-outputs) so that we get rid of this letter issue at the same time !

 Also maybe adding a --mute or --silent option to prevent the output from being printed may be a good idea even though it's also almost as easy to just redirect the output of dogwrap to /dev/null. What do you guys think ?

elafarge:
I don't really see any use case for the -b option but it doesn't hurt to keep it "just in case". 
But I think many users would be glad to use dogwrap in shell scripts, and therefore they would probably prefer it to be silent ?

yannmh:
Like the idea of `--silent` and `-b`.

yannmh:
:ok_hand:

ddaniels888:
Thanks for doing this—I'm super happy to have realtime output! :+1: 

Generally, I'm for less flags until we see an actual use case for them.

If we can reliably provide realtime output, I don't see any reason not to. But so long as it isn't much extra code, I guess it's nice to leave the `--buffer` option in case someone was relying on that (for reasons I don't know).

I also don't see much case for `--silent`. I think just doing the standard pipe to `/dev/null` if you want that is easier, and requires extra code to test/maintain.

elafarge:
Ok Doug, I just got rid of the silent option. PR should now be ready to be merged.

PR #42, opened 2015-04-23
[dogwrap] add priority option

### LINKED ISSUE #18: dogwrap events should determine priority based on result
<a href="https://github.com/miketheman"></img></a> **Issue by [miketheman](https://github.com/miketheman)**
_Wednesday Dec 18, 2013 at 13:07 GMT_
_Originally opened as https://github.com/DataDog/dogapi/issues/71_

----

Currently success and failure both come in as normal (default).
dogwrap command should both auto-select the correct one based on output, as well as allow operator to override with commandline flag.

yannmh:
Fix #18

I don' feel confortable about auto-selecting a priority based on the output (i.e. `SUCCESS` -> `low` and `ERROR` -> 'normal'.)
cc @miketheman

### DIFF
--- a/datadog/dogshell/wrap.py
+++ b/datadog/dogshell/wrap.py
@@ -92,6 +92,8 @@ def main():
     parser.add_option('-m', '--submit_mode', action='store', type='choice',
                       default='errors', choices=['errors', 'all'])
     parser.add_option('-t', '--timeout', action='store', type='int', default=60 * 60 * 24)
+    parser.add_option('-p', '--priority', action='store', type='choice', choices=['normal', 'low'],
+                      default='normal', help="The priority of the event (default: 'normal')")
     parser.add_option('--sigterm_timeout', action='store', type='int', default=60 * 2)
     parser.add_option('--sigkill_timeout', action='store', type='int', default=60)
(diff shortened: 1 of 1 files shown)


miketheman:
@yannmh Thanks - I think this covers the operator-set priority just fine.

What I don't get with this is that there's still no method for a cron job failing to be the "right" priority.

An [example from our Chef handler](https://github.com/DataDog/chef-handler-datadog/blob/master/lib/chef/handler/datadog.rb#L104-L111) where we detect whether it was pass/fail, and set priority accordingly.

If I were to add dogwrap to any cron job (to emit events on completion), I'd want the successes to be quiet, and the failures to be visible.

I'm happy to explain this further - but what's making you uncomfortable with auto-detecting priority?

yannmh:
Thanks for your reply.

I brought some extra changes in my PR. 

The reason for my concern is that this will change the event priorities of current jobs sumitting 'success'  type events, without any more notice than the CHANGELOG: can that be considered as a backward incompatibility ?

yannmh:
I don't think it's an issue. Merging.

miketheman:
Thanks!

PR #45, opened 2015-04-24
metric query API support

yannmh:
* Add support for metric query API
* Allows to re-set `_timeout` parameter dynamically

### DIFF
--- a/datadog/api/graphs.py
+++ b/datadog/api/graphs.py
@@ -16,10 +16,10 @@ class Graph(CreateableAPIResource, ActionAPIResource):
         :param metric_query: metric query
         :type metric_query: string query
 
-        :param start: timestamp of the start of the query.
+        :param start: query start timestamp
         :type start: POSIX timestamp
 
-        :param end: timestamp of the end of the query.
+        :param end: query end timestamp
         :type end: POSIX timestamp
 
         :param event_query: a query that will add event bands to the graph
--- a/datadog/api/metrics.py
+++ b/datadog/api/metrics.py
@@ -1,15 +1,18 @@
 import time
 
-from datadog.api.base import SendableAPIResource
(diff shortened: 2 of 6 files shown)


datadog/api/metrics.py
```diff
@@ -57,3 +63,39 @@ def send(cls, *metrics, **single_metric):
...
+
+        >>> api.Metric.query(start=int(time.time()) - 3600, end=int(time.time()),
+                             query='avg:system.cpu.idle{*}')
+        """
+        # Set the right endpoint
+        cls._class_url = cls._METRIC_QUERY_ENDPOINT
+
+        # `from` is a reserved keyword in Python, therefore
+        # `api.Metric.query(from=...)` is not permited
+        # -> map `start` to `from` and `end` to `to`
```
degemer: It's better anyway since the rest of the api seems to expect `start` and `end` parameters when searching for something (I checked `api//events.py`)

degemer:
Really nice @yannmh !
I would have preferred some more :a: - :zzz:, but it doesn't really matter.
Could you just add a real integration test (probably retrieving the metrics sent here https://github.com/DataDog/datadogpy/blob/master/tests/integration/api/test_api.py#L346) ?

datadog/api/graphs.py
```diff
@@ -16,10 +16,10 @@ def create(cls, **params):
         :param metric_query: metric query
         :type metric_query: string query
 
-        :param start: timestamp of the start of the query.
+        :param start: query start timestamp
         :type start: POSIX timestamp
 
-        :param end: timestamp of the end of the query.
+        :param end: query start timestamp
```
remh: typo s/start/end

yannmh:
Added a real integration test. Thanks for your feedback @degemer !

PR #46, opened 2015-05-05
Add `revoke` method to Screenboard

allegrem:
cc @yannmh 

Related to https://github.com/DataDog/documentation/pull/395

### DIFF
--- a/tests/integration/api/test_api.py
+++ b/tests/integration/api/test_api.py
@@ -3,6 +3,7 @@ import datetime
 import os
 import time
 import unittest
+import urllib2
 
 # 3p
 from nose.plugins.attrib import attr
@@ -553,6 +554,15 @@ class TestDatadog(unittest.TestCase):
 
         share_res = dog.Screenboard.share(get_res['id'])
         assert share_res['board_id'] == get_res['id']
+        public_url = share_res['public_url'].replace('8080', '5000')
+
+        response = urllib2.urlopen(public_url)
+        assert response.code == 200
+
+        revoke_res = dog.Screenboard.revoke(get_res['id'])
+        with self.assertRaises(urllib2.HTTPError) as cm:
+            urllib2.urlopen(public_url)
+        assert cm.exception.code == 404
 
         delete_res = dog.Screenboard.delete(update_res['id'])
         assert delete_res['id'] == update_res['id']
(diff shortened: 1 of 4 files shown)


yannmh:
Looks great. Thanks @allegrem

allegrem:
I will add some tests for this PR as soon as possible

tests/integration/api/test_api.py
```diff
@@ -553,6 +554,17 @@ def _compare_screenboard(board1, board2):
 
         share_res = dog.Screenboard.share(get_res['id'])
         assert share_res['board_id'] == get_res['id']
+        public_url = share_res['public_url'].replace('8080', '5000')
```
allegrem: The `replace('8080', '5000')` is a bit annoying, but I can't find a way to get rid of it in local env. Any idea @yannmh ? I guess it would work when testing against the staging env.

tests/integration/dogshell/test_dogshell.py
```diff
@@ -338,6 +340,21 @@ def test_screenboards(self):
         finally:
             os.unlink(updated_file)
 
+        # Share the screenboard
+        out, _, _ = self.dogshell(["screenboard", "share", str(screenboard["id"])])
+        out = json.loads(out)
+        assert out['board_id'] == screenboard['id']
+        # Verify it's actually shared
+        public_url = out['public_url'].replace('8080', '5000')
```
allegrem: same here @yannmh

tests/integration/api/test_api.py
yannmh: I am usually testing against staging or a prod demo account, so I never experienced this inconvenience before.
Still, I don't think we should commit this extra line: it's too much conditioned by your environment thus it could easily break in a different scenario (i.e. testing from your VM or not).

Maybe we can make this more flexible using an environment variable like we do to pass the Datadog URL ?

tests/integration/dogshell/test_dogshell.py
```diff
@@ -9,6 +9,7 @@
 import time
 import tempfile
 import unittest
+import urllib2
```
yannmh: Let's use `requests` instead. Easier, shorter and compatible Python 2.6—3.4  (not the case for urllib2) !

yannmh:
Thanks for the new feature @allegrem  :boom: 
I've added two small suggestions. Let me know what you think about it.

allegrem:
I fixed everything you suggested @yannmh . Ready to merge?

yannmh:
:+1: !

PR #51, opened 2015-05-21
Update host muting commands

wang-arthur:
- Also add a test for host muting


yannmh:
It looks great ! Thanks @wang-arthur.

I added my two cents to the PR to fix dogshell tests `setUp` method.

wang-arthur:
Thanks @yannmh!

PR #53, opened 2015-05-27
catch any excp when parsing datadg.cnf & fall back

### LINKED ISSUE #52: `initialize` method failing if `/etc/dd-agent/datadog.conf` is not accessible
Reported by a user:
```
The datadog python client v0.4 will exit with an error if the user running the script doesn't have permissions to access the datadog config file in initialize_datadog(). It appears it only accesses the file in order to get the hostname of the current machine. The code looks like it tries to fall back to other methods if the file doesn't exist, but doesn't appear to handle the permissions failure case.
```

yannmh:
Fix #52.


remh:
Looks good to me.

PR #58, opened 2015-06-12
Clarify monitor muting endpoints

wang-arthur:
- Adjust dogshell descriptions to distinguish between mute_all/unmute_all and mute/unmute
- Add new monitor unmute arg (`all_scopes`) to allow clearing all mute settings for a given monitor
- Update monitor mute tests
- Include additional information in 403 response exceptions

### DIFF
--- a/datadog/api/base.py
+++ b/datadog/api/base.py
@@ -121,7 +121,7 @@ class HTTPClient(object):
                 cls._timeout_counter += 1
                 raise HttpTimeout('%s %s timed out after %d seconds.' % (method, url, _timeout))
             except requests.exceptions.HTTPError as e:
-                if e.response.status_code == 404 or e.response.status_code == 400:
+                if e.response.status_code in (400, 403, 404):
                     pass
                 else:
                     raise
--- a/tests/integration/api/test_api.py
+++ b/tests/integration/api/test_api.py
@@ -356,10 +356,8 @@ class TestDatadog(unittest.TestCase):
         dog.Metric.send(metric="test.metric", points=1.0)
         dog.Metric.send(metric="test.metric", points=(time.time(), 1.0))
 
+    @attr('monitor')
     def test_monitors(self):
-        # Stop testing 'silenced' parameter
-        # Depending on the enabled flag (i.e. API release), 'silenced' can be a boolean or a dictionnary
-
         query = "avg(last_1h):sum:system.net.bytes_rcvd{host:host0} > 100"
 
         monitor_id = dog.Monitor.create(query=query, type="metric alert")['id']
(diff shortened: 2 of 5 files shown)


datadog/api/base.py
```diff
@@ -124,6 +124,14 @@ def request(cls, method, path, body=None, attach_host_name=False, response_forma
...
                     pass
                 else:
+                    if e.response.text:
+                        try:
+                            error_msg = json.loads(e.response.text).get('errors')
+                            if error_msg:
+                                raise requests.exceptions.HTTPError("{0} ({1})".format(
+                                                                    e.message, error_msg[0]))
+                        except json.JSONDecodeError:
+                            pass
```
yannmh: Could you provide further details about the use case for this please ?

wang-arthur: Sure! Some 403 responses include custom error messages. For example, if a user tries to `$ dog monitor unmute_all` when the `mute_all` setting isn't enabled, the 403 response includes the error message: `'Monitoring is not currently globally muted'`. Simply raising `HttpError`, however, just gives the user a generic `'403 Client forbidden error'`.

tests/integration/api/test_api.py
```diff
@@ -652,7 +653,7 @@ def test_monitor_muting(self):
         eq(dt, None)  # No response is expected.
 
         # We shouldn't be able to mute a simple alert on a scope.
-        assert_raises(ApiError, dog.Monitor.mute, monitor_id, scope='env:staging')
+        assert_raises(HTTPError, dog.Monitor.mute, monitor_id, scope='env:staging')
```
yannmh: The exception is proper to the API behavior so I think an ApiError makes more sense in this case.

yannmh:
Thanks alot  @wang-arthur. It looks great :+1: 

Before merging it, I think we should clarify which behaviors fall under the `HTTPError` or`ApiError` exceptions.

tests/integration/api/test_api.py
yannmh: Let's switch to an ApiError instead @wang-arthur.

datadog/api/base.py
```diff
@@ -124,6 +124,14 @@ def request(cls, method, path, body=None, attach_host_name=False, response_forma
                 if e.response.status_code == 404 or e.response.status_code == 400:
                     pass
```
yannmh: 403 should probably be caught in this block and raise at this point https://github.com/DataDog/datadogpy/blob/clarify_mute_all/datadog/api/base.py#L160

wang-arthur:
Thanks @yannmh! I made the revisions we discussed and cleaned up the commit history.

yannmh:
Thanks a lot @wang-arthur. Merging it for the v0.7.0 release :ship:

PR #60, opened 2015-06-23
[statsd] fix socket creation thread safeness

### LINKED ISSUE #57: Socket creation in DogStatsd class is not thread safe.
The code in ``get_socket()`` is:

```
    def get_socket(self):
        '''
        Return a connected socket
        '''
        if not self.socket:
            self.socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
            self.socket.connect((self.host, self.port))
        return self.socket
```

This code is not thread safe and can result in failure in a multithreaded application, with potential loss of initial metric data points on process startup.

The function which triggers the function is:

```
        try:
            # If set, use socket directly
        except socket.error:
            log.info("Error submitting packet, will try refreshing the socket")
            self.socket = None
            try:
                self.get_socket().send(packet.encode(self.encoding))
            except socket.error:
                log.exception("Failed to send packet with a newly binded socket")
```

Specifically, the line:

```
```

Because in ``get_socket()`` the newly created socket is assigned direct to ``self.socket``, then a separate thread calling into ``_send_to_server()`` can see the socket before the socket has been connected.

The end result is a socket error:

```
error: [Errno 39] Destination address required
```

in the second thread which can call ``send()`` on the unconnected socket.

This triggers the ``socket.error`` exception clause which wipes out ``self.socket`` and sets it to ``None``.

The original thread when it returns from ``connect()`` on the socket then returns ``None`` as a socket which the first thread then calls ``send()`` on with the resulting exception of:

```
[Sun Jun 07 19:28:26.101636 2015] [wsgi:error] [pid 89948:tid 4361912320]   File ".../datadog/dogstatsd/base.py", line 189, in _send_to_server
[Sun Jun 07 19:28:26.101660 2015] [wsgi:error] [pid 89948:tid 4361912320]     self.get_socket().send(packet.encode(self.encoding))
[Sun Jun 07 19:28:26.101681 2015] [wsgi:error] [pid 89948:tid 4361912320] AttributeError: 'NoneType' object has no attribute 'send'
```

To avoid this problem, ``get_socket()`` should be rewritten as:

```
    def get_socket(self):
        '''
        Return a connected socket
        '''
        if not self.socket:
            sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
            sock.connect((self.host, self.port))
            self.socket = sock
        return self.socket
```

In other words, assign the socket to a local variable and do the ``connect()`` before then assigning it to the class instance.

Note that although this avoids the presenting problem, the code still has a thread race condition due to no thread locking being used. That is, multiple threads could create the socket connection with only one winning out. This will not cause any issue except that an extra socket connection will temporarily exist until it is closed through reference count reduction in CPython or garbage collection in pypy.

This race condition is tolerable, although with the way that checks are done it would be trivial to make socket creation entirely thread safe without introducing thread lock contention beyond the initial point of creation the first time it is required.

yannmh:
Fix #57. Thanks @GrahamDumpleton !

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -47,17 +47,21 @@ class DogStatsd(object):
         self.close_buffer()
 
     def get_socket(self):
-        '''
-        Return a connected socket
-        '''
+        """
+        Return a connected socket.
+
+        Note: connect the socket before assigning it to the class instance to
+        avoid bad thread race conditions.
+        """
         if not self.socket:
-            self.socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-            self.socket.connect((self.host, self.port))
+            sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+            sock.connect((self.host, self.port))
+            self.socket = sock
         return self.socket
 
     def open_buffer(self, max_buffer_size=50):
         """
-        Open a buffer to send a batch of metrics in one packet
+        Open a buffer to send a batch of metrics in one packet.
 
         You can also use this as a context manager.
 
@@ -71,7 +75,7 @@ class DogStatsd(object):
 
     def close_buffer(self):
         """
-        Flush the buffer and switch back to single metric packets
+        Flush the buffer and switch back to single metric packets.
         """
         self._send = self._send_to_server
         self._flush_buffer()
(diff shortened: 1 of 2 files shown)


datadog/dogstatsd/base.py
```diff
@@ -47,17 +47,21 @@ def __exit__(self, type, value, traceback):
...
+        Note: connect the socket before assigning it to the class instance to
+        avoid bad thread race conditions.
+        """
         if not self.socket:
-            self.socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-            self.socket.connect((self.host, self.port))
+            sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+            sock.connect((self.host, self.port))
+            self.socket = sock
         return self.socket
```
jdavisp3: I'm seeing these races in a newer version of this module. Pretty sure there is still a race in this code. If another thread closes the socket between line 59 and 60, `get_socket` could return `None` or a closed socket. Similar race in `close_socket` itself.

PR #63, opened 2015-06-27
update requests library, per CVE-2015-2296

miketheman:
Versions prior to 2.6.0 are vulnerable to a specific session hijack
attack. While we do not use cookies when interacting with the Datadog
API, it is good practice to keep up with security patches like this.

More info: http://www.openwall.com/lists/oss-security/2015/03/14/4


yannmh:
Great, thanks a lot @miketheman.

PR #64, opened 2015-06-30
'metric_type' should be 'type'

yannmh:
API documentation was illegitimately promoting usage of `metric_type`
parameter instead of `type`.
To be consistent and avoid 'backward incompatibilities', properly rename
this parameter.


degemer:
:+1:

PR #65, opened 2015-06-30
Add a timed context manager to statsd.

### LINKED ISSUE #25: Context managed `statsd.timed`
Right now you can use `statsd.timed`, that would be cool when you want to measure the runtime of some random code to be able to do:

```
with statsd.timed('execution_time'):
    [some code...]
```

clokep:
This enables a context manager in statsd so you can easily time a set of code without saving the start time, etc. manually:

```python
    with statsd.timed_context('user.query.time', sample_rate=0.5):
        # Do what you need to ...
        pass

    # Is equivalent to ...
    start = time.time()
    try:
        get_user(user_id)
    finally:
        statsd.timing('user.query.time', time.time() - start)
```

Two tests are also added for this code.

Fixes #25

Please let me know if you want anything changed, etc. I'd love to see this in the next release!

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -127,45 +127,29 @@ class DogStatsd(object):
         """
         self._report(metric, 'ms', value, tags, sample_rate)
 
-    def timed(self, metric, tags=None, sample_rate=1):
+    class _TimedContextManagerDecorator(object):
         """
-        A decorator that will measure the distribution of a function's run
-        time.  Optionally specify a list of tags or a sample rate.
-        ::
-
-            @statsd.timed('user.query.time', sample_rate=0.5)
-            def get_user(user_id):
-                # Do what you need to ...
-                pass
-
-            # Is equivalent to ...
-            start = time.time()
-            try:
-                get_user(user_id)
-            finally:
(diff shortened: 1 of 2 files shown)


yannmh:
Great feature @clokep, I love it. Thanks!

I think we could make `statsd.timed` acting as a decorator and a context manager, c.f.
http://stackoverflow.com/questions/9213600/function-acting-as-both-decorator-and-context-manager-in-python

What do you think ?

clokep:
@yannmh Sure, seems reasonable -- I'll just need to add a `__call__` method to my decorator. I'll try to get to that "soon".

clokep:
@yannmh Please let me know if this works for you. I wasn't sure if you like to squash things or not.

yannmh:
Thanks a lot @clokep, it works great. No need to squash, both commits lead to legit code.
Let's merge it so we can release it tomorrow with the v0.7.0 :rocket: .

clokep:
Thanks! I just realized there's a simpler way to do some of this code...same behavior, just reducing some code.

Feel free to merge https://github.com/percipient/datadogpy/commit/f1c8db2927b6f9f596ffd54afdd63b0350c20eed if you'd like.

literallywar:
do I need to include `pass` at the end of my code?

PR #66, opened 2015-06-30
Yann/timed context cosmetic

yannmh:
Thanks @clokep .


clokep:
:+1: Sorry I got that up just late!

And thanks for being so responsive by the way!

Issue #10, opened 2015-04-03
add option to time in ms


yannmh:
<a href="https://github.com/clutchski"><img src="https://avatars.githubusercontent.com/u/92482?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [clutchski](https://github.com/clutchski)**
_Monday Dec 17, 2012 at 17:20 GMT_
_Originally opened as https://github.com/DataDog/dogstatsd-python/issues/5_

----

add use_ms for `timed` decorator instead of seconds. e.g.

```
from statsd import statsd

statsd.use_ms = True

# This will use ms
statsd.timed('this.will.use.ms')
def foo():
      # pass

# Or ....
from statsd import statsd
statsd.timed('this.will.use.seconds')
def foo():
      # pass

statsd.timed('this.will.use.ms', use_ms=True):
def foo():
      # pass
```

yannmh:
Introduction of metric metadata will give a better and more generic solution to this type of issues.
Closing.

PR #67, opened 2015-07-10
Timer values are milliseconds

g--:
according to the statsd protocol specification, timer values are integers in milliseconds.

https://github.com/b/statsd_spec

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -152,7 +152,8 @@ class DogStatsd(object):
 
         def __exit__(self, type, value, traceback):
             # Report the elapsed time of the context manager.
-            self.statsd.timing(self.metric, time() - self.start,
+            time_ms = int(round(1000 * (time() - self.start)))
+            self.statsd.timing(self.metric, time_ms,
                                self.tags, self.sample_rate)
 
     def timed(self, metric, tags=None, sample_rate=1):
@@ -177,7 +178,8 @@ class DogStatsd(object):
             try:
                 get_user(user_id)
             finally:
-                statsd.timing('user.query.time', time.time() - start)
+                time_ms = int(round(1000 * (time.time() - start)))
+                statsd.timing('user.query.time', time_ms)
         """
         return self._TimedContextManagerDecorator(self, metric, tags, sample_rate)
 
--- a/tests/unit/dogstatsd/test_statsd.py
+++ b/tests/unit/dogstatsd/test_statsd.py
@@ -167,7 +167,7 @@ class TestDogStatsd(object):
 
         t.assert_equal('ms', type_)
         t.assert_equal('timed.test', name)
-        self.assert_almost_equal(0.5, float(value), 0.1)
+        self.assert_almost_equal(500, int(value), 100)
 
     def test_timed_context(self):
         with self.statsd.timed('timed_context.test'):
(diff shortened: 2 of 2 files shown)


g--:
rebuild

g--:
I don't understand why it's failing with python 2.6: import mock is failing.. hopefully it's unrelated?

remh:
Thanks a lot @g-- !

The tests are failing on that line:
https://github.com/testing-cabal/mock/blob/master/mock/mock.py#L132
Python 2.6 doesn't support dictionary comprehension.
But mock doesn't support python 2.6 anyway, so don't worry about it

We'll review your PR real soon, thanks again!

datadog/dogstatsd/base.py
```diff
@@ -152,7 +152,8 @@ def __enter__(self):
 
         def __exit__(self, type, value, traceback):
             # Report the elapsed time of the context manager.
-            self.statsd.timing(self.metric, time() - self.start,
+            time_ms = int(round(1000 * (time() - self.start)))
```
clokep: Isn't the `round` here gratuitous (and potentially confusing)?

remh:
@g-- Also this a compatibility breaking change. 

So we would need to release this change as a major version, but i'm not convinced it's worth it.

It's easy in the datadog web app to multiply the values by 1000.

@g-- @clokep What do you think ?

clokep:
Perhaps add another method that reports it the correct way (in milliseconds) and deprecate the current function. Eventually you can rename the 'correct' method back to this method.

Or add another endpoint in the datadog API which accepts things in milliseconds instead of seconds and use that in the datadog agent (which actually fwds this data off to the servers, IIRC).

g--:
Regarding compatibility, I like @clokep's idea.

As for whether it's worth it--I think it is, because it allows people to migrate from (or to) the statsd python client without their data being corrupted.  (This is what happened to us!)

I'll put together an updated PR with new methods when I get a chance.

datadog/dogstatsd/base.py
g--: Well.. I think it's a good idea to be explicit when converting from floats to integers with what to do with the decimal points.  The choices are round, floor or ceil.  rounding would yield the most accurate long-term data.

However, we're talking about fractions of milliseconds--it feels like it wouldn't matter either way.

yannmh:
Sorry to hear that you had trouble migrating from statsd python @g--.

While Datadog statsd timer values should be in miliseconds, it is, unfortunately, very difficult now to back out without introducing serious backward incompatibilities. 
I don't think creating a new method is worth: as long as the current method remains, the initial problem is still unsolved, i.e. respecting the statsd protocol specification or transparent migration with statsd Python.

Also, I am not sure to understand why data would be corrupted:  Datadog web app is not associating any unit to the timeserie values, and easily allows to apply a multiplicative factor.

g--:
Hi yannmh,

For some context, our existing data doesn't match up to the new data, so there's a sudden jump.  We could add a multiplicative factor.. if there's a way of multiplying data collected after a certain date, that would be a great solution!

Anyways, getting back to datadogpy--my understanding of the statsd protocol is time should be measured in integers of ms.  In the interest of compatibility, I thought it would be nice to converge on the existing standard.  I realize this creates the same problem we just faced for everyone else in reverse.  Separate end points would be one way around this.  What about parameterizing it so the same code paths can be used for both use cases?

yannmh:
> What about parameterizing it so the same code paths can be used for both use cases?

I like the idea ! It was suggested in a previous ticket (c.f. https://github.com/DataDog/datadogpy/issues/10) but not implemented. 
Ideally, the default parameter value would induce a compliant behavior with the current state of `datadogpy`, and be dropped in a future major version to comply with `statsd` protocol.

Thanks for the feedback @g-- @clokep !

yannmh:
I addressed our comments in a new PR #78. I am closing this one.
Thanks again.

PR #68, opened 2015-07-14
new constant tags for statsd client

jofusa:
I have added a optional param (constant_tags) to the Dogstatsd client to append tags to every metrics call.

Useful for setting project wide tags like version.

### DIFF
--- a/tests/unit/dogstatsd/test_statsd.py
+++ b/tests/unit/dogstatsd/test_statsd.py
@@ -113,7 +113,7 @@ class TestDogStatsd(object):
         def test_tags_and_samples(self):
             for i in range(100):
                 self.statsd.gauge('gst', 23, tags=["sampled"], sample_rate=0.9)
-            t.assert_equal('gst:23|g|@0.9|#sampled,bar:baz,foo')
+            t.assert_equal('gst:23|g|@0.9|#sampled')
 
     def test_timing(self):
         self.statsd.timing('t', 123)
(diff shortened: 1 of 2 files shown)


jofusa:
hmm that py26 failure is not happening locally. looks like a set comprehension in mock. anyone else seen this?

yannmh:
Thanks a lot @jofusa, it looks great. We'll review your PR really soon.

Python 2.6 failure seems related with https://github.com/testing-cabal/mock/issues/259. I'll make the necessary to fix it.

yannmh:
https://github.com/DataDog/datadogpy/pull/69 should fix Python 2.6 failure. Would you mind rebasing your PR on top of master please ?

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -113,7 +113,7 @@ def test_tags_and_samples(self):
         def test_tags_and_samples(self):
             for i in range(100):
                 self.statsd.gauge('gst', 23, tags=["sampled"], sample_rate=0.9)
-            t.assert_equal('gst:23|g|@0.9|#sampled')
+            t.assert_equal('gst:23|g|@0.9|#sampled,bar:baz,foo')
```
yannmh: I don't think this test will pass, as not part of `TestDogStatsdConstantTags` class.

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -252,6 +252,95 @@ def test_accessing_socket_multiple_times_returns_same_socket(self):
         t.assert_not_equal(FakeSocket(), dogpound.get_socket())
 
 
+class TestDogStatsdConstantTags(object):
```
yannmh: I don't think we need to duplicate all tests to check the constant tag feature, as they all use the same codepath (i.e. `_report` method). I would just add, to the existing class test, a single test like:
```python
def test_constant_tags(self):
    self.statsd.constant_tags=...
    self.statsd.gauge('gauge', 123.4)
    assert self.recv() == 'gauge:123.4|g|#bar:baz,foo'
```

What do you think ?

jofusa:
sounds great. will update tests

jofusa:
rebased off master

yannmh:
Thanks @jofusa ! It looks good to me.

Let's merge it so it'll be part of the next release, v0.8.0, planned for next week.

PR #70, opened 2015-07-16
Add option for SSL verification with Requests library

rogst:
In my production environment the server was missing root certificates to verify datadogs SSL cert which made the datadog api to fail with a No such file or directory error.
Added the option to use the verify parameter in requests to configure a ca certificates file or to disable verification.

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -32,7 +32,7 @@ else:
 
 
 def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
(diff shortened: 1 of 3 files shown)


yannmh:
Thanks a lot @rogst !
We'll review your PR soon, so we can release your changes with the next release.

yannmh:
Great and consistent with `requests` usage. Thanks again @rogst !
Merging it :rocket:.

PR #71, opened 2015-07-17
Make the metric optional for the timed decorator.

clokep:
If the metric is *NOT* provided, it defaults to the function's module and name.

I find this convenient to wrap some functions where the metric is simply going to be the name of the function, in particular to wrap celery tasks.

I discussed this briefly with @remh on IRC and he suggested I put up a PR and we can discuss it.


yannmh:
Great feature ! Thanks again @clokep.

Merging it for the next release.

clokep:
Thanks! :smile:

PR #76, opened 2015-08-06
[api] option to unmute ApiError exceptions

### LINKED ISSUE #74: Error handling: APIError is never raised
This library never returns the status code of the http request.
Instead, when the status code is 400, 403 or 404, it is supposed to raise an APIError exception with the content of the errors field received from the dispatcher.(https://github.com/DataDog/datadogpy/blob/master/datadog/api/base.py#L152).
For all other 3, 4, 5** status codes, an HTTPError is raised.

So catching errors and interacting with the API should be a matter of catching APIError and HTTPError exceptions.

BUT, the `_swallow` attribute prevents APIError to be raised (https://github.com/DataDog/datadogpy/blob/master/datadog/api/base.py#L171-L172). It's set to True in api/__init__.py and is not modifiable at the moment.

A temporary workaround is to use `from datadog import api api._swallow = False`.

We could evaluate the impact of a potential transition to swallow being True by default and also being set up when initializing the api.

yannmh:
Datadog API HTTP client raises ApiError (400, 403 or 404 status codes)
and HTTPError (others) exceptions for bad requests.
BUT, we are, by default, muting any exceptions before they escape from
the library. This makes difficult to catch errors and interact with the
API.

These changes introduces a `mute` option settable from `initialize`
method, to select the desired behavior.

Fix #74

### DIFF
--- a/tests/unit/api/helper.py
+++ b/tests/unit/api/helper.py
@@ -21,10 +21,10 @@ FAKE_PROXY = {
 }
 
 
-class MockReponse(requests.Response):
+class MockResponse(requests.Response):
 
     def __init__(self, raise_for_status=False):
-        super(MockReponse, self).__init__()
+        super(MockResponse, self).__init__()
         self._raise_for_status = raise_for_status
 
     def raise_for_status(self):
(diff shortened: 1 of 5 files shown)


tests/unit/api/helper.py
```diff
@@ -21,10 +22,15 @@
 
 
 class MockReponse(requests.Response):
-    content = None
+
+    def __init__(self, raise_for_status=False):
+        super(MockReponse, self).__init__()
```
olivielpeau: Nitpick, but could you rename `MockReponse` --> `MockResponse` :)

olivielpeau:
LGTM, love the tests!

Just a small nitpick on a name, apart from that :shipit:

yannmh:
Thanks @olivielpeau. Merging :anchor:

PR #78, opened 2015-08-06
[statsd] option to time in ms

### LINKED ISSUE #10: add option to time in ms

yannmh:
`use_ms` option to report timed values in milliseconds instead of
seconds (default False).

examples:
```
from statsd import statsd

statsd.use_ms = True

statsd.timed('this.will.use.ms')
def foo():
      # pass

from statsd import statsd
statsd.timed('this.will.use.seconds')
def foo():
      # pass

statsd.timed('this.will.use.ms', use_ms=True):
def foo():
      # pass
```

Fix #10, and inspired by #10, #67 
Thanks @g-- !

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -21,17 +21,28 @@ log = logging.getLogger('dogstatsd')
 class DogStatsd(object):
     OK, WARNING, CRITICAL, UNKNOWN = (0, 1, 2, 3)
 
-    def __init__(self, host='localhost', port=8125, max_buffer_size=50, constant_tags=None):
+    def __init__(self, host='localhost', port=8125, max_buffer_size=50,
+                 constant_tags=None, use_ms=False):
         """
         Initialize a DogStatsd object.
 
         >>> statsd = DogStatsd()
 
         :param host: the host of the DogStatsd server.
+        :type host: string
+
         :param port: the port of the DogStatsd server.
-        :param max_buffer_size: Maximum number of metric to buffer before sending to the server
-            if sending metrics in batch
-        :param constant_tags: A list of strings to attach to every metric reported by this client
+        :type port: integer
+
+        :param max_buffer_size: Maximum number of metrics to buffer before sending to the server
+        if sending metrics in batch
+        :type max_buffer_size: integer
+
+        :param constant_tags: Tags to attach to every metric reported by this client
+        :type constant_tags: list of strings
+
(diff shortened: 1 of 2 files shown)


olivielpeau:
LGTM! :alarm_clock:

datadog/dogstatsd/base.py
```diff
@@ -160,10 +173,12 @@ def __enter__(self):
 
         def __exit__(self, type, value, traceback):
             # Report the elapsed time of the context manager.
-            self.statsd.timing(self.metric, time() - self.start,
-                               self.tags, self.sample_rate)
+            elapsed = time() - self.start
+            use_ms = self.statsd.use_ms or self.use_ms
```
clokep: I'm unsure this really makes sense, someone could declare a statsd object with `use_ms` as `True`, then explicitly make a `_TimedContextManagerDecorator` with `use_ms` as `False` and this would resort to using milliseconds...is that expected?

Does this object even need this property or can it just pull it off its `statsd` object?

yannmh: Thanks @clokep, this is, indeed, unexpected. 

I think having this property is worth: it makes convenient to switch specific timers from `s` to `ms` without altering the overall `statsd` configuration.

clokep: I realized after writing this that if the decorator had it, then it had to be passed into here. I think the logic on this statement probably needs to be rethough (likely: if `self.use_ms` is not explicitly passed in, then default to `self.statsd.use_ms`).

yannmh:
Thanks @clokep. I adressed the points you mentionned and rebased my work.

@olivielpeau can you take a second look at it when you have a chance ?

olivielpeau:
Indeed it's better now! LGTM

PR #83, opened 2015-08-31
[deps] use `simplejson` with Python 3.x

yannmh:
`simplejson==3.0.0` introduced Python > 3.3 support. Use it instead of falling back with stdlib `json` package.


olivielpeau:
LGTM! :snake: :ship:

yannmh:
Thanks !

PR #84, opened 2015-09-05
typo fix on line 112 of timeboard,py  pull_all command 

jhotta:


datadog/dogshell/timeboard.py
```diff
@@ -109,7 +109,7 @@ def _title_to_filename(title):
             return no_punct.replace(" ", "_").replace("-", "_").strip("_")
 
         format = args.format
-        res = api.Timeboard.get_all()()
+        res = api.Timeboard.get_all()
```
yannmh: Thanks for fixing the typo :pencil2:  !

yannmh:
Thanks for the fixes @jhotta ! 

Rather than having the `_pretty_json` method defined in multiple sections of the code, what do you think of moving it to a file like `datadog/util/format.py` directory?

remh:
@yannmh yup this should be refactored. Can you do it so that we release this bug in a timely fashion please ?

yannmh:
Closing in favor of #92. Thanks @jhotta

PR #86, opened 2015-09-14
add tag support to dogwrap

dominis:
Usage:
`dogwrap -n test-job -k $API_KEY --tags "test, cron, test2" --submit_mode all "ls -lah"`

### DIFF
--- a/datadog/dogshell/wrap.py
+++ b/datadog/dogshell/wrap.py
@@ -189,6 +189,8 @@ case of error.")
     parser.add_option('-b', '--buffer_outs', action='store_true', dest='buffer_outs', default=False,
                       help="displays the stderr and stdout of the command only once it has \
 returned (the command outputs remains buffered in dogwrap meanwhile)")
+    parser.add_option('--tags', action='store', type='string', dest='tags', default='',
+                      help="comma separated list of tags")
 
     options, args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


yannmh:
Great feature, thanks a lot @dominis.

We'll review your PR shortly.

gnarf:
:+1: I wish this had this already!

I'm not the most proficient python programmer, but perhaps you could also submit a fix for #88 - it seems super easy, I'm just not sure what the opposite of the `.extend` method is :)

Zemmiph0bia:
+:1 Awesome!

@yannmh do you know when this will be merged into master?

yannmh:
Sincere apologies for the delay on this.
I'll take a pass on it today, so it'll be merged this week :rocket:

datadog/dogshell/wrap.py
```diff
@@ -189,6 +189,8 @@ def main():
     parser.add_option('-b', '--buffer_outs', action='store_true', dest='buffer_outs', default=False,
                       help="displays the stderr and stdout of the command only once it has \
 returned (the command outputs remains buffered in dogwrap meanwhile)")
+    parser.add_option('--tags', action='store', dest='tags', default=False,
+                      help="comma separated list of tags")
```
yannmh: I think setting the `default` to `default=''` would be more consistent.

Nitpicky: can we also set the `type` to `type=string` ?  Even if it's the default value (https://docs.python.org/2/library/optparse.html#optparse.Option.type), I think it makes things clearer :smile:

datadog/dogshell/wrap.py
```diff
@@ -239,6 +241,11 @@ def main():
     if notifications:
         event_body.extend([u'notifications: %s\n' % (notifications)])
 
+    if options.tags is not None:
```
yannmh: I don't think `options.tags` can take the `None` value. I would simply check
```python
if  options.tags:
   do_something()
```
to filter out `None`, `False` or `''`

yannmh:
Looking great, thanks for the contribution @dominis !

I just added two nitpicks, otherwise it's ready to be merged.

dominis:
@yannmh i've fixed the things you've mentioned

yannmh:
Thank you !

A new version of `datadogpy` will be out soon, with your changes.

PR #90, opened 2015-10-14
Adds constant_tags to ThreadStats

jofusa:
Added the same constant_tags to the ThreadStats class as I had done to the blocking version.

### DIFF
--- a/datadog/threadstats/base.py
+++ b/datadog/threadstats/base.py
@@ -22,10 +22,16 @@ log = logging.getLogger('dd.datadogpy')
 
 class ThreadStats(object):
 
-    def __init__(self):
-        """ Initialize a dogstats object. """
+    def __init__(self, constant_tags=None):
+        """
+        Initialize a dogstats object.
+
+        :param constant_tags: Tags to attach to every metric reported by this client
+        :type constant_tags: list of strings
+        """
         # Don't collect until start is called.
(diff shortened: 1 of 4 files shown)


yannmh:
Looking great ! Again, thanks @jofusa !

We'll merge your PR shortly so your changes can be shipped with the next release :ship:

datadog/threadstats/base.py
```diff
@@ -116,6 +122,13 @@ def event(self, title, text, alert_type=None, aggregation_key=None,
             'The web server is up again', alert_type='success')
         """
         if not self._disabled:
+            # Append all client level tags to every event
+            if self.constant_tags:
+                if tags:
+                    tags += self.constant_tags
+                else:
+                    tags = self.constant_tags
+
```
yannmh: `constant_tags` are not associated to events in `dogstatsd`. To be consistent, we should do the same with `ThreadStats`, what do you think ?

jofusa: Or I can add to dogstatsd. We have a use case for it.

jofusa:
Added constant tags to base agent. let me know if you like it, or i can rip it out

yannmh:
Thanks a lot @jofusa :exclamation: 
Changes will be released later today in a v0.10.0 version :ship:

PR #94, opened 2015-10-28
Fix bug where tags list increases in size

steven-liu:
## Summary
When previously-defined metric-level tags (non-literal list) are passed into the `gauge()` function,
the metric-level tag list will be modified if `constant_tags` is not `None`. This is due
to the `+=` operator modifying the metric-level list by appending
the list `constant_tags`.

If you're using dogstatsd to send your metrics, this will inevitably lead to a `error: [Errno 90] Message too long` since the UDP packet size has an upper limit.

For an example of this, see the new test, `test_gauge_constant_tags_with_metric_level_tags_twice`.

## Fix
Use the following instead of `+=`:
```python
# WILL modify metric-level tags passed into the function
tags += self.constant_tags

# WON'T modify metric-level tags passed into the function
tags = tags + self.constant_tags
```

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -225,7 +225,7 @@ class DogStatsd(object):
         # Append all client level tags to every metric
         if self.constant_tags:
             if tags:
-                tags += self.constant_tags
+                tags = tags + self.constant_tags
             else:
                 tags = self.constant_tags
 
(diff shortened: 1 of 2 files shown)


KyleJamesWalker:
+1

yannmh:
Thanks for the fix @steven-liu !

PR #101, opened 2015-11-15
Update __init__.py

emad:
Get line below 100 chars in order to pass build constraint.


yannmh:
Thanks @emad  for fixing it :lipstick: !

PR #102, opened 2015-11-16
[api] `Metrics`: support long` data type

yannmh:
`Metric` API supports built-in real numerical data types.

### DIFF
--- a/datadog/api/metrics.py
+++ b/datadog/api/metrics.py
@@ -14,13 +14,16 @@ class Metric(SearchableAPIResource, SendableAPIResource):
     _METRIC_QUERY_ENDPOINT = '/query'
     _METRIC_SUBMIT_ENDPOINT = '/series'
 
+    _SUPPORTED_DATA_TYPES = (int, float, long)
+
     @classmethod
     def _process_points(cls, points):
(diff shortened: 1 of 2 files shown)


olivielpeau:
Looking good, love the test :heart:

datadog/api/metrics.py
```diff
@@ -14,13 +14,16 @@ class Metric(SearchableAPIResource, SendableAPIResource):
     _METRIC_QUERY_ENDPOINT = '/query'
     _METRIC_SUBMIT_ENDPOINT = '/series'
 
+    _SUPPORTED_DATA_TYPES = (int, float, long)
+
     @classmethod
     def _process_points(cls, points):
         now = time.time()
-        if isinstance(points, (float, int)):
+        if isinstance(points, cls._SUPPORTED_DATA_TYPES):
```
mstepniowski: What about `Decimal`? Should we perhaps support all types that can be converted to `float` or `long`?

yannmh: Thanks for the note. You're right, we can handle these cases.

I am adding support to any 'real' number data-type.

PR #103, opened 2015-11-16
[api] `Metrics`: support real numerical data types

yannmh:
Extends `Metrics` data type support to any real number.


yannmh:
Thanks for the feedback @clutchski. I was actually using the OCAML notation :smile: 

I reworded the Pydoc to be clearer.

olivielpeau:
:+1:

PR #104, opened 2015-11-17
[dogwrap] new event format + trim output

### LINKED ISSUE #91: dogwrap: Trim output
When a command has output that makes the message exceed the 4000 character limit for events, the formating gets completely lost and so does the notifications that are appended ( #88 )

There should probably be a strategy in place to trim the output to something that fits in an event.  I'd suggest it be configurable.

Total size of the event can not be > 4000 characters, so a few different strategies make sense to me
* tail - Show only the last ~3500 characters (4k - however many bytes are used in headers/notifications/etc), for when the commands output really only matters what comes at the end
* head - show only the first ~3500 characters (^^^), for when the start of the command is really all that matters
* both (default?)- Take ~1000 from the top and ~2500 from the bottom with a `...trimmed...` in the middle, the first 1/3rd of the budget on the top, and the other 2 thirds on the bottom.

screenshot 2015-10-15 11 47 32

yannmh:
**Changes**
* New event body format:
```
	>>>> CMD <<<<
	>>>> EXIT CODE <<<<
	>>>> STDOUT <<<<
	>>>> STDERR <<<<
	>>>> NOTIFICATIONS <<<<
```
* Trim output to not exceed `MAX_EVENT_BODY_LENGTH` size
  * Keep the first 1/3rd of the budget on the top,
    and the other 2 thirds on the bottom.

Fix #91
Thanks @gnarf !

### DIFF
--- a/datadog/dogshell/wrap.py
+++ b/datadog/dogshell/wrap.py
@@ -32,6 +32,8 @@ from datadog import initialize, api
 SUCCESS = 'success'
 ERROR = 'error'
 
+MAX_EVENT_BODY_LENGTH = 3000
+
 
 class Timeout(Exception):
     pass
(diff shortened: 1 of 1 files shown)


remh:
:+1:

PR #105, opened 2015-12-07
[dogstatsd] add a top-level import for DogStatsd

yannmh:
Add a top-level import for DogStatsd, i.e.
```python
from datadog import DogStatsd
```
is equivalent to

```python
from datadog.dogstatsd.base import DogStatsd
```


truthbk:
:+1:

PR #106, opened 2015-12-08
Enhance compatibility with Google App Engine

yannmh:
###  `get_version` to resolve package version

Create a `get_version` util method to resolve `datadog` package version.

For compatibility purpose with Google App Engine, handle the case where `pkg_resources` is undefined.

### `simplejson` fall back to stdlib `json`

Enhance compatibility with Google App Engine: fall back `simplejson` to stdlib `json`.

### Split API and HTTP clients

Redesign the `api` module to dissociate the API from the HTTP client.
Make HTTP clients pluggable.

**Changes**
Two new modules:
* `api_client`.
  * `APIClient`. Defines
* `http_client`.
  * `HTTPClient`. Defines a pluggable HTTP client skeleton.
    * `RequestClient`. HTTP client based on `requests` 3p module.

One module renamed:
* `api.base` module is now `api.resources`

### `urlfetch` based HTTP client

Create a new HTTP client, `URLFetchClient`, based on `urlfetch` Google App Engine 3p module.

### Don't raise on hostname resolution failures

`hostname` is an optional API parameter. Thus, `initialize` method should not raise when unable to resolve it.


datadog/__init__.py
```diff
@@ -15,9 +15,9 @@
 from datadog import api
 from datadog.dogstatsd import DogStatsd, statsd  # noqa
 from datadog.threadstats import ThreadStats  # noqa
-from datadog.util.hostname import get_hostname
 from datadog.util.compat import iteritems
 from datadog.util.config import get_version
+from datadog.util.hostname import get_hostname
```
degemer: 🍰

datadog/api/api_client.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    """
+    # HTTP transport parameters
+    _backoff_period = _backoff_period
+    _max_timeouts = _max_timeouts
+    _backoff_timestamp = None
+    _timeout_counter = 0
+    _api_version = _api_version
+
+    def __init__(self):
```
degemer: This seems to be the only non-static method: you could remove it, set `_http_client` as a class variable, and `_get_http_client` the first time `submit` or another method is called. (or in the definition, or during the import, I'm not sure what's the best method)

datadog/api/api_client.py
```diff
@@ -0,0 +1,213 @@
...
+
+    @staticmethod
+    def _get_http_client():
+        """
+        Resolve the appropriate HTTP client based on priority and user environment.
+        """
+
+        return RequestClient
+
+    def submit(self, method, path, body=None, attach_host_name=False, response_formatter=None,
```
degemer: Related to the previous comment, this method should be a class method: it only uses class methods/variables.

datadog/api/api_client.py
```diff
@@ -0,0 +1,213 @@
...
+    _api_version = _api_version
+
+    def __init__(self):
+        """
+        Instantiate a HTTP Client.
+        """
+        self._http_client = self._get_http_client()
+
+    @staticmethod
+    def _get_http_client():
```
degemer: I actually don't think this logic should be here; it would be better in `datadog.api.http_client`. This way you can import `from datadog.api.http_client import get_http_client`, and I think it makes more sense to have the logic close to the declaration of the classes. What do you think ?

datadog/api/exceptions.py
```diff
@@ -1,34 +1,47 @@
...
+"""
 
 
-class ApiError(DatadogException):
+class ClientError(Exception):
+    """
+    When HTTP connection to Datadog endpoint is not possible.
+    """
+    def __init__(self, method, url, exception):
+        message = u"Could not request {method} {url}: {exception}. "\
```
degemer: 👏

datadog/api/api_client.py
```diff
@@ -6,7 +6,7 @@
 from datadog.api import _api_version, _max_timeouts, _backoff_period
 from datadog.api.exceptions import ClientError, ApiError, HttpBackoff, \
     HttpTimeout, ApiNotInitialized
-from datadog.api.http_client import RequestClient
+from datadog.api.http_client import get_http_client
```
degemer: 💟

datadog/api/http_client.py
```diff
@@ -85,4 +88,78 @@ class URLFetchClient(HTTPClient):
...
+def get_http_client():
+    """
+    Return the appropriate HTTP client based the defined priority and user environment.
+    """
+    global requests
+    global urlfetch
+    global urlfetch_errors
+
+    try:
+        import requests
```
degemer: I'm not a huge fan of this solution (use of `global` + `import` inside a method).
Why not `import` with `try...except` at the top of the file, defaulting to None, and then here just check for `None` or not ?

degemer:
👍

PR #108, opened 2015-12-15
Update timing doc

matthewdu:
Would be useful in the docs, so we wouldn't need to read source.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -135,7 +135,7 @@ class DogStatsd(object):
 
     def timing(self, metric, value, tags=None, sample_rate=1):
         """
-        Record a timing, optionally setting tags and a sample rate.
+        Record a timing in ms, optionally setting tags and a sample rate.
 
         >>> statsd.timing("query.response.time", 1234)
         """


yannmh:
Thanks for your contribution @matthewdu,

I think there might be a confusion here. The `value` passed to the `timing` method is not processed or converted, thus, it's submitted as it is to Datadog .
For instance,
```python
start = time.time()
try:
    get_user(user_id)
finally:
    statsd.timing('user.query.time', time.time() - start)
```
would report a value in seconds.

However, it's possible to instruct `statsd` to report timing values in seconds **or** milliseconds with the `timed` decorator and context manager. 
```python
@statsd.timed('user.query.time', sample_rate=0.5)
def get_user(user_id):
    # Do what you need to ...
    pass

# or
with statsd.timed('user.query.time', sample_rate=0.5):
    # Do what you need to ...
    pass
```
This can be done by setting `statsd.use_ms` variable (defaut is `False`) or  [`timed`'s `use_ms` optional argument](https://github.com/DataDog/datadogpy/blob/master/datadog/dogstatsd/base.py#L181).
In this case then, the value is [appropriately converted](https://github.com/DataDog/datadogpy/blob/master/datadog/dogstatsd/base.py#L178-L179) and passed to the `timing` method.

matthewdu:
Ok, makes sense. The value submitted is the raw integer. I had assumed that the `ms` here was specifying milliseconds: https://github.com/DataDog/datadogpy/pull/108/files#diff-941afa8d5894e459cf72c50d475bbc9cR142

yannmh:
The `ms` value you are mentionning is actually used by [`dogstatsd`](https://github.com/DataDog/dd-agent/blob/master/aggregator.py#L740) to resolve the metric type.

yannmh:
I am closing this PR. Please feel free to reach out or re-open if you have any question.

PR #110, opened 2016-01-05
Document parameter alert_type in Event.create

aknuds1:
As far as I can tell, you've forgot to document the parameter `alert_type` in `Event.create`. I guess there should also be a test for this, but I wasn't sure how to go about this.

### DIFF
--- a/datadog/api/events.py
+++ b/datadog/api/events.py
@@ -24,6 +24,9 @@ class Event(GetableAPIResource, CreateableAPIResource, SearchableAPIResource):
         :param text: event message
         :type text: string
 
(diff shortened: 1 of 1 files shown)


yannmh:
You're right `alert_type ` is a valid input as listed in the [API documentation](http://docs.datadoghq.com/api/). 
Thanks for adding it @aknuds1  :memo: !

About testing, this could be the subject of a small addition in our integration test suite :white_check_mark:. I'll make the according changes in a separate pull request.

aknuds1:
@yannmh Thanks. Yeah, I was going to add it to the integration test suite but I could not see how to run it :(

PR #111, opened 2016-01-07
Typos

meawoppl:
Its small, but it stuck out!


yannmh:
Thanks for fixing it :pencil2: :no_mouth: .

PR #112, opened 2016-01-14
Fix thread-safety of @timed decorator

mgood:
When using @timed as a decorator, it's not thread-safe to call itself as a
context manager. The decorator creates one instance of this class, and since
the context manager stores the start time on the instance, calling it again
in parallel will overwrite the start time of the earlier call.

So, for the decorator version, the start time needs to be kept in the call
stack.

This includes a test case which demonstrates the problem.

### DIFF
--- a/tests/unit/dogstatsd/test_statsd.py
+++ b/tests/unit/dogstatsd/test_statsd.py
@@ -6,6 +6,7 @@ Tests for dogstatsd.py
 from collections import deque
 import six
 import socket
+import threading
 import time
 
 from nose import tools as t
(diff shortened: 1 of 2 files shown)


yannmh:
This looks great ! Thanks a lot @mgood.

Adding your contribution to our `0.11.0` milestone. It'll be reviewed PR shortly :smiley:

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -220,6 +221,39 @@ def func(a, b, c=1, d=1):
...
+
+        # packet 2
+        packet = self.recv()
+        name_value, type_ = packet.split('|')
+        name, value = name_value.split(':')
+
+        t.assert_equal('ms', type_)
+        t.assert_equal('timed.test', name)
+        self.assert_almost_equal(0.5, float(value), 0.1)
+
```
yannmh: Thanks for the test @mgood. Can we move this test to https://github.com/DataDog/datadogpy/blob/master/tests/performance/test_statsd_thread_safety.py instead ?

mgood: Ah yes, that looks like a better place for it. I've moved on from the project that was using DataDog, but feel free to go ahead and move it over.

yannmh:
Thanks @mgood. 
I rebased your changes to bring the discussed changes, and opened a new one #126. I am closing this one.

PR #114, opened 2016-01-26
attach additional tags as defined in DOGSTATSD_TAGS environment variable

ewdurbin:
creates an additional technique for adding tags to all metrics by defining a comma separated list of tags in the `DOGSTATSD_TAGS` environment variable.

use case:

currently we add a plethora of global tags by adding them to our dd-agent configuration on each node. as we progress towards using container technologies, it would be swell if we could seamlessly inject some of these tags on a container by container basis (app name, version, build) rather than adding them to the dd-agent which will be configured with tags for instance id, region, availability-zone etc.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -45,7 +45,7 @@ class DogStatsd(object):
         :param use_ms: Report timed values in milliseconds instead of seconds (default False)
         :type use_ms: boolean
 
-        :envvar DOGSTATSD_TAGS: Tags to attach to every metric reported by dogstatsd client
+        :envvar DATADOG_TAGS: Tags to attach to every metric reported by dogstatsd client
         :type constant_tags: list of strings
         """
         self.host = host
@@ -54,7 +54,7 @@ class DogStatsd(object):
         self.max_buffer_size = max_buffer_size
         self._send = self._send_to_server
         self.encoding = 'utf-8'
-        env_tags = [tag for tag in os.environ.get('DOGSTATSD_TAGS', '').split(',') if tag]
+        env_tags = [tag for tag in os.environ.get('DATADOG_TAGS', '').split(',') if tag]
         if constant_tags is None:
             constant_tags = []
         self.constant_tags = constant_tags + env_tags
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -54,10 +54,10 @@ class DogStatsd(object):
         self.max_buffer_size = max_buffer_size
         self._send = self._send_to_server
         self.encoding = 'utf-8'
-        self.env_tags = [tag for tag in os.environ.get('DOGSTATSD_TAGS', '').split(',') if tag]
+        env_tags = [tag for tag in os.environ.get('DOGSTATSD_TAGS', '').split(',') if tag]
         if constant_tags is None:
             constant_tags = []
-        self.constant_tags = constant_tags + self.env_tags
+        self.constant_tags = constant_tags + env_tags
         self.use_ms = use_ms
 
     def __enter__(self):
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -4,6 +4,7 @@ DogStatsd is a Python client for DogStatsd, a Statsd fork for Datadog.
 """
 
 import logging
+import os
 from random import random
 from time import time
 import socket
@@ -43,6 +44,9 @@ class DogStatsd(object):
 
         :param use_ms: Report timed values in milliseconds instead of seconds (default False)
         :type use_ms: boolean
+
+        :envvar DOGSTATSD_TAGS: Tags to attach to every metric reported by dogstatsd client
+        :type constant_tags: list of strings
         """
         self.host = host
         self.port = int(port)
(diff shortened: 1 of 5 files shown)


yannmh:
Nice feature, thanks a lot @ewdurbin. We'll review and consider your changes for an upcoming release.

On a side note, I am quite confused by the use case you are describing. It sounds like you are looking for a convenient way to set constant tagging on Datadog Agent metrics (including metrics received from `statsd` protocol). If so, I think the following changes on `dd-agent` repository would be a better fit https://github.com/DataDog/dd-agent/pull/2095.
Your changes would only add constant tagging to metrics sent using `datadogpy` library, i.e. a subset of custom metrics. 
What do you think ?

ewdurbin:
@yannmh there's a bit of finesse.

Example:

Assume we have two docker hosts `hostA` and `hostB` running in separate availability zones, an application `widget` with version `1.0.0` and build number `666`, and an application `buzzer` with version `2.0.0` and build number `999`.

Currently, running dd-agent as a docker container on each docker host, we may configure it to have the following tags:
  - `instance-id`
  - `region`
  - `availability-zone`

Unfortunately this leaves the owners of the application to manually add the `application_name`, `application_version` and `application_build_number` to their code before instantiating their statsd client.

This feature allows us to inject `DOGSTATSD_TAGS` into the docker container with the application specific tags at launch and have it point blindly at the shared dd-agent, but still end up with the following tags:
  - `instance-id`
  - `region`
  - `availability-zone`
  - `application_name`
  - `application_version`
  - `application_build`

without needing to either manually add them in code (passing `constant_tags` to the client) or running a separate dd-agent for each application on the docker host.

yannmh:
It completely makes sense. Thanks for the further explanations @ewdurbin  !

datadog/dogstatsd/base.py
```diff
@@ -50,7 +54,10 @@ def __init__(self, host='localhost', port=8125, max_buffer_size=50,
         self.max_buffer_size = max_buffer_size
         self._send = self._send_to_server
         self.encoding = 'utf-8'
-        self.constant_tags = constant_tags
+        self.env_tags = [tag for tag in os.environ.get('DOGSTATSD_TAGS', '').split(',') if tag]
```
yannmh: Can we limit `env_tags` scope to the function only ?  i.e. 
```python
env_tags = ...

# Instead of
self.env_tags = ...
```

datadog/threadstats/base.py
```diff
@@ -28,10 +29,16 @@ def __init__(self, constant_tags=None):
...
         :param constant_tags: Tags to attach to every metric reported by this client
         :type constant_tags: list of strings
+
+        :envvar DOGSTATSD_TAGS: Tags to attach to every metric reported by ThreadStats client
+        :type constant_tags: list of strings
         """
         # Don't collect until start is called.
         self._disabled = True
-        self.constant_tags = constant_tags
+        self.env_tags = [tag for tag in os.environ.get('DOGSTATSD_TAGS', '').split(',') if tag]
```
yannmh: Same as for `datadog/dogstatsd/base.py`: I reckon there is no benefit to make it an instance variable.

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -402,6 +403,22 @@ def test_accessing_socket_multiple_times_returns_same_socket(self):
...
+        t.assert_equal('gt:123.4|g|#country:china,age:45,blue', statsd.socket.recv())
+
+    def test_tags_from_environment_and_constant(self):
+        os.environ['DOGSTATSD_TAGS'] = 'country:china,age:45,blue'
+        statsd = DogStatsd(constant_tags=['country:canada', 'red'])
+        del(os.environ['DOGSTATSD_TAGS'])
+        statsd.socket = FakeSocket()
+        statsd.gauge('gt', 123.4)
+        t.assert_equal('gt:123.4|g|#country:canada,red,country:china,age:45,blue', statsd.socket.recv())
+
```
yannmh: Thanks for the additional tests @ewdurbin !

Instead of potentially overriding, removing the host's `DOGSTATSD_TAGS ` environment variable, can you use the [`preserve_environ_datadog `](https://github.com/DataDog/datadogpy/blob/a60d519f774b28151339ffb20bd7f88905f12d0e/tests/unit/api/test_api.py#L32) decorator please ? We could also consider moving the method to a new `tests/util/` module :smile: .

yannmh:
It looks great, thanks again @ewdurbin.

I simply added a few nitpicks. Please let me know your thoughts about it, so we can merge your contribution.

tests/unit/dogstatsd/test_statsd.py
ewdurbin: it's not super clear to me how to apply `preserve_environ_datadog` to these test methods cleanly. i went ahead and implemented the logic in a separate helper as a context manager.

yannmh:
Thanks for the quick update @ewdurbin ! Let's merge it

yannmh:
I am thinking about adding the same feature to `dogstatsd`'s HTTP cousin [`ThreadStats`](https://github.com/DataDog/datadogpy/blob/master/datadog/threadstats/base.py). 
Thus, I am wondering if it wouldn't be better to rename `DOGSTATSD_TAGS ` to `DATADOG_TAGS` instead. What do you think @ewdurbin ?

ewdurbin:
This PR already includes the addition to ThreadStats. I'm fine with renaming it to a more general `DATADOG_TAGS`.

yannmh:
Thanks @ewdurbin. Let's merge it :ship:

PR #117, opened 2016-03-01
[threadstats] do not modify metric/event-level tags

yannmh:
Do not modify metric-level `tags` parameters when `constant_tags` is
set.
Similar to #94.


olivielpeau:
:+1:

PR #119, opened 2016-03-02
don't send None to statsd daemon as metric value

dcrosta:
If you do, you'll get errors like this in `/var/log/messages`:

```
Metric value must be a number: metric.name, None
```

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -133,10 +133,8 @@ class DogStatsd(object):
         >>> statsd.decrement('files.remaining')
         >>> statsd.decrement('active.connections', 2)
         """
-        if value is None:
-            return
-
-        self._report(metric, 'c', -value, tags, sample_rate)
+        metric_value = -value if value else value
+        self._report(metric, 'c', metric_value, tags, sample_rate)
 
     def histogram(self, metric, value, tags=None, sample_rate=1):
         """
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -133,6 +133,9 @@ class DogStatsd(object):
         >>> statsd.decrement('files.remaining')
         >>> statsd.decrement('active.connections', 2)
         """
+        if value is None:
+            return
+
         self._report(metric, 'c', -value, tags, sample_rate)
 
     def histogram(self, metric, value, tags=None, sample_rate=1):
@@ -231,6 +234,9 @@ class DogStatsd(object):
 
         More information about the packets' format: http://docs.datadoghq.com/guides/dogstatsd/
         """
+        if value is None:
+            return
+
         if sample_rate != 1 and random() > sample_rate:
             return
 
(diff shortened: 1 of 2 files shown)


yannmh:
Thanks @dcrosta for fixing this. I am adding your contribution to our 0.11.0 milestone.

dcrosta:
Cool. Is there a projected date for 0.11.0?

yannmh:
@dcrosta, we are planning a release in about two weeks :ship: . 

Additionally, I just noticed that some recent merges are conflicting with your branch: would you mind doing a quick rebase of your changes on top of the current state of the `master` branch please ?

dcrosta:
@yannmh ok, it's rebased -- waiting on Travis now.

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -421,6 +421,27 @@ def test_tags_from_environment_and_constant(self):
...
+        assert self.recv() is None
+
+    def test_timing_doesnt_send_None(self):
+        self.statsd.timing('metric', None)
+        assert self.recv() is None
+
+    def test_histogram_doesnt_send_None(self):
+        self.statsd.histogram('metric', None)
+        assert self.recv() is None
+
```
yannmh: @dcrosta what would you think about concatenating those tests to a single one ?

datadog/dogstatsd/base.py
```diff
@@ -129,6 +129,9 @@ def decrement(self, metric, value=1, tags=None, sample_rate=1):
         >>> statsd.decrement('files.remaining')
         >>> statsd.decrement('active.connections', 2)
         """
+        if value is None:
+            return
```
yannmh: Can we replace this by:
```python

metric_value = -value if value else value
self._report(metric, 'c', metric_value, tags, sample_rate)
```

I like the idea of keeping the `if value is None:` assertion within the `_report` method.

yannmh:
Out of curiosity @dcrosta, could you share with us the use case that led to send `None` values via `statsd` ?

tests/unit/dogstatsd/test_statsd.py
dcrosta: I typically do one assertion per test -- then if the test fails, you know from the test name alone what the problem is. If you guys use a different style, I can definitely consolidate them, let me know what your preference is.

dcrosta:
We have a value we pass to `gauge()` which is sometimes `None` -- our code doesn't (currently) check whether it's `None` before calling `gauge()`. Obviously we could fix this, but I thought it might make more sense to share this with others who might have the problem.

datadog/dogstatsd/base.py
dcrosta: Sure, done.

yannmh:
Thanks for your reactivity @dcrosta. 

Let's merge your changes :ship: ! (oops seems like an other rebase is needed)

dcrosta:
OK, rebased again.

yannmh:
Thanks again !

PR #122, opened 2016-03-09
[dogshell] add hostname by default to event/metric posts

### LINKED ISSUE #115: `dog` commandline tool does not auto-detect hostname
https://github.com/DataDog/datadogpy/blob/ac0857cc0402752d041dc6b7d6bf45b9ddd1b696/datadog/dogshell/common.py#L97-L98
https://github.com/DataDog/datadogpy/blob/e387239ab7c1aa84cc69cfe4bd18b63cead68ad0/datadog/dogshell/metric.py#L31-L34

This then requires each use of the `dog` command to specify the hostname, or the `--localhostname` argument, but there is no auto-detection in place for default.

My Principle of Least Surprise tells me that _something_ should be detected by default, and the user should be allowed to override it with command line flags, config file values, environment variables, etc.

yannmh:
**[dogshell] add hostname by default to event posts**

By default, `dogshell` event posts (`dog event post` command) are now
automatically associated with the local host.
The `--host` and `--no-host` (new) options allow to override this value
with a given host, or no host.

 ```python
 # Event is associated with the local host
 dog event post "title" "content"

 # Event is associated with `foo.bar` host
 dog event post --host "foo.bar" "title" "content"

 # Event is not associated with any host
 dog event post --no-host "title" "content"
 ```
**[dogshell] add hostname by default to metric posts**

By default, `dogshell` event posts (`dog metric post` command) are now
automatically associated with the local host.
The `--host` and `--no-host` (new) options allow to override this value
with a given host, or no host.
The `--localhostname` option is deprecated, as made default when no
`--host` is set.

 ```python
 # Metric is associated with the local host
 dog metric post "metric.example" 1

 # Metric is associated with `foo.bar` host
 dog metric post --host "foo.bar" "metric.example" 1

 # Metric is not associated with any host
 dog metric post --no-host "metric.example" 1
 ```

Fix #115

### DIFF
--- a/datadog/api/base.py
+++ b/datadog/api/base.py
@@ -84,7 +84,7 @@ class HTTPClient(object):
                 if 'series' in body:
                     # Adding the host name to all objects
                     for obj_params in body['series']:
-                        if 'host' not in obj_params:
+                        if obj_params.get('host', "") == "":
                             obj_params['host'] = _host_name
                 else:
                     if body.get('host', "") == "":
--- a/datadog/api/base.py
+++ b/datadog/api/base.py
@@ -87,7 +87,7 @@ class HTTPClient(object):
                         if 'host' not in obj_params:
                             obj_params['host'] = _host_name
                 else:
-                    if 'host' not in body:
+                    if body.get('host', "") == "":
                         body['host'] = _host_name
 
             # If defined, make sure tags are defined as a comma-separated string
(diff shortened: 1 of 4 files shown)


yannmh:
@JohnLZeller can you take a pass at it please ?

JohnLZeller:
Looks good! :+1:

PR #126, opened 2016-03-11
[dogstatsd] fix thread-safety of @timed decorator

yannmh:
**[dogstatsd][test] rewrite thread safety tests**
Rewrite DogStatsd thread safety tests.

**[dogstatsd] fix thread-safety of @timed decorator**
Rebase of #112.
> When using @timed as a decorator, it's not thread-safe to call itself as a context manager. The decorator creates one instance of this class, and since the context manager stores the start time on the instance, calling it again in parallel will overwrite the start time of the earlier call.
So, for the decorator version, the start time needs to be kept in the call
stack.
This includes a test case which demonstrates the problem.

Rebased to move tests to `tests/unit/dogstatsd/test_statsd.py`
Thanks again @mgood !

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -174,8 +174,11 @@ class DogStatsd(object):
 
             @wraps(func)
             def wrapped(*args, **kwargs):
-                with self:
+                start = time()
+                try:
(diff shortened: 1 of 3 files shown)


mgood:
:+1:

PR #134, opened 2016-05-23
[dogstatsd] Allow to dynamically use default route as a statsd host

remh:
When you run in a containerized environment, the dogstatsd server will
likely expose its port on the host network. If you have an application
that runs within a container, it needs to connect to it using the
default route.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -8,6 +8,7 @@ import os
 from random import random
 from time import time
 import socket
+import struct
 from functools import wraps
 
 try:
@@ -23,7 +24,7 @@ class DogStatsd(object):
     OK, WARNING, CRITICAL, UNKNOWN = (0, 1, 2, 3)
 
     def __init__(self, host='localhost', port=8125, max_buffer_size=50, namespace=None,
-                 constant_tags=None, use_ms=False):
+                 constant_tags=None, use_ms=False, use_default_route=False):
         """
         Initialize a DogStatsd object.
 
@@ -50,6 +51,11 @@ class DogStatsd(object):
 
         :envvar DATADOG_TAGS: Tags to attach to every metric reported by dogstatsd client
         :type constant_tags: list of strings
+
+        :param use_default_route: Dynamically set the statsd host to the default route
+        (Useful when running the client in a container)
+        :type use_default_route: boolean
+
         """
         self.host = host
         self.port = int(port)
@@ -63,6 +69,9 @@ class DogStatsd(object):
         self.constant_tags = constant_tags + env_tags
         self.namespace = namespace
         self.use_ms = use_ms
+        self.use_default_route = use_default_route
+        if self.use_default_route:
+            self.host = self._get_default_route()
 
     def __enter__(self):
         self.open_buffer(self.max_buffer_size)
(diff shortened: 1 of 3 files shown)


datadog/dogstatsd/base.py
```diff
@@ -79,6 +100,9 @@ def get_socket(self):
         avoid bad thread race conditions.
         """
         if not self.socket:
+            if self.use_default_route:
+                self.host = self._get_default_route()
```
yannmh: Why do we need this if it's already set in the constructor ?

datadog/dogstatsd/base.py
```diff
@@ -79,6 +100,9 @@ def get_socket(self):
...
+
```
yannmh: This could be a good candidate for a Python *property* too.
```python
    @property.setter
    def use_default_route(self, value):
        self._use_default_route = value
       
       # Invalid the socket 
       self.socket = None

        if self._use_default_route:
            self.host = self._get_default_route()
```
with the following benefits:
1. We don't need to repeat `self.host = self._get_default_route()` twice.
2. The `host` and `socket` attributes would be systematically updated on `use_default_route` changes.

yannmh:
Looks good, I simply added a nitpick. Feel free to consider it or not, and merge the PR :)

datadog/dogstatsd/base.py
remh: We don't do it for the host or port though. Should we ?

aeneaswiener:
On Mac OS X this doesn't degrade gracefully:

```
>>> from datadog import initialize, statsd
>>> initialize(statsd_use_default_route=True)
>>> statsd.event("Test event", "Test body")
No handlers could be found for logger "dogstatsd"
  File "<stdin>", line 1, in <module>
  File "/Users/aeneaswiener/cytora/streaming-pipeline/crawlers/moreover-crawler/env/lib/python2.7/site-packages/datadog/dogstatsd/base.py", line 363, in event
    self._send(string)
  File "/Users/aeneaswiener/cytora/streaming-pipeline/crawlers/moreover-crawler/env/lib/python2.7/site-packages/datadog/dogstatsd/base.py", line 299, in _send_to_server
  File "/Users/aeneaswiener/cytora/streaming-pipeline/crawlers/moreover-crawler/env/lib/python2.7/site-packages/datadog/dogstatsd/base.py", line 107, in get_socket
    sock.connect((self.host, self.port))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
TypeError: coercing to Unicode: need string or buffer, NoneType found
```

Is this expected / Is there a fix for this?

yannmh:
That's definitely not expected.

Thanks for reporting @aeneaswiener. We are working on a fix.

remh:
@aeneaswiener thanks for the feedback.

The main use case for this option is when you run your application in a (docker) container and you have your statsd server listening on the host. So for this use case, OS X support is not necessary.

Can you tell us a bit more about your use case ?

aeneaswiener:
@remh We run our Python applications locally during development. Of course we could introduce an if clause when initialising dogstatsd to initialise differently depending on the environment, but it would be nicer if the ``statsd`` instance would just fail silently if there is no route to dogstatsd.

remh:
@aeneaswiener not sure that failing silently is the best thing to do here. This is a bad practice that could hide underlying issues.

We could raise a better exception though if we can't find the default router, something like
```python
raise IOError("Unable to find default route. This option is only supported on Linux")
```
Would that work ?

aeneaswiener:
That would work, yes. Though I am not sure if IO error is the best one to
raise as it leaks an implementation detail of the API.

Issue #139, opened 2016-05-27
`pull_all` fails on dashboards whose name contains utf8 chars


Phylliade:
When running a `dog timeboard pull_all ./`, the cmd fails for dashes whose name contains UTF8 chars. For example `dialtøne`.

A fix is to convert strings to unicode in : 
https://github.com/DataDog/datadogpy/blob/master/datadog/dogshell/timeboard.py#L180-L182

A fix is underway (just waiting to have write access to the repo...)

Stacktrace : 
```
Traceback (most recent call last):
  File "/usr/local/bin/dog", line 9, in <module>
    load_entry_point('datadog==0.12.0', 'console_scripts', 'dog')()
  File "/Library/Python/2.7/site-packages/datadog/dogshell/__init__.py", line 74, in main
    args.func(args)
  File "/Library/Python/2.7/site-packages/datadog/dogshell/timeboard.py", line 130, in _pull_all
    args.timeout, format, args.string_ids)
  File "/Library/Python/2.7/site-packages/datadog/dogshell/timeboard.py", line 182, in _write_dash_to_file
    print("{0} {1}".format(dash_id, filename))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf8' in position 7: ordinal not in range(128)
```

yannmh:
Thanks for reporting this issue ❌ 🔣  @Phylliade . 

I am looking forward to review your fix 😃

PR #140, opened 2016-05-31
Phylliade/unicode fix

Phylliade:
A fix to https://github.com/DataDog/datadogpy/issues/139


yannmh:
Thanks for the quick fix 🔨 !

Issue #142, opened 2016-07-08
dictionary changed size during iteration for self._metrics in ThreadStats


leozc:
While using **ThreadStats** client under uwsgi:

> Error flushing metrics and events
> dictionary changed size during iteration
> on 
> Line: datadog/threadstats/metrics.py in flush at line 142
> past_intervals = [i for i in self._metrics if i < interval] 

The root cause is while iterating, new items are being added to self._metrics

yannmh:
Thanks for reporting the issue @leozc. We'll work on a fix for the next release !

yannmh:
Fixed with https://github.com/DataDog/datadogpy/pull/151

overset:
This is a regression in python3 that is susceptible to the keys() view object changing which is not threadsafe. In python3 with latest `datadogpy` we're still getting:

```
RuntimeError: dictionary changed size during iteration
  File "datadog/threadstats/base.py", line 307, in flush
    metrics, dists = self._get_aggregate_metrics_and_dists(timestamp or time())
  File "datadog/threadstats/base.py", line 348, in _get_aggregate_metrics_and_dists
    rolled_up_metrics = self._metric_aggregator.flush(flush_time)
  File "datadog/threadstats/metrics.py", line 173, in flush
    past_intervals = [i for i in self._metrics.keys() if i < interval]
  File "datadog/threadstats/metrics.py", line 173, in <listcomp>
    past_intervals = [i for i in self._metrics.keys() if i < interval]
```

This is related to the python3 [view object](https://docs.python.org/3.7/library/stdtypes.html?highlight=keys#dict-views) that `.keys()` returns:

> The objects returned by dict.keys(), dict.values() and dict.items() are view objects. They provide a dynamic view on the dictionary’s entries, which means that when the dictionary changes, the view reflects these changes.

py2.7 https://docs.python.org/2/library/stdtypes.html?highlight=keys#dict.keys
```
keys()
Return a copy of the dictionary’s list of keys. See the note for dict.items().
```
py3.7 https://docs.python.org/3.7/library/stdtypes.html?highlight=keys#dict.keys
```
keys()
Return a new view of the dictionary’s keys. See the documentation of view objects.
```

Here's a quick script to repro similar logic as in https://github.com/DataDog/datadogpy/blob/master/datadog/threadstats/metrics.py#L173 that fails with python3 and works with python2:

```python
import threading

def prune(data):
    data.pop(1)

def print_half(data):
[...]

PR #143, opened 2016-07-09
Fix dictionary changed size during iteration #142

leozc:
copy the key to a list before iterating to prevent the error.

It is less performance since keys were copied, and there should be better approaches

### DIFF
--- a/datadog/threadstats/metrics.py
+++ b/datadog/threadstats/metrics.py
@@ -140,6 +140,7 @@ class MetricsAggregator(object):
         """ Flush all metrics up to the given timestamp. """
         interval = timestamp - timestamp % self._roll_up_interval
         past_intervals = [i for i in list(self._metrics.keys()) if i < interval]
+        metrics = []
         for i in past_intervals:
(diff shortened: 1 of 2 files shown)


yannmh:
Thanks for fixing the issue @leozc 🙇 .

We'll merge your contribution very shortly.

datadog/threadstats/metrics.py
```diff
@@ -139,7 +139,7 @@ def add_point(self, metric, tags, timestamp, value, metric_class, sample_rate=1,
     def flush(self, timestamp):
         """ Flush all metrics up to the given timestamp. """
         interval = timestamp - timestamp % self._roll_up_interval
-        past_intervals = [i for i in self._metrics if i < interval]
+        past_intervals = [i for i in list(self._metrics.keys()) if i < interval]
```
yannmh: `self._metrics.keys()` already creates a copy of `self._metrics` keys. Let's remove the extra `list(...)`.

yannmh:
@leozc I made the change and opened a new PR #151. I am closing this one, thanks again for fixing the issue!

PR #147, opened 2016-08-08
_TimedContextManagerDecorator enter -> self

ross:
Have `_TimedContextManagerDecorator` return `self` from `__enter__` so that it can be modified within the block. e.g.

```Python
with timed('foo.bar', tags=['key:val']) as timer:
    # do some stuff
    if blip:
       timer.tags.append('result:blip')
```

This can be worked around by assigning timed's result to a var and then doing a `with timer`, but this is much cleaner and shouldn't result in any backwards compatibility issues.

### DIFF
--- a/tests/unit/dogstatsd/test_statsd.py
+++ b/tests/unit/dogstatsd/test_statsd.py
@@ -314,8 +314,8 @@ class TestDogStatsd(object):
         """
         # In seconds
         with self.statsd.timed('timed_context.test') as timer:
-            self.assertIsInstance(timer,
-                                  DogStatsd._TimedContextManagerDecorator)
+            self.assert_is_instance(timer,
+                                    DogStatsd._TimedContextManagerDecorator)
             time.sleep(0.5)
 
         packet = self.recv()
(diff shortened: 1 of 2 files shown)


yannmh:
Thanks for the change @ross.

I'll review the changes this week, and have it out with the next release by the end of the following week 🚢 .

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -313,7 +313,9 @@ def test_timed_context(self):
         Measure the distribution of a context's run time.
         """
         # In seconds
-        with self.statsd.timed('timed_context.test'):
+        with self.statsd.timed('timed_context.test') as timer:
+            self.assertIsInstance(timer,
```
yannmh: I need to update the tests to inherit from `unittest`. In the meantime would you mind using 
```python
t.assert_is_instance(...)
```

rather than 
```python
self.assertIsIstance(...)
```

ross: Done.

yannmh: Almost 😄  it should be `t.assert...` rather than `self.assert...`

ross: Interesting. I've haven't run across `nose.tools` being used before. Should be in place now.

yannmh:
🙇

PR #152, opened 2016-08-24
[dogstatsd] support unicode tags 🔣

### LINKED ISSUE #132: Unicode Characters in Tags Cause Problems
Seen in datadogpy 0.9.0 and 0.11.0 using Python 2.7.6 and 2.7.11.

It doesn't seem possible to use unicode characters in tags. Running the following causes an ``UnicodeDecodeError`` to be raised from the [use of ``str`` here](https://github.com/DataDog/datadogpy/blob/master/datadog/dogstatsd/base.py#L267):

```python
# -*- coding: utf-8 -*-

from __future__ import (absolute_import, division, print_function,
                        unicode_literals)

from datadog import statsd

statsd.increment('tagged_counter', 1, tags=['person:Jóhn', 'foo:bar'])
```

If the tags are encoded before they are sent to datadogpy, the error is raised in 0.11.0 [a couple lines before when the payload is being constructed](https://github.com/DataDog/datadogpy/blob/master/datadog/dogstatsd/base.py#L265).

```python
# -*- coding: utf-8 -*-

from __future__ import (absolute_import, division, print_function,
                        unicode_literals)

from datadog import statsd

statsd.increment('tagged_counter', 1, tags=['person:Jóhn'.encode('utf-8'), 'foo:bar'])
```

yannmh:
Using unicode with DogStatsD tags causes `UnicodeDecodeError`
exceptions.

Do not raise on unicode tags. Fix #132.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -2,19 +2,22 @@
 """
 DogStatsd is a Python client for DogStatsd, a Statsd fork for Datadog.
 """
-
-import logging
-import os
+# stdlib
+from functools import wraps
 from random import random
 from time import time
+import logging
+import os
 import socket
 import struct
-from functools import wraps
 
 try:
     from itertools import imap
 except ImportError:
     imap = map
+
+# datadog
+from datadog.util.compat import text
 
 
 log = logging.getLogger('dogstatsd')
@@ -289,7 +292,7 @@ class DogStatsd(object):
         if tags:
             payload.extend(["|#", ",".join(tags)])
 
-        encoded = "".join(imap(str, payload))
+        encoded = "".join(imap(text, payload))
 
         # Send it
         self._send(encoded)
--- a/tests/unit/dogstatsd/test_statsd.py
+++ b/tests/unit/dogstatsd/test_statsd.py
@@ -103,8 +103,8 @@ class TestDogStatsd(object):
         t.assert_equal('gt:123.4|g|#country:china,age:45,blue', self.recv())
 
     def test_tagged_counter(self):
-        self.statsd.increment('ct', tags=['country:canada', 'red'])
-        t.assert_equal('ct:1|c|#country:canada,red', self.recv())
+        self.statsd.increment('ct', tags=[u'country:españa', 'red'])
+        t.assert_equal(u'ct:1|c|#country:españa,red', self.recv())
 
     def test_tagged_histogram(self):
         self.statsd.histogram('h', 1, tags=['red'])
--- a/datadog/util/compat.py
+++ b/datadog/util/compat.py
@@ -24,6 +24,8 @@ if is_p3k():
     def iternext(iter):
         return next(iter)
 
+    text = str
+
 else:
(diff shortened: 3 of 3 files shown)


tests/unit/dogstatsd/test_statsd.py
```diff
@@ -103,8 +103,8 @@ def test_tagged_gauge(self):
         t.assert_equal('gt:123.4|g|#country:china,age:45,blue', self.recv())
 
     def test_tagged_counter(self):
-        self.statsd.increment('ct', tags=['country:canada', 'red'])
-        t.assert_equal('ct:1|c|#country:canada,red', self.recv())
+        self.statsd.increment('ct', tags=[u'country:españa', 'red'])
```
truthbk: that's right 💯

datadog/dogstatsd/base.py
```diff
@@ -289,7 +289,7 @@ def _report(self, metric, metric_type, value, tags, sample_rate):
         if tags:
             payload.extend(["|#", ",".join(tags)])
 
-        encoded = "".join(imap(str, payload))
+        encoded = "".join(imap(unicode, payload))
```
degemer: Should the empty string be `u''` ? :trollface:

degemer:
👍

Julian:
This doesn't seem correct to me. I'm trying to diagnose a related issue (I'll update when I do), but from a very quick read, IMHO #132 should be wontfixed and this backed out. tags should either be unicode or bytes, and the current behavior asks for bytes. This changeset will break existing tags.

Julian:
OK, it looks like the current behavior actually isn't consistent about which it's after, so I'm not sure what's intended. I can survey what the behavior is on different behaviors I guess after finding my own issue. This patchset is certainly assuming that everything passed to dogstatsd is unicode though (or accidentally-ASCII-encoded-bytes-on-py2).

yannmh:
Hi @Julian,
I am not sure to understand exactly what issue you are facing. Could you give us more details about it please ? For instance
* Steps to reproduce
* Results you received versus the ones you would expect

Thanks!

Julian:
@yannmh I've had to switch to debugging something else, I'll try to circle back and provide more detail, but this patchset changed something from calling `str` to calling `unicode`. That's not really safe -- the former fails when you call it on non-ASCII representable unicode, the latter fails in the inverse case, on non-ASCII representable bytes. So either the string literal, and the surrounding code here, are misleading, and should be unicode objects (and probably the documentation updated to say that tags should always be unicode objects now), or something has changed here.

You can reproduce by e.g. passing in a tag of `"foo:שלום"`, i.e. bytes, instead of `u"foo:שלום"`. It did look like however that the former is already broken because the `DogStatsd` object for some reason wants to call `packet.encode("utf-8")` on everything going over the socket, instead of just asking for bytes -- I'm not really sure what the intention is -- it seems likely that it's to accept unicode everywhere (that's not the decision I would have made, but maybe it's the one that exists already). I'll try to provide some more details if I get some free time to read through what's here.

Issue #135, opened 2016-05-26
Provide elapsed time when using 'timed' context manager


tuukkamustonen:
Consider:

```python
with statsd.timed('mymetric') as t:
    ...
t.ms  # get elapsed time
```

This is possible with `statsd` python package, but unfortunately not with datadog package (elapsed time is not stored in `_TimedContextManagerDecorator`. If I made a PR, would be willing to merge something like this:

```python
    class _TimedContextManagerDecorator(object):

        def __init__(self, statsd, metric=None, tags=None, sample_rate=1, use_ms=None):
            ...
            self.elapsed = None

        def __enter__(self):
            ...
            return self

        def _send(self, start):
            self.elapsed = time() - start
            ...
```

And yeah, I could call `time() - t.start` myself, but it wouldn't match the sent value as `time()` could be different already...

yannmh:
Nice idea @tuukkamustonen 💡 , I was not actually aware that such a method exists in the `statsd` Python package.
We'll be happy to review your PR if you decide to make one :)

tuukkamustonen:
Fixed in #154.

PR #154, opened 2016-08-31
Provide elapsed time from _TimedContextManagerDecorator

tuukkamustonen:
Tackles #135


yannmh:
Thanks for tackling this @tuukkamustonen 🙇

PR #146, opened 2016-08-03
add support for coroutine functions

thehesiod:
Without this change only the time it takes to create the coroutine is reported.

PR #159, opened 2016-09-21
[dogstatsd] `timed` support for coroutine functions

yannmh:
Rebase of #146.

**Additional changes:**
* Move Python 3.5+ evaluation methods to `datadog.util.compat` module
* Move `TimedContextManager` to `datadog.dogstatsd.context` module
* Quarantine `TimedContextManager` coroutine decorator in
  `datadog.dogstatsd.context_async`
* Unit test coverage to assert DogStatsd compatibility with coroutine
  functions

Thanks a lot of @thehesiod 🙇

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ nosetests.xml
 .DS_Store
 .eggs/
 .env/
+.idea/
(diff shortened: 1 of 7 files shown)


thehesiod:
cool, btw coroutines are supported in python versions < 3.5, but the await syntax is not.  If one wanted they could get it to work with older versions of python3 with the use of @coroutine and yield from.  Personally I thought it wasn't worth the backwards compatibility effort :)

PR #161, opened 2016-09-21
[logging] rename loggers and set handlers to null

yannmh:
Split `dd.datadogpy` logger to one per module, i.e.
* `datadog.api`
* `datadog.dogstatsd`
* `datadog.threadstats`

Setup the `NullHandler` handler by default as advised in
https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -8,6 +8,7 @@ without hindering performance.
 * datadog.dogshell: a command-line tool, wrapping datadog.api, to interact with Datadog REST API.
 """
 # stdlib
+import logging
 import os
 import os.path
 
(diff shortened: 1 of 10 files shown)


datadog/api/resources.py
```diff
@@ -8,7 +8,7 @@
 from datadog.api.api_client import APIClient
 
 
-log = logging.getLogger('dd.datadogpy')
+log = logging.getLogger('datadog.api')
```
degemer: Not used, 🔥  ?

datadog/util/config.py
```diff
@@ -9,7 +9,7 @@
 # CONSTANTS
 DATADOG_CONF = "datadog.conf"
 
-log = logging.getLogger('dd.datadogpy')
+log = logging.getLogger('datadog.api')
```
degemer: Not used, 🔥  ?

datadog/__init__.py
```diff
@@ -22,6 +23,10 @@
 
 __version__ = get_version()
 
+# Loggers
+logging.getLogger('datadog.api').addHandler(logging.NullHandler())
+logging.getLogger('datadog.dogstatsd').addHandler(logging.NullHandler())
```
degemer: What about `datadog.threadstats` ?

yannmh:
Fix https://github.com/DataDog/datadogpy/pull/161

Issue #156, opened 2016-09-02
statsd_use_default_route triggers failure on OSX


miketheman:
We often develop on OSX, and use a common settings file to initialize the library.

With the setting of use_default_route on OSX, we trigger this easily-reproducible failure:

```python
>>> from datadog import initialize, statsd
>>> datadog_options = { 'statsd_use_default_route': True }
>>> initialize(**datadog_options)
>>> statsd.increment('test')
No handlers could be found for logger "dogstatsd"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/miketheman/.../python-2.7.12/lib/python2.7/site-packages/datadog/dogstatsd/base.py", line 153, in increment
    self._report(metric, 'c', value, tags, sample_rate)
  File "/Users/miketheman/.../python-2.7.12/lib/python2.7/site-packages/datadog/dogstatsd/base.py", line 298, in _report
    self._send(encoded)
  File "/Users/miketheman/.../python-2.7.12/lib/python2.7/site-packages/datadog/dogstatsd/base.py", line 303, in _send_to_server
    (self.socket or self.get_socket()).send(packet.encode(self.encoding))
  File "/Users/miketheman/.../python-2.7.12/lib/python2.7/site-packages/datadog/dogstatsd/base.py", line 110, in get_socket
    sock.connect((self.host, self.port))
  File "/usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
TypeError: coercing to Unicode: need string or buffer, NoneType found
```

[_get_default_route](https://github.com/DataDog/datadogpy/blob/0c87da34bdcd0a6b0efc79e68026f752d5d425c0/datadog/dogstatsd/base.py#L86-L96) is failing to find the `/proc/net/route` file - it doesn't exist on OSX, but the `log.error` is never seen, `None` is returned, and used in [`get_socket`](https://github.com/DataDog/datadogpy/blob/0c87da34bdcd0a6b0efc79e68026f752d5d425c0/datadog/dogstatsd/base.py#L110), which is an invalid input.

[...]

PR #166, opened 2016-10-03
[dogstatsd] better exceptions on route resolution

yannmh:
DogStatsd fails resolving the system's default route in the following
scenarios:
* no proc filesystem
* no default gateway found in `/proc/net/route`

Raise user intelligible exceptions in those cases.

Fix https://github.com/DataDog/datadogpy/issues/156

### DIFF
--- a/datadog/api/api_client.py
+++ b/datadog/api/api_client.py
@@ -4,11 +4,15 @@ import time
 
 # datadog
 from datadog.api import _api_version, _max_timeouts, _backoff_period
-from datadog.api.exceptions import ClientError, ApiError, HttpBackoff, \
-    HttpTimeout, ApiNotInitialized
+from datadog.api.exceptions import (
+    ClientError,
+    ApiError,
+    HttpBackoff,
+    HttpTimeout,
+    ApiNotInitialized
+)
 from datadog.api.http_client import resolve_http_client
 from datadog.util.compat import json, is_p3k
-
 
 log = logging.getLogger('datadog.api')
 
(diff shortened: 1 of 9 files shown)


yannmh:
@remh can you take a quick pass on it please?

datadog/api/api_client.py
```diff
@@ -4,12 +4,16 @@
 
 # datadog
 from datadog.api import _api_version, _max_timeouts, _backoff_period
-from datadog.api.exceptions import ClientError, ApiError, HttpBackoff, \
-    HttpTimeout, ApiNotInitialized
+from datadog.api.exceptions import (
```
degemer: 🍰

datadog/dogstatsd/route.py
```diff
@@ -0,0 +1,32 @@
...
+def get_default_route():
+    """
+    Return the system default interface using the proc filesystem.
+
+    Returns:
+        string: default route
+
+    Raises:
+        `NotImplementedError`: No proc filesystem is found (non-Linux systems)
+        `StopIteration`: No default route found
```
degemer: Not sure to see the relation between "no default route found" and https://docs.python.org/2/library/exceptions.html#exceptions.StopIteration ? It doesn't seem to be used as an iterator anywhere, am I missing something ?

yannmh: The `get_default_route` method can be considered as an iterator on the `/proc/net/route` file: it keeps going  until a "match" is found, i.e.
```python
for iterator:
   if condition:
      return 
```

Considering your comment, I believe this might not be a good choice. My goal was to to stick with a Python builtin exception and this one seemed to be the best candidate .

yannmh:
Thanks @degemer 🙇

PR #167, opened 2016-10-08
fix socket "leak"

thehesiod:
Our unittests have been printing out warnings for awhile since we're unable to close the socket after it's nulled out.  This ensures the socket is closed if it was opened.


thehesiod:
btw tests are failing because the socket mock doesn't have a "close" method, test issue

yannmh:
It looks good to me. Thanks for the fix @thehesiod 🙇

PR #171, opened 2016-10-18
Correct docstring for histogram

tdhopper:
By default, histogram sends a more limited set of metrics than the documetation claims.


yannmh:
Thanks @tdhopper

PR #174, opened 2016-10-24
Bump default timeout from 3 to 60 seconds.

ojongerius:
We have  1000+ monitors in our setup, and often hit the 3 default timeout of 3 seconds when poking around using datadog/api.

This bumps the default setting to 60 seconds, which allows us to retrieve our monitors. I do realise  bumping the default value is not the most subtle approach, but as 3 seconds is pretty short -much shorter than any CLI http client I've seen, and considering the REST response times for larger customers are often over 3 seconds this could be good enough.

### DIFF
--- a/datadog/api/__init__.py
+++ b/datadog/api/__init__.py
@@ -10,7 +10,7 @@ _cacert = True
 
 # HTTP(S) settings
 _proxies = None
-_timeout = 3
+_timeout = 60
 _max_timeouts = 3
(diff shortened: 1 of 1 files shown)


yannmh:
We've been thinking about doing it for a while: let's 🚢  it.  

Thanks @ojongerius 🙇

forkloop:
When can you release this change? We are also starting to see more timeout recently.

PR #181, opened 2017-01-05
Add metric metadata endpoints

flapjack103:
- Adds GET and PUT operations for the metadata endpoint.
- Some confusion around https://github.com/DataDog/datadogpy/compare/agrant/metrics_metadata?expand=1#diff-0741b06f081c976a404268a44c97cc2fR15 and if this should be `metrics` or `metadata` ?

### DIFF
--- a/datadog/api/metadata.py
+++ b/datadog/api/metadata.py
@@ -1,0 +1,64 @@
+# stdlib
+import time
+from numbers import Number
+
+# datadog
+from datadog.api.exceptions import ApiError
+from datadog.api.resources import SearchableAPIResource, SendableAPIResource
+
+
+class Metadata(SearchableAPIResource, SendableAPIResource):
+    """
+    A wrapper around Metric Metadata HTTP API
+    """
+    _class_url = '/metrics'
+    _json_name = 'metrics'
+
+    @classmethod
+    def get(cls, metric_name):
+        """
+        Get metadata information on an existing Datadog metric
+
+        param metric_name: metric name (ex. system.cpu.idle)
+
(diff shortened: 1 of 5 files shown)


datadog/api/metadata.py
```diff
@@ -0,0 +1,60 @@
...
+        :type statsd_interval: integer
+
+        :return: JSON response from HTTP request
+
+        >>> api.Metadata.update(metric_name='api.requests.served', metric_type="counter")
+        """
+        if not metric_name:
+            raise KeyError("'metric_name' parameter is required")
+
+        return super(Metadata, cls)._trigger_class_action('PUT', 'metrics', metric_name, **params)
```
yannmh: Can we simply do 
```
return super(Metadata, cls).update(id=metric_name, **params)
```
?

datadog/api/metadata.py
```diff
@@ -0,0 +1,60 @@
+# datadog
+from datadog.api.resources import GetableAPIResource, CreateableAPIResource, \
+    UpdatableAPIResource, ActionAPIResource
+
+
+class Metadata(GetableAPIResource, CreateableAPIResource, UpdatableAPIResource, ActionAPIResource):
```
flapjack103: I added some clarification in the update docstring. The update request will create the metadata if it doesn't exist.

flapjack103:
Thanks for the review and fixing the violations. Let me know if the changes are good or if it needs anything else.

yannmh:
Looks great, thanks @flapjack103 🙇

PR #182, opened 2017-01-11
Python 2.6.x workaround for CentOS 6.8

crsuarez:


setup.py
```diff
@@ -43,6 +43,7 @@
     classifiers=[
         "License :: OSI Approved :: BSD License",
         "Operating System :: OS Independent",
+        "Programming Language :: Python :: 2.6",
```
yannmh: We dropped Python 2.6 support with `datadogpy==0.14.0`. As a result, our CI does not run anymore tests using this version of Python. 
Let's not add it as a classifier as we don't guarantee that things work, and will in the future.

yannmh:
Thanks for the change @crsuarez. I added a few comments, please let me know what you think.

Also, would you mind rebasing your changes on the latest master?

crsuarez:
@yannmh . I updated the PR with the suggestions presented. :)

datadog/dogshell/event.py
```diff
@@ -38,6 +38,9 @@ def print_event_details(event):
     prettyprint_event(event)
 
 
+time_pat = re.compile(r'(?P<delta>[0-9]*\.?[0-9]+)(?P<unit>[mhd])')
```
yannmh: Do we need this?

crsuarez:
I don't know who intended to put this however I was not able to PR with
that. I'll delete it. But I think someone else put it on purpose .

yannmh:
Thanks, it might be a side effect from rebasing 🙇 . Let's merge

PR #184, opened 2017-01-23
[threadstats] argument `metric_name` → `metric`

### LINKED ISSUE #175: ThreadStats methods use `metric_name` instead of `metric`
I'm trying to wrap datadog in a generic class that uses a ThreadStats object on heroku (where dogstatsd isn't running) but a regular DogStatsd object on EC2 (where dogstatsd is running).  I'm stymied in this by the fact that the basic statsd methods (`increment()`, `gauge()`, and `timing()`) have slightly different calling signatures for ThreadStats and DogStatsd.  In particular DogStatsd uses `metric` whereas ThreadStats uses `metric_name`.

Please write a shim or function decorator or something that lets downstream consumers of the classes call these methods using either `metric` or `metric_name` in the calling signature.

yannmh:
For consistency with `statsd`, rename the `metric_name` methods'
parameters to `metric`.

Fix #175. Thanks @dblackdblack

### DIFF
--- a/datadog/threadstats/base.py
+++ b/datadog/threadstats/base.py
@@ -146,7 +146,7 @@ class ThreadStats(object):
                 source_type_name=source_type_name, date_happened=date_happened, priority=priority,
                 tags=event_tags, host=hostname)
 
-    def gauge(self, metric_name, value, timestamp=None, tags=None, sample_rate=1, host=None):
+    def gauge(self, metric, value, timestamp=None, tags=None, sample_rate=1, host=None):
         """
         Record the current ``value`` of a metric. The most recent value in
         a given flush interval will be recorded. Optionally, specify a set of
@@ -158,10 +158,10 @@ class ThreadStats(object):
         >>> stats.gauge('cache.bytes.free', cache.get_free_bytes(), tags=['version:1.0'])
         """
         if not self._disabled:
-            self._metric_aggregator.add_point(metric_name, tags, timestamp or time(), value, Gauge,
+            self._metric_aggregator.add_point(metric, tags, timestamp or time(), value, Gauge,
                                               sample_rate=sample_rate, host=host)
 
-    def increment(self, metric_name, value=1, timestamp=None, tags=None, sample_rate=1, host=None):
+    def increment(self, metric, value=1, timestamp=None, tags=None, sample_rate=1, host=None):
         """
         Increment the counter by the given ``value``. Optionally, specify a list of
         ``tags`` to associate with the metric. This is useful for counting things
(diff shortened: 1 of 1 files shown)


dblackdblack:
This is a backwards-incomaptible change: if people are calling ThreadStats with the old naming scheme, this will break their usage.  That's why I recommended using a shim which allows either `metric` or `metric_name` to be passed in, perhaps with a deprecation warning.

yannmh:
@dblackdblack I couldn't find anything ruling on, whether or not, the change of a positional parameter name in a method signature is considered as backward incompatible. 

Python lets users define positional arguments with keywords, although I don't know if this is a "common"/"recommended"/"ok"  practice. Would you have more insights on that?

Also, and out of curiosity, is there any specific reason why you are using those methods with keywords?

As for using a shim, or decorator that would help supporting both parameter names, I am worried that this is adding too much complexity for such a small gain.

dblackdblack:
it is often a matter of personal preference whether to use positional or keyword arguments when calling a function.  Personally, I almost always use keyword arguments because that improves readability of code, which is why I ran into this issue in the first place.

yannmh:
Thanks for you input @dblackdblack,

It's a small change but as you highlighted it can break backward compatibility, and I would like to keep the code clear. As we are starting to get more of this kind of cards, I created a `1.0.0` milestone and added it to it.

marceloboeira:
👍

jirikuncar:
@yannmh you can make it backward compatible by keeping the old `metric_name` as last keyword argument.

```python
>>> def old(metric_name, value, timestamp=None):
...     return metric_name, value, timestamp
...
>>> old(**{'metric_name': 'name', 'value': 1})
('name', 1, None)
>>> def new(metric=None, value=None, timestamp=None, metric_name=None):
...     return metric or metric_name, value, timestamp
...
>>> new(**{'metric_name': 'name', 'value': 1})
('name', 1, None)
>>> new('name', 1, timestamp='now')
('name', 1, 'now')
```

datadog/threadstats/base.py
```diff
@@ -146,7 +146,7 @@ def event(self, title, text, alert_type=None, aggregation_key=None,
                 source_type_name=source_type_name, date_happened=date_happened, priority=priority,
                 tags=event_tags, host=hostname)
 
-    def gauge(self, metric_name, value, timestamp=None, tags=None, sample_rate=1, host=None):
+    def gauge(self, metric, value, timestamp=None, tags=None, sample_rate=1, host=None):
```
jirikuncar:
```suggestion
    def gauge(self, metric=None, value=None, timestamp=None, tags=None, sample_rate=1, host=None, metric_name=None):
```
- [ ] `raise TypeError` when `value is None` or `(metric or metric_name) is None`

gzussa:
Closing as part of the related issue #175

Issue #178, opened 2016-12-20
DeprecationWarning for config.py


donato:
Hi all, we are seeing our logfiles get polluted with datadog warnings which look like this:

```
/app/python/lib/python3.5/site-packages/datadog/util/config.py:116: DeprecationWarning: This method will be removed in future versions.  Use 'parser.read_file()' instead.
  config.readfp(skip_leading_wsp(open(config_path)))
```

Using Python 3.5 with version of datadog==0.14.0

yannmh:
Thanks @donato,

We are working on it for the next release 🚀

zippolyte:
Fixed in https://github.com/DataDog/datadogpy/pull/451, closing

PR #185, opened 2017-01-23
[util] do not read hostname from `datadog.conf`

### LINKED ISSUE #176: Why do we try to read `/etc/dd-agent/datadog.conf`?
Why does the `datadog` library try to read the Datadog Agent config on `/etc/dd-agent/datadog.conf`? It seems to only be used in https://github.com/DataDog/datadogpy/blob/969c7036ab95fc6c3aa23eead354de1ecf8fc108/datadog/util/hostname.py#L54, and it is causing some issues for us.

To be able to read the config file, the service running the `datadog` library must have read privileges for that file. However, the [default settings from the datadog cookbook](https://github.com/DataDog/chef-datadog/blob/master/recipes/dd-agent.rb#L80-L82) marks the config file as owned by `dd-agent:root`, with permissions `0640`. We don't want to run our services as either the `dd-agent` user or a user in the `root` group, and the datadog cookbook doesn't allow us to change the permissions.

It all works fine even without the read permissions - `datadog` defaults to parsing `hostname -f`. However, as of #161, we get a log line like

```
2016-10-30 20:57:28,874: INFO: (-) datadog.api: No agent or invalid configuration file found
```

every time we start our services.

Why do we need to read the hostname from the config file - isn't `hostname -f` good enough? Or do people often have custom hostnames in the datadog agent config file? How do people solve the config file permission issue?

yannmh:
To be able to read the `datadog.conf` Datadog Agent configuration file,
the user must have read privilege to the file (default to `0640
dd-agent:root`).

It is also useless: if a `datadog.conf` file is present on the host, one
can use `statsd` to have metrics, events... sent with the appropriate
hostname.

Fix #176, #178. Thanks @tgwizard, @donato.

### DIFF
--- a/datadog/util/compat.py
+++ b/datadog/util/compat.py
@@ -33,7 +33,6 @@ get_input = input
 
 # Python 3.x
 if is_p3k():
-    from io import StringIO
     import builtins
     import configparser
     import urllib.request as url_lib, urllib.error, urllib.parse
@@ -51,7 +50,6 @@ if is_p3k():
 # Python 2.x
 else:
     import __builtin__ as builtins
-    from cStringIO import StringIO
     from itertools import imap
     import ConfigParser as configparser
     import urllib2 as url_lib
--- a/datadog/util/config.py
+++ b/datadog/util/config.py
@@ -1,20 +1,8 @@
 import os
-import string
 import sys
 
 # datadog
-from datadog.util.compat import configparser, StringIO, is_p3k, pkg
-
-# CONSTANTS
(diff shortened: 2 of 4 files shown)


donato:
Thanks @yannmh - let me know when I should upgrade and test this out :)

dabcoder:
We took care of this in https://github.com/DataDog/datadogpy/pull/428, rather than fully deprecating this agent config hostname look up, it adds an option to turn this on/off.

PR #187, opened 2017-02-08
Update README.md

lukas-gitl:
I was wondering why the data was not submitted. Might want to consider adding a note...

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ from datadog import ThreadStats
 stats = ThreadStats()
 stats.start()
 stats.increment('home.page.hits')
+# sleep(10)  # data is only sent when full interval passed, so you might want to add a sleep for testing
 
 ```
 


yannmh:
Thanks for the note @lukas-gitl.

You are right, `ThreadStats` submits metrics to Datadog periodically and asynchronously, by default every 10 seconds (c.f [`flush_interval `](https://github.com/DataDog/datadogpy/blob/master/datadog/threadstats/base.py#L46)). Note that you can change this interval time, as well as trigger the flush manually using the [`ThreadStats.flush`](https://github.com/DataDog/datadogpy/blob/master/datadog/threadstats/base.py#L268-L273) method.

Using a `time.sleep` breaks the asynchronous benefits of the `ThreadStats` model. If one wants to know when data are submitted, I'd recommend disabling `flush_in_thread` and flushing metrics manually instead, i.e. 

```python
stats = ThreadStats()
stats.start(flush_in_thread=False)

# Populate metrics
stats.increment('home.page.hits')

# Flush metrics
stats.flush()  # hangs until metrics are posted
```

PR #199, opened 2017-05-16
add socket_path option to enable unix socket traffic to dogstatsd6

xvello:
### What does this PR do?

dogstatsd6 will allow custom metrics to be sent via a datagram unix socket instead of UDP (see https://github.com/DataDog/datadog-agent/pull/252). This PR add a use_unix_socket option to the DogstatsD object in order to allow using this communication channel.

### Motivation

Containers make it harder to reach the dogstatsd server from client applications and we don't have a future-proof network solution that could work across all orchestrators. This allows to bypass the network stack altogether and use a host-local protocol. See https://github.com/DataDog/docker-dd-agent/issues/195 for more context.

As unix datagram and UDP share the same semantics, the patch is pretty small and unobstrusive.

Socket connection is set to non blocking: if the server is unresponsive, the packets are dropped, mirroring what would happen with UDP.

### Additional Notes

Still WIP as we need to:

- [ ] Work on better error handling if the server stops (currently, the exceptions are caught and packets are dropped, but CPU usage increases significantly)

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -282,6 +282,9 @@ class DogStatsd(object):
         try:
             # If set, use socket directly
             (self.socket or self.get_socket()).send(packet.encode(self.encoding))
+        except socket.timeout:
+            # dogstatsd is overflowing, drop the packets (mimicks the UDP behaviour)
+            return
         except socket.error:
             log.info("Error submitting packet, will try refreshing the socket")
 
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -286,16 +286,8 @@ class DogStatsd(object):
             # dogstatsd is overflowing, drop the packets (mimicks the UDP behaviour)
             return
         except socket.error:
-            log.info("Error submitting packet, will try refreshing the socket")
-
+            log.info("Error submitting packet, dropping the packet and closing the socket")
             self.close_socket()
-
-            try:
-                self.get_socket().send(packet.encode(self.encoding))
-            except socket.error:
-                self.close_socket()
-
-                log.exception("Failed to send packet with a newly bound socket")
 
     def _send_to_buffer(self, packet):
         self.buffer.append(packet)
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -24,7 +24,8 @@ class DogStatsd(object):
     OK, WARNING, CRITICAL, UNKNOWN = (0, 1, 2, 3)
 
     def __init__(self, host='localhost', port=8125, max_buffer_size=50, namespace=None,
-                 constant_tags=None, use_ms=False, use_default_route=False):
+                 constant_tags=None, use_ms=False, use_default_route=False,
+                 use_unix_socket=None):
         """
         Initialize a DogStatsd object.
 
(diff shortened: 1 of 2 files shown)


datadog/dogstatsd/base.py
```diff
@@ -24,7 +24,8 @@ class DogStatsd(object):
     OK, WARNING, CRITICAL, UNKNOWN = (0, 1, 2, 3)
 
     def __init__(self, host='localhost', port=8125, max_buffer_size=50, namespace=None,
-                 constant_tags=None, use_ms=False, use_default_route=False):
+                 constant_tags=None, use_ms=False, use_default_route=False,
+                 use_unix_socket=None):
```
masci: I'd call this param `socket_path`

xvello:
Added an exception handling for socket.timeout (silently drop the packets), but I need some python-fu on how to handle dogstatsd crashing (generic socket.error).

Currently, the exception is graceful handled and traffic resumes as soon as dogstatsd restarts, but the CPU usage is significant (trying on a new socket for every packet). Should we detect repeat failures and throttle?

yannmh:
@xvello is it still a WIP?

xvello:
@yannmh removed the WIP flag as stress test went well
I'd need some input from you on whether to include a "connection failed N times, dropping packets for a second" mechanism (or similar), to avoid high CPU usage when the statsd server is down.
Could either be in this PR or in another down the road.

datadog/dogstatsd/base.py
```diff
@@ -265,6 +282,9 @@ def _send_to_server(self, packet):
         try:
             # If set, use socket directly
             (self.socket or self.get_socket()).send(packet.encode(self.encoding))
+        except socket.timeout:
+            # dogstatsd is overflowing, drop the packets (mimicks the UDP behaviour)
+            return
```
yannmh: Why do we need this? From what I understand, we should never get a timeout on non-blocking mode.

yannmh:
I am in favor of keeping the code very simple, i.e. avoid adding any back-off mechanism. 

Do you know what errors to expect? `error: [Errno 61] Connection refused` sounds like one, are there more? 
If the list is relatively small, maybe we can filter those and avoid the systematic retry we do on: https://github.com/DataDog/datadogpy/blob/b06997af5947d9cf185b6f49a5733e02b8e5c451/datadog/dogstatsd/base.py#L294.
It may also not be necessary to re-create the socket.

datadog/dogstatsd/base.py
xvello: The python socket implementation [implements non-blocking send as a zero-timeout send](https://docs.python.org/2/library/socket.html#socket.socket.setblocking) and sends a `socket.timeout` exception if the write does not return immediately (the queue, managed by the kernel, is full), I chose to silently drop the packet, mirroring what UDP would do. With dogstatsd6's goroutine-based intake, this should only happen if the CPU is saturated.

If needed, the socket queue length is configurable via sysctl net.unix.max_dgram_qlen (10 on my test machines).

xvello:
Here are the errors I could produce:

### socket.connect
- `[Errno 2] No such file or directory`: the socket is not created yet (dogstatsd not started?)
- `[Errno 111] Connection refused`: the socket file is present but no dogstatsd is listening (server killed -9 and didn't unlink the socket file)

In these cases, we should retry for the next packet

### socket.send
- `[Errno 111] Connection refused` once then `[Errno 107] Transport endpoint is not connected` if dogstatds stops listening (unlinking the socket or not). To resume communication once dogstatsd is ready again, we need to recreate the connection, even if the filename didn't change (the receiver changed) : we should drop the packet, but retry later to re-create the connection to catch the new dogstatsd. Indeed, trying twice per packet should be avoided in this case.

- `socket.timeout` if the queue is full (dogstatsd not reading fast enough) : I'm in favor of dropping the packet

datadog/dogstatsd/base.py
yannmh: > I chose to silently drop the packet, mirroring what UDP would do

@xvello I wonder if we should still add some logging here. What do you think?

yannmh: > That could turn into log spamming really fast

That's a good point!

xvello:
Tested one last time communicating with dogstatsd6, the packet drop logic works, so does recovery after dsd restart. Let's merge that!
@yannmh do we have something else waiting for a release?

Issue #201, opened 2017-05-16
Dogwrap fails to event with unicode in stderr, stdout when using buffer_outs option


ronindesign:
### [invalid] Real issue from `-b` (buffer_opts) with `stderr`,`stdout` redirect

**Environment**
* Shared Hosting (cPanel)
* Python 2.7.13
```
~$ ./.localpython/Python-2.7.12/python --version
Python 2.7.13
```
* Dog/Dogwrap v0.14.0
```
~$ ./.local/bin/dog --version
dog 0.14.0
~$ ./.local/bin/dogwrap --version
dogwrap 0.14.0
```

**Context**
Using `dogwrap` via cron to run a PHP cli script:
`/home/MYUSER/.local/bin/dogwrap -n "My Cron" -k APIKEY -b --submit_mode all --notify_error="Notifying @MYEMAIL1 @MYEMAIL2" --tags=cron,test "php -q cron.php" >/dev/null 2>&1`

The cron runs every 15 minutes. When there are no tasks, a small summary string of (ascii) text is sent to `stdout` from PHP cli (dogwrap correctly captures it, and I see it in my events view):
```
My Automation Task
===================================
...
 [OK] Completed
```

However, each and every time `cron.php` _does_ have a task to do, `dogwrap` fails to send the event to datadog server. The dogwrap command doesn't _seem_ to exit or crash, it simply finishes and prints buffered output from `stdout` and `stderr` like everything is fine, but the event never seems to send. The output in these cases, seen by removing null route `>/dev/null` is:
```
0/1 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]   0% < 1 sec/< 1 sec 16.0 MiB
 1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100% 2 secs/2 secs 34.0 MiB
My Automation Task
=================================
...
 [OK] Completed
```
Thus, the only seemingly significant difference I can see between the outputs is the use of extended character encoding via UTF-8. Specifically the use of:
(UTF-8) Light Shade (U+2591) | ASCII: 9617
https://www.utf8icons.com/character/9617/light-shade
(UTF-8) Dark Shade ( U+2593) | ASCII: 9619
https://www.utf8icons.com/character/9619/dark-shade

I believe this is likely a unicode encoding issue.

[...]

PR #203, opened 2017-05-18
Encode from unicode before writing to stdout, stderr.

ronindesign:
Fixes encoding issue in non-specific `sys.stdout.encoding=None` environment. See #201


yannmh:
Thanks a lot for writing the fix @ronindesign 🙇

ronindesign:
You're welcome. I'm working on a more complete patch for unicode support in `dogshell.wrap` that is downstream from this, but I wanted to fast track this fix so the printing would at least work.

PR #207, opened 2017-05-20
Service check only sends non-null parameters

### LINKED ISSUE #206: Dogshell.dog service_check, monitor includes null values
`dog` script's `service_check`, `monitor` options are not required, but are still sent (as null) when not provided. Assuming this is the behavior for all script commands.

Reproduce:
```shell
$ dog service_check check check_pg host0 0
ERROR: "message" parameter should be a string
```

The actual payload / body sent to server includes nulled options, when not specified.
```
{"status": 0, "tags": null, "timestamp": null, "host_name": "host0", "message": null, "check": "check_pg"}
```

Relates:
https://github.com/DataDog/datadogpy/blob/master/datadog/dogshell/service_check.py#L32
https://github.com/DataDog/datadogpy/blob/master/datadog/api/service_checks.py#L39
https://github.com/DataDog/datadogpy/blob/master/datadog/api/api_client.py#L112
https://github.com/DataDog/datadogpy/blob/master/datadog/api/api_client.py#L127

This could be fixed locally in datadog.dogshell.service_check` to not submit `None` options;
or this could be fixed in `datadog.api.service_checks` or `datadog.api.api_client`.

The first would have the most localized effect / least possibility for wider side effects to other tests, but if there are no cases for submitting null values to the DataDog API, then I would definitely recommend mitigating this directly in `datadog.api.api_client`.

ronindesign:
Fixes #206.

### DIFF
--- a/datadog/api/service_checks.py
+++ b/datadog/api/service_checks.py
@@ -32,8 +32,13 @@ class ServiceCheck(ActionAPIResource):
 
         :returns: Dictionary representing the API's JSON response
         """
-        if 'status' in params and params['status'] not in CheckStatus.ALL:
-            raise ApiError('Invalid status, expected one of: %s'
-                           % ', '.join(str(v) for v in CheckStatus.ALL))
+
+        # Validate checks, include only non-null values
+        for param, value in params.items():
+            if value is None:
+                del params[param]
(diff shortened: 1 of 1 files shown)


yannmh:
Thanks again @ronindesign. 

Your detailed explanations in #206 are very much appreciated.

PR #213, opened 2017-06-23
[dogstatsd] add distribution that tracks accurate percentiles

jeerim:
Adding a distribution metric type that tracks accurate percentiles. This is only collected by the agent and not being supported by the API at the moment.


arbll:
Looks like this was superseded by #249, closing.

Issue #216, opened 2017-07-03
Issue when using a list for `hosts` parameters


pmourlanne:
The documentation says that the `hosts` argument for `Event.create` is a list of strings : https://github.com/DataDog/datadogpy/blob/master/datadog/api/events.py#L48

When using a list of strings as host:
```
datadog.api.Event.create(**{
   'title': 'plop',
   'text': 'alo',
   'host': ['firsthost, 'secondhost', ],
})
```

The datadog API answers:
```
{'errors': ['Invalid JSON structure']}
```

I think the documentation should be updated to say the `host` argument expects a string.

yannmh:
Apologies for the late answer @pmourlanne. 

I fixed the documentation with #220.

PR #220, opened 2017-08-04
[api] event's `host` must be a string, not a list

yannmh:
Fix the documentation. When creating an event, the associated `host`
must be a string, not a list.


yannmh:
Fix #216

PR #227, opened 2017-08-23
[dogshell] add pull and push commands for monitor

taraslayshchuk:
I was wondering why it is missing. It is bunch of manual work.

### DIFF
--- a/datadog/dogshell/monitor.py
+++ b/datadog/dogshell/monitor.py
@@ -1,10 +1,15 @@
+# stdlib
+import platform
+
 # 3p
-from datadog.util.format import pretty_json
+import argparse
 
 # datadog
 from datadog import api
-from datadog.dogshell.common import report_errors, report_warnings
+from datadog.dogshell.common import report_errors, report_warnings, print_err
 from datadog.util.compat import json
+from datadog.util.format import pretty_json
+from datetime import datetime
 
 
 class MonitorClient(object):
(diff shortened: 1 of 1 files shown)


taraslayshchuk:
@yannmh, could you please review?

datadog/dogshell/monitor.py
```diff
@@ -173,6 +207,100 @@ def _show_all(cls, args):
...
+        api._timeout = args.timeout
+
+        def _name_to_filename(name):
+            # Get a lowercased version with most punctuation stripped out...
+            no_punct = ''.join([c for c in name.lower() if c.isalnum() or c in [" ", "_", "-"]])
+            # Now replace all -'s, _'s and spaces with "_", and strip trailing _
+            return no_punct.replace(" ", "_").replace("-", "_").strip("_")
+
+        format = args.format
+        res = api.Monitor.get_all(
```
platinummonkey: This won’t work for orgs with lots of monitors as it will auto paginate. At super scale there is also an alternative pagination method, I’ve implemented in a different tool in this PR: https://github.com/airbnb/interferon/pull/58

datadog/dogshell/monitor.py
```diff
@@ -173,6 +207,100 @@ def _show_all(cls, args):
...
+        for f in args.file:
+            try:
+                monitor_obj = json.load(f)
+            except Exception as err:
+                raise Exception("Could not parse {0}: {1}".format(f.name, err))
+
+            if 'id' in monitor_obj:
+                # Always convert to int, in case it was originally a string.
+                monitor_obj["id"] = int(monitor_obj["id"])
+                res = api.Monitor.update(**monitor_obj)
```
platinummonkey: Not a fan of this as it’s not very smart. For example you can’t change a monitor type but this will attempt and fail and leave the user confused.

PR #228, opened 2017-08-23
[dogshell] use env vars for api_key and app_key

### LINKED ISSUE #144: dogrc file shouldn't be required if api/app key env vars are set
```
$ DATADOG_API_KEY=foo DATADOG_APP_KEY=bar dog timeboard show_all
~/.dogrc does not exist. Would you like to create it? [Y/n] 
```

Even tho though env vars are used later on in the init https://github.com/DataDog/datadogpy/blob/master/datadog/__init__.py#L63-L64

This makes it difficult to automate things that only have access to env vars and not the file system.

taraslayshchuk:
Would check `DATADOG_APP_KEY` and `DATADOG_API_KEY` environment variables before using `~/.dogrc`.
Fix #144.

### DIFF
--- a/datadog/dogshell/__init__.py
+++ b/datadog/dogshell/__init__.py
@@ -28,10 +28,10 @@ def main():
                         default=os.path.expanduser('~/.dogrc'))
     parser.add_argument('--api-key', help="your API key, from "
                         "https://app.datadoghq.com/account/settings#api",
-                        dest='api_key', default=None)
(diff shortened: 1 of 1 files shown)


taraslayshchuk:
@yannmh, could you please review?

marceloboeira:
👍

yannmh:
Thanks @taraslayshchuk.

ipleten:
Wheeee!

PR #230, opened 2017-09-22
Add list metrics to datadog.api.metrics

jbain:
Adding support to the SDK for listing active metrics


yannmh:
Looks good! Thank you @jbain

PR #231, opened 2017-10-10
add support for default_sample_rate

dtao:
The idea is quite similar to constant_tags: when initializing a
DogStatsd object, you can specify a value that should apply to *all*
metrics.

This change is backwards compatible. You can still pass in sample_rate
on a per-call basis and the value will override the default.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -25,7 +25,7 @@ class DogStatsd(object):
 
     def __init__(self, host='localhost', port=8125, max_buffer_size=50, namespace=None,
                  constant_tags=None, use_ms=False, use_default_route=False,
-                 socket_path=None):
+                 socket_path=None, default_sample_rate=1):
         """
         Initialize a DogStatsd object.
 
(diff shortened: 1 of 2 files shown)


dtao:
Not trying to be pushy, just curious: have any maintainers looked at this? I'd love to know if there's support for the general idea, at least.

A concrete use case I had in mind would be to define different default values in different environments based on request volume. For example in a Django project, one might set this via a Django setting to 1 in dev and QA environments but something lower, say 0.1 or 0.05 (or whatever), in production environments.

dabcoder:
@dtao Apologies for the very long delay in getting back to you. We looked into this and this looks good, there is now a conflict in `datadog/dogstatsd/base.py`. Would you have a chance to fix it? If not let us know and we can follow up as well. 
Thanks!

dtao:
@dabcoder I appreciate you taking a look at this after all this time.

The truth is that my team is no longer using DataDog (!) so my _personal_ motivation for getting this across the line is, at this point, pretty diminished—hopefully understandably :)

That is by no means a knock on you or your team, as I understand that triaging pull requests from the community is likely not anyone's top priority day to day. (Believe me, I can relate!)

I would say that if you have the time and energy to fix the merge conflict on your end—I'm guessing it isn't a ton of work, I just haven't even really looked—that would probably be best. Otherwise, it might be another long stretch of time before this PR sees any more movement.

dabcoder:
@dtao Thanks for the very prompt feedback, much appreciated. 

> I would say that if you have the time and energy to fix the merge conflict on your end—I'm guessing it isn't a ton of work, I just haven't even really looked—that would probably be best. Otherwise, it might be another long stretch of time before this PR sees any more movement.

Yes indeed, took care of it in #470, but again, thanks for initiating this.

PR #232, opened 2017-10-25
Fix tests broken under Python 3

shargan:
Tested under 3.5 and 2.7. Everything in between ought to work as well.


yannmh:
Looks good to me. Thank you again @shargan !

PR #233, opened 2017-10-25
Speed up DogStatsd._report()

shargan:
As currently written, `DogStatsd._report()` can become a bottleneck to high-throughput applications. Profiling against a rough microbenchmark (see below) showed the main time sinks to be `str.join()` and `DogStatsd._report()` itself.

This diff is a little ugly, but dramatically more performant than extending lists and joining strings:
```
(.venv) datadogpy (master)$ python bench.py
Simple: 1000000 metrics in 8.31s (120314.07 metrics/s)
Buffered: 1000000 metrics in 3.87s (258083.36 metrics/s)

(.venv) datadogpy (master)$ git checkout shargan/performance
Switched to branch 'shargan/performance'

(.venv) datadogpy (shargan/performance)$ python bench.py
Simple: 1000000 metrics in 5.32s (187969.19 metrics/s)
Buffered: 1000000 metrics in 1.72s (580796.61 metrics/s)
```

I see similar performance improvements on both Python 2.7 and 3.5, as well as when the stats get more complicated (including several tags and a sample rate).

The micro-benchmark in question:
```python
import itertools
import time

from datadog.dogstatsd import DogStatsd

STAT_NAMES = itertools.cycle(["stat1", "stat2", "stat3", "stat4", "stat5"])

def bench_buffered(num, stats):
    stats.open_buffer()
    for _ in range(num):
        stats.increment(next(STAT_NAMES))
    stats.close_buffer()

def bench_simple(num, stats):
    for _ in range(num):
        stats.increment(next(STAT_NAMES))

if __name__ == "__main__":
    num = 1000000

    for name, func in (("Simple", bench_simple), ("Buffered", bench_buffered)):
        stats = DogStatsd(namespace="foo.bar.baz")
        start = time.time()
        func(num, stats)
        delta = time.time() - start
        print("%s: %d metrics in %.2fs (%.2f metrics/s)"
              % (name, num, delta, num / delta))
```

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -11,10 +11,7 @@ import socket
 # datadog
 from datadog.dogstatsd.context import TimedContextManagerDecorator
 from datadog.dogstatsd.route import get_default_route
-from datadog.util.compat import (
-    imap,
-    text,
-)
+from datadog.util.compat import text
 
 # Logging
 log = logging.getLogger('datadog.dogstatsd')
@@ -83,6 +80,8 @@ class DogStatsd(object):
         if constant_tags is None:
             constant_tags = []
         self.constant_tags = constant_tags + env_tags
+        if namespace is not None:
+            namespace = text(namespace)
         self.namespace = namespace
         self.use_ms = use_ms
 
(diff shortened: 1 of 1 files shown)


yannmh:
Thanks a lot @shargan for the detailed content on your PR 🙇 

The changes look great and I am looking forward to see them included in the next release!

PR #234, opened 2017-10-26
change http-method GET to POST on api.Screenboard#share

seiro-ogasawara:
The http-method of the share of the Screenboard API changed from GET to POST and it stopped working, I fixed it.


yannmh:
Thank you @seiro-ogasawara for fixing this.

PR #236, opened 2017-10-31
dogwrap python3 string changes

martin308:
Wrapping a command with python3 is broken due to string changes in python3. This decodes the string using the `_fwd_out` encoding. 

Another option is to just use the `utf-8` decoded line below for both.

```
Exception in thread Thread-1:
  File "/usr/local/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/dbt/src/datadog/datadog/dogshell/wrap.py", line 68, in run
    self._fwd_out.write(line)
TypeError: write() argument must be str, not bytes
```

### DIFF
--- a/datadog/dogshell/wrap.py
+++ b/datadog/dogshell/wrap.py
@@ -65,7 +65,7 @@ class OutputReader(threading.Thread):
         '''
         for line in iter(self._out.readline, b''):
             if self._fwd_out is not None:
-                self._fwd_out.write(line)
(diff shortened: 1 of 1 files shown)


agamdua:
Would be great if this patch is accepted and a release is made, many distros include >Python3.5 by default, and installing python2 just to use `dogwrap` for a cron job wouldn't make sense.

yannmh:
The changes look good. Thanks @martin308 !

PR #241, opened 2017-11-10
Fix ServiceCheck none parameters for calls from dogshell, etc on python3.

timvisher:
https://github.com/DataDog/datadogpy/issues/206#issuecomment-343512407 introduced a basic python3 error where we attempted to iterate over a dictionary we were modifying, yielding an exception.

We didn't have any unit tests around this so I added one. This fixes both the actual call that I was making and provides a regression test.

I'm happy to clean up the commits once I get basic validation that this is the right direction.

If it's more important than I realize to actually filter out the `None` params, we'll need to change the way that we do that (iterating over `keys` and building up a new dictionary with the appropriate key/value pairs.

### DIFF
--- a/datadog/api/service_checks.py
+++ b/datadog/api/service_checks.py
@@ -35,9 +35,7 @@ class ServiceCheck(ActionAPIResource):
 
         # Validate checks, include only non-null values
         for param, value in params.items():
-            if value is None:
-                del params[param]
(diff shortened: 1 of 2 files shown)


yannmh:
Thanks @timvisher 🙇 . 

Cleaning out the `None` parameters is not necessary anymore.

PR #242, opened 2017-11-16
ThreadStats - Send correct metric type instead of always sending Gauge

nilabhsagar:
# ThreadStats - Send correct Metric Type instead of always sending Gauge

**Problem**
At present in ThreadStats various types of metric gets created like gauge, count, histogram, etc. but at the time of flushing for each metric type is set to "gauge". This is a problem as sending count from server becomes gauge in the web. Changing the type in web will not have any effect as from the server it is always coming as gauge.

**Solution**
Assign the right metric type based on the metric functionality.

### DIFF
--- a/datadog/threadstats/metrics.py
+++ b/datadog/threadstats/metrics.py
@@ -94,8 +94,10 @@ class Histogram(Metric):
         metrics = [
             (timestamp, self.min, '%s.min' % self.name, self.tags, self.host, MetricType.Gauge),
             (timestamp, self.max, '%s.max' % self.name, self.tags, self.host, MetricType.Gauge),
-            (timestamp, self.count, '%s.count' % self.name, self.tags, self.host, MetricType.Count),
-            (timestamp, self.average(), '%s.avg' % self.name, self.tags, self.host, MetricType.Gauge)
+            (timestamp, self.count, '%s.count' % self.name, 
+             self.tags, self.host, MetricType.Count),
+            (timestamp, self.average(), '%s.avg' % self.name, 
+             self.tags, self.host, MetricType.Gauge)
         ]
         length = len(self.samples)
         self.samples.sort()
--- a/datadog/threadstats/metrics.py
+++ b/datadog/threadstats/metrics.py
@@ -8,6 +8,7 @@ import itertools
 from datadog.util.compat import iternext
 from datadog.threadstats.constants import MetricType
 
+
 class Metric(object):
     """
     A base metric class that accepts points, slices them into time intervals
@@ -94,9 +95,9 @@ class Histogram(Metric):
         metrics = [
             (timestamp, self.min, '%s.min' % self.name, self.tags, self.host, MetricType.Gauge),
             (timestamp, self.max, '%s.max' % self.name, self.tags, self.host, MetricType.Gauge),
-            (timestamp, self.count, '%s.count' % self.name, 
+            (timestamp, self.count, '%s.count' % self.name,
              self.tags, self.host, MetricType.Count),
-            (timestamp, self.average(), '%s.avg' % self.name, 
+            (timestamp, self.average(), '%s.avg' % self.name,
              self.tags, self.host, MetricType.Gauge)
         ]
         length = len(self.samples)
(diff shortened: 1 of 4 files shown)


nilabhsagar:
The above pull request will fix the issue #245

jayshao:
@nilabhsagar looks like this failed the Travis build for Python 2.5, but only for line-length. Are you still looking at this?

nilabhsagar:
@jayshao Yes, I can fix it. Could you help me with the exact line number to fix.

nilabhsagar:
@jayshao Its fixed now, All Travis build passed.

nilabhsagar:
Hi, Any plan to review and merge with the master?

nilabhsagar:
@yannmh Request you to review.

nicsnoek:
Please merge ASAP. Thanks @nilabhsagar for fixing this enormity ;-)

irabinovitch:
@niamster Thanks for reaching out.  We have been looking into your PR, and unfortunately it is not mergeable as is.  Currently the server side API would also require changes to support some of these metric types.  I expect we'll be able to provide more insight into viability and/or timeline next week.

nilabhsagar:
@irabinovitch Sure, Let me know when some steps are required from my side.

yannmh:
Thanks for working on a fix @nilabhsagar 🙇  I apologize for the late reply.

Your changes look good, and so do the tests. 
In order to merge the fix, we'd need to make two additional changes:

1. Metrics type `count` (or `rate`) require an additional `interval` parameter. It instruments the backend to update the metric metadata accordingly and makes it able to perform time rollup operations (such as `as_rate()`, `as_count`...). 
  The `flush` method can be updated to account for this extra parameter, similar to https://github.com/DataDog/dd-agent/blob/master/aggregator.py#L208-L222.
2. We encourage to use the metric type `rate` instead of `count`. Both types are very similar (two representations of the same information), but the `rate` type is generally more used at Datadog, hence more tested. This can be performed simply by dividing the value by the interval as in https://github.com/DataDog/dd-agent/blob/master/aggregator.py#L210.

Please let me know if you'd like make these changes, or have us add them to your existing PR.

nilabhsagar:
@yannmh I have done the changes based on your suggestion. I hope it is aligned with your suggestion. Do let me know if I am missing something. 

Please verify the commit "7cb37c9" as others are just line wrapping issue fix.

datadog/threadstats/metrics.py
```diff
@@ -55,9 +57,9 @@ def __init__(self, name, tags, host):
     def add_point(self, value):
         self.count.append(value)
 
-    def flush(self, timestamp):
+    def flush(self, timestamp, interval):
         count = sum(self.count, 0)
-        return [(timestamp, count, self.name, self.tags, self.host)]
+        return [(timestamp, count, self.name, self.tags, self.host, MetricType.Rate, interval)]
```
yannmh: We want to replace `count` by `count/float(interval)` to report a proper rate.

datadog/threadstats/metrics.py
```diff
@@ -88,21 +90,26 @@ def add_point(self, value):
...
         metrics = [
-            (timestamp, self.min, '%s.min' % self.name, self.tags, self.host),
-            (timestamp, self.max, '%s.max' % self.name, self.tags, self.host),
-            (timestamp, self.count, '%s.count' % self.name, self.tags, self.host),
-            (timestamp, self.average(), '%s.avg' % self.name, self.tags, self.host)
+            (timestamp, self.min, '%s.min' % self.name,
+             self.tags, self.host, MetricType.Gauge, interval),
+            (timestamp, self.max, '%s.max' % self.name,
+             self.tags, self.host, MetricType.Gauge, interval),
+            (timestamp, self.count, '%s.count' % self.name,
```
yannmh: Same as above. We need to transform the count to a proper rate.

yannmh:
@nilabhsagar added a small comment.

Other than that, it looks good!

yannmh:
Thank you @nilabhsagar. This will be part of our next release, which will go out this week.

PR #249, opened 2018-01-22
Add beta support for sending global distribution metrics via dogstatsd

jbarciauskas:
Adds support to datadogpy for sending global distribution values via dogstatsd.

This is a beta feature. 

I did not add support to the threadstats package because it doesn't seem to make sense - no local aggregation should be done on these.


yannmh:
Looks good. Thanks @jbarciauskas.

PR #261, opened 2018-04-13
Add hosts API

samanthadrago:
Adds `Hosts.get_all` and `Hosts.totals` to Datadogpy.


DaveedL:
The Tags section of the API also uses the old function: https://docs.datadoghq.com/api/?lang=python#get-host-tags

PR #263, opened 2018-04-18
Use instance variable instead of overwriting a method.

florean:
This simplifies troubleshooting.


florean:
Is anybody here?  This is a two byte change for what is an obvious mistake.  As a paying DataDog customer with a large Python codebase, I find the neglect of your Python client disturbing.

yannmh:
I apologize for the delay on this @florean. We are keeping track of the issues and PRs from this repo internally, and we unfortunately neglected to provide you with an update.

We are always working to improve these processes so that they don't happen in the future. However, should this occur again, please don't hesitate to drop a note to our support team.

We are planning a new release today, and this will be included. You will be credited in the changelog and release notes.

PR #264, opened 2018-04-19
Check if there are packets to send before sending an empty UDP packet

Tenzer:
This avoid cases where the buffer is initialised but then doesn't receive any
packets to send before it's flushed. That can either be because of conditional
sending of metrics or because a sample rate is used which means no packets has
been generated this time around. In those cases an empty UDP packet is still
sent when the buffer is flushed.

This fixes that by checking if there are any packets to send before attempting
to send them off.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -146,7 +146,10 @@ class DogStatsd(object):
         Flush the buffer and switch back to single metric packets.
         """
         self._send = self._send_to_server
-        self._flush_buffer()
+
+        if self.buffer:
+            # Only send packets if there are packets to send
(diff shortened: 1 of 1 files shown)


yannmh:
Apologies for the late answer @Tenzer. 

What do you think of moving this logic into the `close_buffer` method? This prevents the `_send_to_server` method from systematically checking the packet when it might not be relevant (hence limit the overhead).

Tenzer:
It sounds fine to me, I have updated the change to do that now.

yannmh:
Thank you @Tenzer !

Tenzer:
You're welcome :)

PR #267, opened 2018-05-02
Force integer division for truncation of stdout/stderr in dogwrap

glasnt:
👋

Observed behaviour: 

When `dogwrap` runs in a Python 3 environment with `submit_mode=all`, and there is both `stdout` and `stderr`, the following error occurred: 

```shell
    ...
    top_third=text[:max_len / 3],
TypeError: slice indices must be integers or None or have an __index__ method
```

Cause: 

This is because in Python 3, `int / int` no longer returns an int, it returns a float. And floats are invalid index values.

Solution: 

This integer division functionality can be forced by using `//` rather than `/` in both Python 3 and Python 2. (Tested in Python 2.7.12, Python 3.5.2; operator introduced in Python 2.2+ ([PEP238](https://docs.python.org/3/whatsnew/2.2.html#pep-238-changing-the-division-operator))

This change has only been tested manually, because it doesn't appear `dogwrap` is currently being tested (please correct me if I'm wrong)

### DIFF
--- a/datadog/dogshell/wrap.py
+++ b/datadog/dogshell/wrap.py
@@ -167,8 +167,8 @@ def trim_text(text, max_len):
         u"*...trimmed...*\n" \
         u"```\n" \
         u"{bottom_two_third}\n".format(
-            top_third=text[:max_len / 3],
-            bottom_two_third=text[len(text) - (2 * max_len) / 3:]
+            top_third=text[:max_len // 3],
+            bottom_two_third=text[len(text) - (2 * max_len) // 3:]
         )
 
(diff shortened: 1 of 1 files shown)


glasnt:
Oops, should have checked the existing PRs 😕

This is a partial duplicate of #260, but this PR includes a third change, ensuring all instances of division are forced to integer division

cabouffard:
Good catch on the third change! :)

yannmh:
@glasnt @cabouffard thanks a lot for the fix 🙇

Issue #55, opened 2015-05-28
Dogshell: Unicode error when piping/handling output containing international chars


samanthadrago:
I set up a monitor which contains an international character (ä).

When I run `dog monitor show_all` in the terminal, it outputs all monitors as expected.

But when I try to pipe the output or handle it at all, like `dog monitor show_all > dog.txt`, I get a Unicode error:

```
Traceback (most recent call last):
  File "/usr/local/bin/dog", line 9, in <module>
    load_entry_point('datadog==0.5.0', 'console_scripts', 'dog')()
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/datadog/dogshell/__init__.py", line 71, in main
    args.func(args)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/datadog/dogshell/monitor.py", line 147, in _show_all
    (d["type"])]))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe4' in position 26: ordinal not in range(128)
```

The user who reported this used a workaround: on line 142 of datadog/dogshell/monitor.py, replace `(cls._escape(d["message"]))` with `(str(d["message"].encode('ascii','backslashreplace')))` but there may be a better way to handle this.

From: https://datadog.zendesk.com/agent/tickets/26802

yannmh:
Moving to `0.11.0`.

yannmh:
This exception is raised when a characters cannot be encoded with the standard output encoding. 

More specifically, Python 2.x defaults the standard output encoding to `ascii`, when redirecting to a file, i.e.
```
python2.7 -c "import sys; print sys.stdout.encoding" | cat
None
```
compared to
```
python2.7 -c "import sys; print sys.stdout.encoding"                                                                                
UTF-8
```

For this reason `dog monitor show_all` command works well in the terminal, but fails when redirected to a file.

I see different options to address this issue:

1. Escape non `ASCII` characters whatever the output encoding is set to.
2. Escape output encoding unsupported characters only.
3. Force/set the standard output encoding to `UTF-8`
4. Do nothing: standard output encoding can be set through the `PYTHONIOENCODING` environment variable. Potentially a warning message could invite users to set it.

Additionally I found this interesting reading about the subject: http://stackoverflow.com/questions/4545661/unicodedecodeerror-when-redirecting-to-file

[...]

PR #268, opened 2018-05-03
Default to utf-8 when no encoding type found parsing dogwrap output

glasnt:
Observation: 

When using `dogwrap` (via a Python 2 `pip install`) to run commands via `cron`, in cases where the `submit_mode` is set to all, it was expected that both `stdout` and `stderr` would appear in DataDog. This was not the case. 

The event would appear, but the `stdout`/`stderr` was lost 

The output would appear correctly if the `dogwrap` command was run in the shell, just not in `cron`

Cause: 

In Python 2, the string from the `OutputReader` was being inspected for an `encoding`. If this command was run within a shell, an encoding was found (in this case, `ANSI_X3.4-1968`). However, when run from `cron`, no encoding was detected. Thus, attempting to execute a string decode (`line.decode(None)`) resulted in an error. 

In Python 3, the encoding inspection defaults to UTF-8, which is a valid encoding type. 

This meant the OutputReader would end up returning no values for `stdout`/`stderr`, but the return code and command were still available and sent onto DataDog for client displaying

Solution: 

In cases where an encoding type is not identified, default to UTF-8

This will ensure the type is never None, and the default value is used. 

An alternative solution to this PR is to not decode if no encoding type is identified, but that might cause more issues down the line.

---

Environment used for debugging and code testing: Ubuntu 16.04, Python 2.7.12/3.5.2

### DIFF
--- a/datadog/dogshell/wrap.py
+++ b/datadog/dogshell/wrap.py
@@ -65,7 +65,8 @@ class OutputReader(threading.Thread):
         '''
         for line in iter(self._out.readline, b''):
             if self._fwd_out is not None:
-                self._fwd_out.write(line.decode(self._fwd_out.encoding))
+                fwd_out_encoding = self._fwd_out.encoding or 'UTF-8'
+                self._fwd_out.write(line.decode(fwd_out_encoding))
             line = line.decode('utf-8')
             self._out_content += line
         self._out.close()


yannmh:
Thanks for the detailed description of the problem @glasnt. I believe this is related with https://github.com/DataDog/datadogpy/issues/55.

Setting the default encoding type to UTF-8 sounds reasonable to me.

PR #272, opened 2018-05-11
fix http leak

### LINKED ISSUE #271: leak with 3.6.4 and datadog.api.ServiceCheck.check
with a script like the following:
```python
from datadog import initialize as dd_init, api as dd_api
from datadog.api.constants import CheckStatus as dd_CheckStatus
import logging

HOST_NAME = 'XXXX'

def test():
    response = dd_api.ServiceCheck.check(**{'check': 'execution_state', 'host_name': HOST_NAME, 'status': 0})
    assert response['status'] == 'ok', "Response failed: {}".format(response)

def main():
    logging.basicConfig(level=logging.INFO)
    logger = logging.getLogger()

    dd_init(DATADOG_KEY, host_name=HOST_NAME)

    while True:
        test()


if __name__ == '__main__':
    main()
```

memory usage will continue creeping up.  Using tracemalloc you'll find memory will be attributed to a stack like the following:
```
      File "/Users/amohr/.pyenv/versions/3.6.4/lib/python3.6/site-packages/datadog/api/resources.py", line 370
        return APIClient.submit(method, name, body)
      File "/Users/amohr/.pyenv/versions/3.6.4/lib/python3.6/site-packages/datadog/api/service_checks.py", line 42
        return super(ServiceCheck, cls)._trigger_action('POST', 'check_run', **body)
      File "/Users/amohr/Library/Preferences/PyCharm2017.3/scratches/re_leak.py", line 15
        response = dd_api.ServiceCheck.check(**{'check': 'CalendarSyncer.execution_state', 'host_name': HOST_NAME, 'status': 0})
      File "/Users/amohr/Library/Preferences/PyCharm2017.3/scratches/re_leak.py", line 30
        test()
      File "/Users/amohr/Library/Preferences/PyCharm2017.3/scratches/re_leak.py", line 36
        main()
```

I traced it to: datadog.api.http_client:RequestClient.request missing a `s.close()`.  It appears if you don't close a requests session it will not go away :(

thehesiod:
fixes #271


datadog/api/http_client.py
```diff
@@ -94,6 +94,8 @@ def request(cls, method, url, headers, params, data, timeout, proxies, verify, m
                 u"Datadog's usage. We recommand upgrading it ('pip install -U requests')."
                 u"If you need help or have any question, please contact support@datadoghq.com"
             )
+        finally:
```
truthbk: Although this is 100% fine, I think I would prefer if we could use `requests` context manager for sessions...

we can basically wrap the entire method with:

```python
with requests.Session() as s:
    http_adapter = requests.adapters.HTTPAdapter(max_retries=max_retries)
    s.mount('https://', http_adapter)
    
    # REST OF METHOD...
```

If you don't want to update that, just let me know as this is blatant nitpicking, and we can merge this as is!

thehesiod:
btw, for future reference, with help I found one of the leaks was due to https://github.com/python/cpython/pull/4587, because this would cause a warning of unclosed socket

datadog/api/http_client.py
thehesiod: updated

thehesiod: let me know what you think

thehesiod:
failures don't seem related to my change

yannmh:
Thanks @thehesiod !

PR #279, opened 2018-06-08
Append all statsd client level tags to status checks.

marshallbrekka:
This makes the python statsd client behavior consistent with the ruby and
golang statsd clients, which also include the client level tags on service
checks.


yannmh:
Looks great to me. Thanks for making the change 🙇

PR #294, opened 2018-09-07
Add note about the Agent running for StatsD

burningion:
Customer got stuck trying to configure StatsD, because they didn't realize the Agent needed to be running in order to get library working.

Adds one line note, referring back to documentation where we show how StatsD works, along with further instructions for configuration.


luck02:
:+1: 

I'm just in the process of adding instrumentation to a monitoring tool and there's no mention of what is required to get this running in this library.  If anything I'd want to expand on this some more and describe how you'd get this running in dev (IE configs for a container to provide this functionality.

PR #295, opened 2018-09-11
Make start time optional for downtime creation

elock37:
The current API does not require `start` per the [API Reference](https://docs.datadoghq.com/api/?lang=python#schedule-monitor-downtime). It appears this is noted in the current code by the `default=None` inclusion. I've added the `--` to make the arg optional.


elock37:
Incidentally included a test commit to return only IDs. Will open a clean PR to address the default behavior for the start time argument.

PR #304, opened 2018-10-25
Remove simple json as a required dependency. Install only if json is …

### LINKED ISSUE #265: Remove 'simplejson' as a required dependency
Hi. I just ran into the same problem outlined in this issue: https://github.com/requests/requests/issues/3052
> I've recently run into an issue where another library installed simple-json and because the requests library defaults to it if available, it caused all of our json requests to fail due to a decoding problem.

My opinionated summary of that thread is that in 2018, there's no reason to use `simplejson` anymore because:
* If people cared about performance, they'd use [ujson](https://docs.micropython.org/en/latest/pyboard/library/ujson.html)
* Python versions that didn't have `json` in the standard lib have long fallen off support grids
* simple is better than complex

Since `datadog` [already has logic to fall back](https://github.com/DataDog/datadogpy/blob/07ca80a5eb268306dd4cb5a4fb748bd8e94401ac/datadog/util/compat.py#L102) to the standard lib's `json`, it could simply remove `simplejson` from `setup.py` (probably for Python 2 too) and document that `simplejson` will be used if it's installed.

https://github.com/DataDog/datadogpy/blob/master/setup.py#L12

What do you think? Removing simplejson could simplify this project

alexpjohnson:
…not included in the stdlib

Closes #265

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,10 +14,11 @@ install_reqs = [
     "requests>=2.6.0",
 ]
 
-if sys.version_info[0] == 2:
-    install_reqs.append("simplejson>=2.0.9")
-else:
-    install_reqs.append("simplejson>=3.0.0")
+try:
+    import json
+except ImportError:
+    if sys.version_info[0] == 2:
+        install_reqs.append("simplejson>=2.0.9")
 
 if [sys.version_info[0], sys.version_info[1]] < [2, 7]:
     install_reqs.append("argparse>=1.2")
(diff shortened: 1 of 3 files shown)


yannmh:
Thanks a lot @alexpjohnson. 
I forked your branch and added an extra commit to remove all references of `simplejson`. As stated in #265 we don't have anymore reason to use it.

yannmh:
Closing this PR in favor of https://github.com/DataDog/datadogpy/pull/309

PR #305, opened 2018-10-26
Alias `dog` script names as `dogshell`

dwminer:
/usr/bin/dog conflicts with the existing sheepdog package in common
Linux distributions. dogshell seems like a sensible choice since
it's already called that in the source module.

### DIFF
--- a/datadog/dogshell/__init__.py
+++ b/datadog/dogshell/__init__.py
@@ -3,6 +3,8 @@
 # Copyright 2015-Present Datadog, Inc
 # stdlib
 import os
+import warnings
+import sys
 
 # 3p
 import argparse
@@ -28,6 +30,9 @@ from datadog.util.config import get_version
 
 
 def main():
+    if sys.argv[0].endswith("dog"):
+        warnings.warn("dog is pending deprecation. Please use dogshell instead.", PendingDeprecationWarning)
+
     parser = argparse.ArgumentParser(description="Interact with the Datadog API",
                                      formatter_class=argparse.ArgumentDefaultsHelpFormatter)
     parser.add_argument('--config', help="location of your dogrc file (default ~/.dogrc)",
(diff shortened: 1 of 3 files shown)


dabcoder:
@dwminer Thanks for this PR and sorry for the long delay in addressing it. This change would not be backward compatible as it would require users to update their scripts/applications/wherever `dog` and `dogwrap` are used basically. 
Would aliasing any of the `dog` commands be an alternative solution for this issue?

dabcoder:
Closing this for now but let us know if you have additional feedback to provide.

Conan-Kudo:
@dabcoder Unfortunately, sheepdog is still actively used by libvirt, so it won't be changing anytime soon. Aliases don't help, because the problem is that datadog is trying to use the `/usr/bin/dog` file, occupied by sheepdog for _years_.

jd:
This looks a bit too drastic. This is mostly a packaging issue and it can be easily be solved by distributions by diverting (e.g. see `dpkg-divert`, by not shipping our `dog` command or renaming it to `dogshell` themselves.

I'd suggest adding `dogshell`, but not removing`dog` as a first step. Distros are free to ignore the old `dog` command and not ship it if they don't want too, but regular users installing via `pip` won't be lost.

dwminer:
@jd Sure, what we're already doing in the Fedora packaging is renaming `dog` to `dogshell`. That works, and it'd be a slight step-up to also have that binary shipped by default. It's not ideal for users installing via pip to potentially have `dog` from sheepdog overridden in the search path, though. @dabcoder would it make sense to deprecate the dog command now and remove it in some future release?

dwminer:
Done, @jirikuncar. Thanks for the feedback.

jirikuncar:
@dwminer the important difference is that `dog` and `dogshell` `are not deprecated at the moment`, hence `Pending...`.

zippolyte:
Unrelated test failures, fixed on master, merging

PR #306, opened 2018-10-30
Completely include tests/ in PyPI tarball

dotlambda:
As of now, `tests/unit/dogstatsd/fixtures/route` is missing.

Follow-up to https://github.com/DataDog/datadogpy/pull/259.


yannmh:
Thank you.

PR #308, opened 2018-11-01
Add support for distributions to threadstats and the API

jbarciauskas:
This PR adds support for the distribution metrics API, which is very similar to the metrics API except it accepts a list of points per timestamp rather than a single point. Also add support in the threadstats library to accumulate distribution values and send them on flush.

### DIFF
--- a/datadog/api/metrics.py
+++ b/datadog/api/metrics.py
@@ -76,7 +76,6 @@ class Metric(SearchableAPIResource, SendableAPIResource, ListableAPIResource):
 
         return rec_parse(points_lst)
 
-
     @classmethod
     def list(cls, from_epoch):
         """
@@ -126,7 +125,8 @@ class Metric(SearchableAPIResource, SendableAPIResource, ListableAPIResource):
     @classmethod
     def send_dist(cls, metrics=None, **single_metric):
         """
-        Submit a distribution metric or a list of distribution metrics to the distribution metric API
+        Submit a distribution metric or a list of distribution metrics to the distribution metric
+        API
 
         :param metric: the name of the time series
         :type metric: string
--- a/datadog/api/metrics.py
+++ b/datadog/api/metrics.py
@@ -1,5 +1,6 @@
 # stdlib
 import time
+from collections import Iterable
 from numbers import Number
 
 # datadog
@@ -49,7 +50,7 @@ class Metric(SearchableAPIResource, SendableAPIResource, ListableAPIResource):
                 # Distributions contain a list of points
                 else:
                     timestamp = point[0]
-                    if isinstance(point[1], list):
+                    if isinstance(point[1], Iterable):
                         value = [float(p) for p in point[1]]
                     else:
                         value = float(point[1])
--- a/datadog/api/metrics.py
+++ b/datadog/api/metrics.py
@@ -16,6 +16,7 @@ class Metric(SearchableAPIResource, SendableAPIResource, ListableAPIResource):
     _METRIC_QUERY_ENDPOINT = 'query'
     _METRIC_SUBMIT_ENDPOINT = 'series'
     _METRIC_LIST_ENDPOINT = 'metrics'
+    _METRIC_DIST_ENDPOINT = 'distribution_points'
 
     @classmethod
     def _process_points(cls, points):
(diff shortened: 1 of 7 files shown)


datadog/threadstats/base.py
```diff
@@ -289,7 +302,7 @@ def flush(self, timestamp=None):
             self._is_flush_in_progress = True
 
             # Process metrics
-            metrics = self._get_aggregate_metrics(timestamp or time())
+            (metrics, dists) = self._get_aggregate_metrics_and_dists(timestamp or time())
```
jbarciauskas: this is a bit ugly but not sure if there's a better approach

jbarciauskas:
Tested using this script: https://gist.github.com/jbarciauskas/059d319a5ec118840537df5bf3018028

datadog/threadstats/base.py
```diff
@@ -206,6 +208,19 @@ def histogram(self, metric_name, value, timestamp=None, tags=None, sample_rate=1
             self._metric_aggregator.add_point(metric_name, tags, timestamp or time(), value,
                                               Histogram, sample_rate=sample_rate, host=host)
 
+    def distribution(self, metric_name, value, timestamp=None, tags=None, sample_rate=1, host=None):
+        """
+        Sample a distirbution value. Distributions will produce metrics that
```
mstepniowski: `s/distirbution/distribution`

datadog/threadstats/base.py
```diff
@@ -350,8 +374,11 @@ def _get_aggregate_metrics(self, flush_time=None):
                 'tags': metric_tags,
                 'interval': interval
             }
-            metrics.append(metric)
-        return metrics
+            if metric_type != MetricType.Distribution:
+                metrics.append(metric)
+            else:
+                dists.append(metric)
```
mstepniowski: [nit] This condition should be reversed.

datadog/threadstats/base.py
chuck: "distribution"

datadog/threadstats/base.py
```diff
@@ -298,6 +313,14 @@ def flush(self, timestamp=None):
             else:
                 log.debug("No metrics to flush. Continuing.")
 
+            count_dists = len(dists)
+            if count_dists:
+                self.flush_count += 1
```
chuck: Maybe track metric and distribution flushes separately instead of combining the counts?

datadog/api/metrics.py
```diff
@@ -42,8 +43,19 @@ def rec_parse(points_lst):
...
+                # Distributions contain a list of points
+                else:
+                    timestamp = point[0]
+                    if isinstance(point[1], list):
+                        float_points = []
+                        for p in point[1]:
+                            float_points.append(float(p))
+                        value = float_points
+                    else:
+                        value = float(point[1])
```
mstepniowski: Better to use list comprehensions here and relax type requirements a bit:

```python
from collections import Iterable

timestamp = point[0]
# more flexible - allows clients to use iterables other than the built-in list
if isinstance(point[1], Iterable):
    # more terse and faster
    value = [float(p) for p in point[1]]
else:
    value = float(point[1])
```

Would like to see some tests as well.

Also, this is outside of the scope of this PR but we should convert `rec_parse` to a non-recursive function for simplicity, speed and memory savings.

datadog/threadstats/base.py
mstepniowski: Looks OK to me. Parens around `(metrics, dists)` are not needed.

```suggestion
            metrics, dists = self._get_aggregate_metrics_and_dists(timestamp or time())
```

datadog/threadstats/base.py
jbarciauskas: it includes events so going to keep it

yannmh:
The Python client currently follows this interface: `api.$Resource.$action` where resources are the different REST resources the API can manage, and action are the CRUD methods. For example
`api.Metric.send`, `api.Monitor.update`, `api.Host.get` etc …
This means we have a new endpoint = a new resource.
If we follow the same logic, adding support for DistributionMetric means having something like
`api.DistributionMetric.send` separate from `api.Metric.send`. 
Do you see any issue with that @jbarciauskas ? If you're ok, our team can take over and make the required changes.

datadog/api/metrics.py
```diff
@@ -106,6 +116,39 @@ def send(cls, metrics=None, **single_metric):
...
+    @classmethod
+    def send_dist(cls, metrics=None, **single_metric):
+        """
+        Submit a distribution metric or a list of distribution metrics to the distribution metric
+        API
+
+        :param metric: the name of the time series
+        :type metric: string
+
+        :param points: a (timestamp, value) pair or list of (timestamp, value) pairs
```
yannmh: @jbarciauskas confirming, this should be `(timestamp, [list of values]) or list of (timestamp, [list of values])` right?

jbarciauskas: Yep

yannmh:
Closed by https://github.com/DataDog/datadogpy/pull/312. Thank you!

PR #312, opened 2018-11-07
 Add support for distributions to threadstats and the API

yannmh:
Based on https://github.com/DataDog/datadogpy/pull/308. 
Move distributions to its own `api.Distribution` resource.

Thanks again @jbarciauskas


yannmh:
Correct.

PR #322, opened 2018-11-23
add monitor subcommands that work from file

Hefeweizen:
These subcommands read a filename from the command line, and then use
specified file for all values.  These features allow a workflow of `dog
monitor show <id> > monitor_id.json && dog monitor fupdate
monitor_id.json`.  This workflow allows version control of all monitors,
as well as repo-driven updates.

### DIFF
--- a/datadog/dogshell/monitor.py
+++ b/datadog/dogshell/monitor.py
@@ -1,4 +1,5 @@
 # stdlib
+import argparse
 import json
 
 # 3p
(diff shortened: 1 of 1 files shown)


yannmh:
This is very helpful. Thank you @Hefeweizen

Hefeweizen:
@yannmh Thanks; I appreciate the review.

I did mvp for my needs and only got this working with monitors.  I think it'd be nice if this capability (load from file) were available for all object types (e.g. dashboards, &c).  Should I piecemeal this in on each subcommand, or is there something architectural that could be done?

yannmh:
I agree this would also be great for other resources. 
Adding a `fpost` and `fupdate` to all of them doesn't feel right ... instead I'd like to explore the possibility of re-using the existing "post" and "update" subcommands and maybe some extra parameter to specific that the we intend to pass content as json, e.g. something like
```
cat my_monitor.json | dog monitor update --json
```

Hefeweizen:
@yannmh  I chose fupdate/fpost because I just wanted traction on the idea and to get something started without having to build too much argument handling into the existing _post.  For better integration, it seems something like this may work:

```
    def _post(cls, args):
       if args.json:
          _post_stdin()
       else:
         _post_from_args()

    def _post_from_args(cls, args):
        """
        previously, just `_post()`
        """
        api._timeout = args.timeout
        format = args.format
        options = None
        ....

    def _post_stdin(cls, args):
       monitor = json.load(sys.stdin)
       ....
```

This provides the capability locally, but we'd have to make the changes across every subcommand (monitors, dashboards, &c).  Is there anything we can do structurally to provide the capability across all of them in one swoop?

Issue #223, opened 2017-08-11
API client can't handle 'ascii' codec.


axelabs:
I've created a monitor that has a '½' in its name.  When calling api.Monitor.unmute on it's ID when the state is already not muted, I get the following errors:
```
Traceback (most recent call last):
  File "<stdin>", line 62, in <module>
  File "/your_mums_venv/lib/python2.7/site-packages/datadog/api/monitors.py", line 88, in unmute
    return super(Monitor, cls)._trigger_class_action('POST', 'unmute', id, **params)
  File "/your_mums_venv/lib/python2.7/site-packages/datadog/api/resources.py", line 193, in _trigger_class_action
    return APIClient.submit(method, cls._class_url + "/" + str(id) + "/" + name, params)
  File "/your_mums_venv/lib/python2.7/site-packages/datadog/api/api_client.py", line 173, in submit
    log.error(str(error))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xbd' in position 74: ordinal not in range(128)
```
_Tested with python 2.7.5 and datadog==0.16.0._

Tyler-Churchill:
Encountered this as well.
Stripe had a hiccup and gave back `"APIError(u"There was an unexpected error processing this request\u2013we've been notified of this issue. We'll update our status page at https://status.stripe.com/ with more information if we're seeing systemic issues.",)"`

Which caused `UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 2812: ordinal not in range(128)`

yannmh:
Fixed with https://github.com/DataDog/datadogpy/pull/346

synth:
This issue occurs for me with dogwrap v0.28.0

<img width="907" alt="Screenshot 2019-04-20 11 29 26" src="https://user-images.githubusercontent.com/10546/56460943-91c28c80-635f-11e9-9900-9da40cdd2837.png">

synth:
@yannmh I can reliably reproduce this whenever there is unicode character in my log output

synth:
Created test case and more detail in #371 :)

yannmh:
Thank you @synth we are looking at the issue.

PR #346, opened 2019-01-15
Remove redundant string conversion to prevent UnicodeEncodeError

chrismdd:
This PR removes a string conversion that causes https://github.com/DataDog/datadogpy/issues/223. It is reasonable to assume that the values of the `errors` Array are strings, so this conversion is unnecessary, and it causes a `UnicodeEncodeError` in Python 2 if the API response contains non-ASCII characters.

Tested with Python 2.7.15 and 3.7.2.


chrismdd:
@yannmh thanks for the review. Could you please merge this too (I don't have write access)

Issue #212, opened 2017-06-20
not all exceptions being caught


jsamuel:
I see cases where exceptions aren't being caught from calls to `increment()` and other stats reporting functions. According to the code, it's only catching `socket.error` but there are other errors that can happen, at least in multithreaded applications.

```
  File "/opt/python3.6/lib/python3.6/site-packages/datadog/dogstatsd/base.py", line 153, in increment
    self._report(metric, 'c', value, tags, sample_rate)
  File "/opt/python3.6/lib/python3.6/site-packages/datadog/dogstatsd/base.py", line 262, in _report
    self._send(encoded)
  File "/opt/python3.6/lib/python3.6/site-packages/datadog/dogstatsd/base.py", line 267, in _send_to_server
    (self.socket or self.get_socket()).send(packet.encode(self.encoding))
AttributeError: 'NoneType' object has no attribute 'send'
```

I think you should be catching every exception and never letting an exception escape from your code. Otherwise, these types of bugs in dogstats can cause bugs in people's applications if they are using dogstats. As it stands, it looks like the only safe usage appears to be to write a wrapper around dogstats calls to make sure they can never raise exceptions.

jbayardo:
Hi. Any plans on fixing this? I am getting the same exception from one of our core apps

melinath:
Hey, just encountered this as well. Are there any plans to fix this?

dabcoder:
Yes indeed, the linked PR should address this issue. Thanks for reporting.

PR #349, opened 2019-01-25
Handle more socket exceptions

dabcoder:
Attempts to fix https://github.com/DataDog/datadogpy/issues/212.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -7,6 +7,7 @@ from random import random
 import logging
 import os
 import socket
+from threading import Lock
 
 # datadog
 from datadog.dogstatsd.context import TimedContextManagerDecorator
@@ -58,6 +59,8 @@ class DogStatsd(object):
         UDP. If set, disables UDP transmission (Linux only)
         :type socket_path: string
         """
+
+        self.lock = Lock()
 
         # Connection
         if socket_path is not None:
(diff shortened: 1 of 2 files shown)


datadog/dogstatsd/base.py
```diff
@@ -289,9 +289,12 @@ def _send_to_server(self, packet):
         except socket.timeout:
             # dogstatsd is overflowing, drop the packets (mimicks the UDP behaviour)
             return
-        except socket.error:
+        except (socket.error, socket.herror, socket.gaierror) as se:
             log.info("Error submitting packet, dropping the packet and closing the socket")
```
gzussa: What about changing this to a `warning` instead?

datadog/dogstatsd/base.py
```diff
@@ -289,9 +289,12 @@ def _send_to_server(self, packet):
...
             self.close_socket()
+        except Exception as e:
+            log.info("Unexpected error: ", e)
```
gzussa: Actually maybe sending `error` would be more appropriate.

datadog/dogstatsd/base.py
```diff
@@ -289,9 +289,12 @@ def _send_to_server(self, packet):
         except socket.timeout:
             # dogstatsd is overflowing, drop the packets (mimicks the UDP behaviour)
             return
-        except socket.error:
+        except (socket.error, socket.herror, socket.gaierror) as se:
             log.info("Error submitting packet, dropping the packet and closing the socket")
             self.close_socket()
```
nmuesch: Should we be closing the socket in the second `except` clause as well?

dabcoder: Not sure, `# If set, use socket directly`, would the last exception cover the case of not using the socket and thus not needing to close it? Or am I on the wrong track there?

datadog/dogstatsd/base.py
```diff
@@ -289,9 +293,12 @@ def _send_to_server(self, packet):
         except socket.timeout:
             # dogstatsd is overflowing, drop the packets (mimicks the UDP behaviour)
             return
-        except socket.error:
-            log.info("Error submitting packet, dropping the packet and closing the socket")
+        except (socket.error, socket.herror, socket.gaierror) as se:
+            log.warning("Error submitting packet: {}, dropping the packet and closing the socket".format(str(se)))
```
ofek:
```suggestion
            log.warning("Error submitting packet: {}, dropping the packet and closing the socket".format(se))
```

PR #353, opened 2019-02-07
DD Initialize now can update the statsd object default namespace

IceS2:
Added a new parameter to the Initialize method in order to update the statsd object default namespace.


IceS2:
The Python 3.4 Travis Build failed because it was unable to download:
Unable to download 3.4 archive. The archive may not exist. Please consider a different version.

PR #355, opened 2019-02-20
add example of proxy config to docs

cswatt:
Just adds a line to spell out how proxies should be configured. This was requested by a customer.

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -41,7 +41,7 @@ def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
     :param app_key: Datadog application key
     :type app_key: string
 
-    :param proxies: Proxy to use to connect to Datadog API; \
+    :param proxies: Proxy to use to connect to Datadog API;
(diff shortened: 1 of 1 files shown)


datadog/__init__.py
```diff
@@ -41,9 +41,9 @@ def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
     :param app_key: Datadog application key
     :type app_key: string
 
-    :param proxies: Proxy to use to connect to Datadog API
+    :param proxies: Proxy to use to connect to Datadog API; \
```
ofek: I don't know how backslashes render, you can just do:
```suggestion
    :param proxies: Proxy to use to connect to Datadog API;
```

datadog/__init__.py
```diff
@@ -41,9 +41,9 @@ def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
...
+        for example, 'proxies': {'http': "http:<user>:<pass>@<ip>:<port>/"}
```
ofek:
```suggestion
                    for example, 'proxies': {'http': "http:<user>:<pass>@<ip>:<port>/"}
```

PR #356, opened 2019-02-21
Update to dogshell monitor _post method - tag ingestion

bradleyjay:
Modified datadogpy/datadog/dogshell/monitor.py:

- Added --tags to post_parser.add_argument
- Added tag ingestion via args.tags to _post method 
- Added tags to _post method res object

### DIFF
--- a/datadog/dogshell/monitor.py
+++ b/datadog/dogshell/monitor.py
@@ -104,7 +104,7 @@ class MonitorClient(object):
                 raise Exception('bad json parameter')
 
         if args.tags:
-            tags = sorted(set([t.strip() for t in args.tags.split(',') if t]))
+            tags = sorted(set([t.strip() for t in args.tags.split(',') if t.strip()]))
         else:
             tags = None
 
--- a/datadog/dogshell/monitor.py
+++ b/datadog/dogshell/monitor.py
@@ -104,7 +104,7 @@ class MonitorClient(object):
                 raise Exception('bad json parameter')
 
         if args.tags:
-            tags = sorted(set([t.strip() for t in 
+            tags = sorted(set([t.strip() for t in
                                args.tags.split(',') if t]))
         else:
             tags = None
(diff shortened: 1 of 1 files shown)


datadog/dogshell/monitor.py
```diff
@@ -101,8 +102,14 @@ def _post(cls, args):
                 options = json.loads(args.options)
             except:
                 raise Exception('bad json parameter')
+
+        if args.tags:
+            tags = sorted(set([t.strip() for t in args.tags.split(',') if t]))
```
platinummonkey: I’d check for truth status with the .strip as well.

platinummonkey: Bump @bradleyjay

bradleyjay: @platinummonkey Sorry for the delay, thank you for putting eyes on this 🙇

Sure! So we do this in a couple of places in Dogshell, I got this straight from metrics.py:
https://github.com/DataDog/datadogpy/blob/9bb02535c119a90c60399423795743e763cec6c2/datadog/dogshell/metric.py#L46-L48

When you say check for truth status, can you clarify? The one hole I see is if args.tags is something other than a string, though I'm not sure the argparser allows that. But for safety with .strip(), we could swap line 107 to read:

`if args.tags **and type(args.tags) == str:`**

Otherwise, I think we're ok, but I could be misunderstanding - `if t` will evaluate as False if `t` is an empty string, and `if args.tags` will do the same. Either way, if args.tags simply doesn't exist, we're in trouble anyway. If there are no tags, `t.strip` is happy with any string.

Should this be something we change overall with Dogshell? I can make those changes as well, if so. How would you recommend changing it?

platinummonkey:
```python
tags = sorted(set([t.strip() for t in args.tags.split(',') if t.strip()]))
```

because if `t == " "` then `t.strip()` will yield `t == ""` as an added item to the set. when in reality that should not be added.

bradleyjay: Good god you're right. And it slips right past a str( ) type check. 

Thanks Cody, adding.

KeriPMiller:
Hey @bradleyjay Looks like you can merge this PR. Let me know when you're g2g and I can notify the customer 🎉

platinummonkey:
cc @masci maybe for merge rights?

PR #363, opened 2019-03-15
add env vars support for dogstatsd host, port and entity id

ahmed-mez:
Add environment variables support for Client configuration

Support of 3 environment variables for Client configuration:

`"DD_AGENT_HOST"` used to provide the Agent hostname.
`"DD_DOGSTATSD_PORT"` used to provide the DogStatsD port.
`"DD_ENTITY_ID"` used to provide an identifier to the Client.

If Client constructor's arguments are not set, prioritize `"DD_AGENT_HOST"` and `"DD_DOGSTATSD_PORT"` over default values.
Add `"DD_ENTITY_ID"` to tags with the tag name `dd.internal.entity_id`.

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -89,17 +89,21 @@ def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
         statsd.host = None
         statsd.port = None
     else:
-        # Check host and port env vars, override arguments if env vars exist
-        DD_AGENT_HOST = os.environ.get('DD_AGENT_HOST', '')
-        if DD_AGENT_HOST != '':
-            statsd_host = DD_AGENT_HOST
+        # If statsd_host argument is not set, try to get host from DD_AGENT_HOST env var
+        if statsd_host is None:
+            DD_AGENT_HOST = os.environ.get('DD_AGENT_HOST', '')
+            if DD_AGENT_HOST != '':
+                statsd_host = DD_AGENT_HOST
 
-        DD_DOGSTATSD_PORT = os.environ.get('DD_DOGSTATSD_PORT', '')
-        if DD_DOGSTATSD_PORT != '':
-            try:
-                statsd_port = int(DD_DOGSTATSD_PORT)
-            except ValueError:
-                pass
+        # If statsd_port argument is not set, try to get port from DD_DOGSTATSD_PORT env var
+        if statsd_port is None:
+            DD_DOGSTATSD_PORT = os.environ.get('DD_DOGSTATSD_PORT', '')
+            if DD_DOGSTATSD_PORT != '':
+                try:
+                    statsd_port = int(DD_DOGSTATSD_PORT)
+                except ValueError:
+                    pass
+        
         if statsd_host or statsd_use_default_route:
             statsd.host = statsd.resolve_host(statsd_host, statsd_use_default_route)
         if statsd_port:
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -47,6 +47,12 @@ def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
     :param api_host: Datadog API endpoint
     :type api_host: url
 
+    :envvar DD_AGENT_HOST: the host of the DogStatsd server.
+    :type DD_AGENT_HOST: string
+
+    :envvar DD_DOGSTATSD_PORT: the port of the DogStatsd server.
+    :type DD_DOGSTATSD_PORT: integer
+
     :param statsd_host: Host of DogStatsd server or statsd daemon
     :type statsd_host: address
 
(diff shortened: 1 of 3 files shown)


datadog/__init__.py
```diff
@@ -83,6 +89,17 @@ def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
         statsd.host = None
         statsd.port = None
     else:
+        # Check host and port env vars, override arguments if env vars exist
```
ofek: Please give arguments precedence as it says in the PR description.

datadog/dogstatsd/base.py
```diff
@@ -62,6 +73,20 @@ def __init__(self, host='localhost', port=8125, max_buffer_size=50, namespace=No
...
+
+        DD_DOGSTATSD_PORT = os.environ.get('DD_DOGSTATSD_PORT', '')
+        if DD_DOGSTATSD_PORT != '' and port == 8125:
+            try:
+                port = int(DD_DOGSTATSD_PORT)
+            except ValueError:
+                log.warning("Port number provided in DD_DOGSTATSD_PORT env var is not an integer: \
+                %s, using %s as port number", DD_DOGSTATSD_PORT, port)
+                pass
+
```
ofek: We may have to set the default arguments to `None` to detect actual args. @yannmh wdyt?

ahmed-mez: so should I set the default arguments to `None` ? @ofek @yannmh

datadog/__init__.py
```diff
@@ -83,6 +89,21 @@ def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
...
+
+        # If statsd_port argument is not set, try to get port from DD_DOGSTATSD_PORT env var
+        if statsd_port is None:
+            DD_DOGSTATSD_PORT = os.environ.get('DD_DOGSTATSD_PORT', '')
+            if DD_DOGSTATSD_PORT != '':
+                try:
+                    statsd_port = int(DD_DOGSTATSD_PORT)
+                except ValueError:
+                    pass
+
```
yannmh: I don't think you need these. The "already instantiated" statsd client, would have had those values set in the constructor below.

datadog/dogstatsd/base.py
yannmh: I don't have a strong opinion on this. If we decide to keep the default args, I'd name the default argument, so it becomes
```python
        # Check host and port env vars
        DD_AGENT_HOST = os.environ.get('DD_AGENT_HOST', '')
        if DD_AGENT_HOST and host == DEFAULT_HOST:
            host = DD_AGENT_HOST
```

Other notes:
1. is it a "conventional" to name Python variables that store env variable with UPPERCASE? 
2 `os.environ.get('DD_AGENT_HOST', '')` -> `os.environ.get('DD_AGENT_HOST')`

datadog/dogstatsd/base.py
```diff
@@ -83,6 +108,11 @@ def __init__(self, host='localhost', port=8125, max_buffer_size=50, namespace=No
         if constant_tags is None:
             constant_tags = []
         self.constant_tags = constant_tags + env_tags
+        ENTITY_ID_TAG_NAME = "dd.internal.entity_id"
```
yannmh: this constant should be moved outside of the constructor

datadog/dogstatsd/base.py
```diff
@@ -62,6 +73,20 @@ def __init__(self, host='localhost', port=8125, max_buffer_size=50, namespace=No
...
+            host = DD_AGENT_HOST
+
+        DD_DOGSTATSD_PORT = os.environ.get('DD_DOGSTATSD_PORT', '')
+        if DD_DOGSTATSD_PORT != '' and port == 8125:
+            try:
+                port = int(DD_DOGSTATSD_PORT)
+            except ValueError:
+                log.warning("Port number provided in DD_DOGSTATSD_PORT env var is not an integer: \
+                %s, using %s as port number", DD_DOGSTATSD_PORT, port)
+                pass
```
yannmh: I don't think you need the 'pass'

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -97,6 +97,15 @@ def test_initialization(self):
...
         t.assert_equal(statsd.port, 8125)
 
+        # Set `statsd` host and port using env vars
+        with preserve_environment_variable('DD_AGENT_HOST'):
+            os.environ['DD_AGENT_HOST'] = 'myenvvarhost'
+            with preserve_environment_variable('DD_DOGSTATSD_PORT'):
+                os.environ['DD_DOGSTATSD_PORT'] = '4321'
+                initialize()
+        t.assert_equal(statsd.host, "myenvvarhost")
+        t.assert_equal(statsd.port, 4321)
```
yannmh: can we make this a separate test?

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -573,6 +582,32 @@ def test_tags_from_environment_and_constant(self):
...
+    def test_entity_tag_and_tags_from_environment_and_constant(self):
+        with preserve_environment_variable('DATADOG_TAGS'):
+            os.environ['DATADOG_TAGS'] = 'country:china,age:45,blue'
+            with preserve_environment_variable('DD_ENTITY_ID'):
+                os.environ['DD_ENTITY_ID'] = '04652bb7-19b7-11e9-9cc6-42010a9c016d'
+                statsd = DogStatsd(constant_tags=['country:canada', 'red'])
+        statsd.socket = FakeSocket()
+        statsd.gauge('gt', 123.4)
+        t.assert_equal('gt:123.4|g|#country:canada,red,country:china,age:45,blue,dd.internal.entity_id:04652bb7-19b7-11e9-9cc6-42010a9c016d', statsd.socket.recv())
+
```
yannmh: 👍

tests/unit/dogstatsd/test_statsd.py
ahmed-mez: of course 👍

ahmed-mez:
> Added a couple more comments.

yannmh:
Looks good to me.

PR #368, opened 2019-04-08
No annoying hostname output

brendanlong:
This fixes an issue we're seeing in ECS, where every time a new container starts, we get something like this printed to stderr:

> hostname: 91d9cd0c1b70: Host not found

Since Datadog's log UI treats everything on stderr as ERROR level by default, that means we get output like this (for a job that runs every minute):

Screenshot from 2019-04-08 13-06-12

I have no idea _why_ `hostname` is printing this error, but I can confirm that this change fixes it.

### DIFF
--- a/datadog/util/hostname.py
+++ b/datadog/util/hostname.py
@@ -70,7 +70,7 @@ def get_hostname():
         def _get_hostname_unix():
             try:
                 # try fqdn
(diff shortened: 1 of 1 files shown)


brendanlong:
Note that on line 81 we ignore all errors

PR #370, opened 2019-04-12
fix(threadstats): ensure MetricsAggregator is threadsafe

TheKevJames:
As described in #142, the MetricsAggregator is not threadsafe under Python 3. This is a regression caused by a change in how the `.keys()` method works under Python 2 vs Python 3: in Python 3, calling `.keys()` returns a view rather than a copy of a list.

The solution here is to use a reentry lock around the modifiers and accesses to `self._metrics`, in order to ensure that any concurrent calls to `MetricsAggregator.flush()` and `MetricsAggregator.add_point()` have no race conditions.

Note that the stack trace which this changeset fixes is:
```
RuntimeError: dictionary changed size during iteration
  File "datadog/threadstats/base.py", line 307, in flush
    metrics, dists = self._get_aggregate_metrics_and_dists(timestamp or time())
  File "datadog/threadstats/base.py", line 348, in _get_aggregate_metrics_and_dists
    rolled_up_metrics = self._metric_aggregator.flush(flush_time)
  File "datadog/threadstats/metrics.py", line 173, in flush
    past_intervals = [i for i in self._metrics.keys() if i < interval]
  File "datadog/threadstats/metrics.py", line 173, in <listcomp>
    past_intervals = [i for i in self._metrics.keys() if i < interval]
```

### DIFF
--- a/datadog/threadstats/metrics.py
+++ b/datadog/threadstats/metrics.py
@@ -4,6 +4,7 @@ Metric roll-up classes.
 from collections import defaultdict
 import random
 import itertools
+import threading
 
 from datadog.util.compat import iternext
 from datadog.threadstats.constants import MetricType
@@ -152,6 +153,7 @@ class MetricsAggregator(object):
     """
 
     def __init__(self, roll_up_interval=10):
+        self._lock = threading.RLock()
         self._metrics = defaultdict(lambda: {})
         self._roll_up_interval = roll_up_interval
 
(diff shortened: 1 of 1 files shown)


edkorthof:
Thanks for the fix. But to clear -- this wasn't just unsafe in python3; in python2 it was also unsafe (but less likely to trigger); I provided a script which allows reproduction in python2.7.

PR #374, opened 2019-05-07
[dashboards] Add support for dashboard list API v2

jbenais:
- Add `DashboardList.v2.get_items`, `DashboardList.v2.add_items`, `DashboardList.v2.update_items` and `DashboardList.v2.delete_items` API calls

### DIFF
--- a/datadog/api/api_client.py
+++ b/datadog/api/api_client.py
@@ -28,7 +28,6 @@ class APIClient(object):
     _max_timeouts = _max_timeouts
     _backoff_timestamp = None
     _timeout_counter = 0
-    _api_version = _api_version
 
     # Plugged HTTP client
     _http_client = None
@@ -44,8 +43,8 @@ class APIClient(object):
         return cls._http_client
 
     @classmethod
-    def submit(cls, method, path, body=None, attach_host_name=False, response_formatter=None,
-               error_formatter=None, **params):
+    def submit(cls, method, path, api_version=_api_version, body=None, attach_host_name=False,
+               response_formatter=None, error_formatter=None, **params):
         """
         Make an HTTP API request
 
@@ -112,11 +111,10 @@ class APIClient(object):
             if isinstance(body, dict):
                 body = json.dumps(body)
                 headers['Content-Type'] = 'application/json'
-
             # Construct the URL
             url = "{api_host}/api/{api_version}/{path}".format(
                   api_host=_api_host,
-                  api_version=cls._api_version,
+                  api_version=api_version,
                   path=path.lstrip("/"),
             )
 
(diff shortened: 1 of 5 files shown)


datadog/api/resources.py
```diff
@@ -35,18 +38,19 @@ def create(cls, attach_host_name=False, method='POST', id=None, params=None, **b
             params = {}
 
         path = cls._resource_name
+        api_version = getattr(cls, '_api_version', _api_version)
```
enbashi: Isn't this redundant here and before all other calls to `submit` in this file since you are also defaulting to `v1` in the implementation of `submit` ? I think it is sufficient to default it to `v1` in `api_client.py` only.

datadog/api/api_client.py
```diff
@@ -44,8 +43,8 @@ def _get_http_client(cls):
         return cls._http_client
 
     @classmethod
-    def submit(cls, method, path, body=None, attach_host_name=False, response_formatter=None,
-               error_formatter=None, **params):
+    def submit(cls, method, path, api_version=_api_version, body=None, attach_host_name=False,
```
enbashi: let's also add `:param api_version ...` in the comments below to be consistent with the other arguments.

datadog/dogshell/dashboard_list.py
```diff
@@ -57,21 +57,47 @@ def setup_parser(cls, subparsers):
...
+            'add_dashboards_v2', help="Add dashboards to an existing dashboard list"
+        )
+        add_dashboards_v2_parser.add_argument(
+            'dashboard_list_id', help="Dashboard list to add dashboards to"
+        )
+        add_dashboards_v2_parser.add_argument(
+            'dashboards',
+            help='A JSON list of dashboard dicts, e.g. ' +
+            '[{"type": "custom_timeboard", "id": 1234}, ' +
+            '{"type": "custom_screenboard", "id": 123}]'
```
enbashi: let's use the new ID format here in the help message.

jbenais:
> Why did we choose to have `v2` as a sub variable within `DashboardList`? Would it be better to have a `DashboardListV2` such a way we can do `api.DashboardListV2` directly?

Adding a DashboardListV2 class means that we would have to do the same for all other classes that would need to be versioned and it can become very messy at some point. We also wanted to keep the name DashboardList because we want users to use the latest version of our API at the end. Having DashboardListV2 could let users think that they can use either DashboardList or DashboardListV2, which we don't want. This is why we wanted to implement a solution that would allow, in the future, to manage API updates in a simple way. With this approach, you can specify your endpoint version easily.

> Also, did we consider having something like `api.v2.DashboardList`?

In general, when it comes to versioning, adding the prefix 'v2' after 'api' means that the whole api has been updated to another version, which is not the case here since only 4 endpoints have been updated. We have therefore chosen to place 'v2' after DashboardList to avoid any confusion and also because it seems clear enough for the users.

PR #376, opened 2019-05-08
Handle http code 429 for rate limiting

ssc3:
Handle http code 429 rate limiting in external python library


ssc3:
> Can we unit test these?

These are unit tested here already: https://github.com/DataDog/datadogpy/blob/master/tests/unit/api/test_api.py

This lib already has a comprehensive set of tests.

There are no unit tests to check for http response code. That would be a great long term addition to test every single endpoint.

PR #377, opened 2019-05-13
origin detection documentation

ahmed-mez:
Document the origin detection feature introduced by https://github.com/DataDog/datadogpy/pull/363 with an example of how to set up the `DD_ENTITY_ID` env var.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,14 +73,15 @@ Thread Safety
 
 Origin detection over UDP
 -------------
-Origin detection is a method to detect which pod DogStatsD packets are coming from in order to add the pod's tags to the tag list.
+Origin detection is a method to detect which pod `DogStatsD` packets are coming from in order to add the pod's tags to the tag list.
+The `DogStatsD` client attaches an internal tag, `entity_id`. The value of this tag is the content of the `DD_ENTITY_ID` environment variable if found, which is the pod’s UID.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,3 +70,17 @@ stats.increment('home.page.hits')
...
 `DogStatsD` and `ThreadStats` are thread-safe.
+
+Origin detection over UDP
+-------------
+Origin detection is a method to detect which pod DogStatsD packets are coming from in order to add the pod's tags to the tag list.
+
+To enable origin detection over UDP, add the following lines to your application manifest
+```
+env:
+  - name: DD_ENTITY_ID
```
truthbk: There is no other mention to the `DD_ENTITY_ID` environment variable in the docs, and that would be desirable to provide some more context to the users. For instance, in the datadog-go client we have this:
```
- If the `DD_ENTITY_ID` environment variable is found, its value will be injected as a global `dd.internal.entity_id` tag. This tag will be used by the Datadog Agent to insert container tags to the metrics. You should only `append` to the `c.Tags` slice to avoid overwriting this global tag.
```

ahmed-mez: thanks!
this line ```The DogStatsD client attaches an internal tag, `entity_id`. The value of this tag is the content of the `DD_ENTITY_ID` environment variable, which is the pod’s UID.``` just below the example mention the `DD_ENTITY_ID`, I guess I can put it just before the example with context more like the `datadog-go` client description

ahmed-mez: can I get a second review ?

PR #378, opened 2019-05-17
Return Rate Limit Headers 

fdhoff:
Reference: https://docs.datadoghq.com/api/?lang=python#rate-limiting

These changes make it so the response headers returned by Datadog are passed to the user. The rate-limit headers can be useful for monitoring usage and backing off when necessary.


fdhoff:
Awesome, thank you @gzussa. 
Looks like someone with write access has to merge this as I cannot.

PR #379, opened 2019-05-28
Better string handling for dogwrap

### LINKED ISSUE #223: API client can't handle 'ascii' codec.

### LINKED ISSUE #371: Dogwrap doesn't support unicode characters
This is a follow up to #223 

There are two issues at play here:
1. Wrap.py doesn't handle subprocess output with unicode properly and causes fatal exception.
2. Wrap.py should be robust enough to handle any exception without causing fatal error so as to not cause subprocess to fail. 

I will go into details on each. 

This is in reference to dogwrap v0.28.0 on Docker Linux. The issue does not occur on Mac OSX 10.14.4. I have not tested on other OS's.
 
### Wrap.py handling of unicode characters
Test case (echo_unicode.sh):
```
echo "This is one string: øø"
echo 'A unicode \xc6\x8e string \xc3\xb1'
```

Command:
```
dogwrap -k $API_KEY -n 'test unicode' --submit_mode all './echo_unicode'
```

Exception:
```
Exception in thread Thread-1:
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/local/lib/python2.7/dist-packages/datadog/dogshell/wrap.py", line 69, in run
    self._fwd_out.write(line.decode(fwd_out_encoding))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 20: ordinal not in range(128)
```

### Wrap.py handling of exceptions

The dogwrap python script fails completely when there is a unicode character output as described above. This is extremely problematic because it causes subprocess to fail. In our case, the subprocess is DelayedJob (DJ) which seems to frequently work with database transactions. When dogwrap fails, it causes the DJ process to be killed leaving a database transaction open which causes all other DJ processes to hang wreaking utter havoc on our database .

The solution should be more robust exception handling to not let the subprocess fail - after all this is just a wrapper script. 
 
You can see in the python code that the error handling needs to be improved. The error occurs in OutputReader which is itself wrapped in a try/except block, but the except block only catches Timeout exceptions. Recommend to add additional catch on all Exception types:

https://github.com/DataDog/datadogpy/blob/bc3321558941ff04ccbfac3b8e26268b6942d4f3/datadog/dogshell/wrap.py#L112-L152

zippolyte:
- Arguments are bytes, so the join must be made on a binary string, otherwise python will try to convert and can encounter errors.
- Stop decoding lines before writing to stdout, since `write` expects bytes.
- Move decoding the output to the end of the execution, to make sure we won't raise exceptions while command is still running, which can cause unexpected behavior, like in issue #371 .

Resolves #371, closes #223

### DIFF
--- a/datadog/dogshell/wrap.py
+++ b/datadog/dogshell/wrap.py
@@ -65,8 +65,7 @@ class OutputReader(threading.Thread):
         '''
         for line in iter(self._out.readline, b''):
             if self._fwd_out is not None:
-                fwd_out_encoding = self._fwd_out.encoding or 'UTF-8'
-                self._fwd_out.write(line.decode(fwd_out_encoding, 'ignore'))
+                self._fwd_out.write(line)
             line = line.decode('utf-8')
             self._out_content += line
         self._out.close()
--- a/datadog/dogshell/wrap.py
+++ b/datadog/dogshell/wrap.py
@@ -54,7 +54,7 @@ class OutputReader(threading.Thread):
         '''
         threading.Thread.__init__(self)
         self.daemon = True
-        self._out_content = u""
+        self._out_content = ""
         self._out = proc_out
         self._fwd_out = fwd_out
 
@@ -66,7 +66,6 @@ class OutputReader(threading.Thread):
         for line in iter(self._out.readline, b''):
             if self._fwd_out is not None:
                 self._fwd_out.write(line)
-            line = line.decode('utf-8')
             self._out_content += line
         self._out.close()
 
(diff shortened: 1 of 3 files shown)


synth:
Sorry for the delay in testing, but got to it today. It works! +1 on adding tests ;-)

zippolyte:
Thanks for testing this @synth, we'll likely do a release on PyPI this week

PR #382, opened 2019-05-29
Do not send JSON body in GET requests

### LINKED ISSUE #375: api.Monitor.search() returns "400 malformed request" html
Any call to _api.Monitor.search()_ with parameters or without returns a 400 status code as response. For this call the request body is an empty JSON object('{}') rather than being _None_ (as for other 'GET' requests) and the headers accordingly contain _'Content-Type': 'application/json'_

zippolyte:
The EU instance (api.datadoghq.eu) refuses GET requests with a body (even if empty JSON dict `{}`) and responds with `400 Bad Request`. 
This PR simply set the body to `None` when making GET requests.

Fixes #375

### DIFF
--- a/tests/unit/api/helper.py
+++ b/tests/unit/api/helper.py
@@ -143,8 +143,8 @@ class DatadogAPITestCase(unittest.TestCase):
 
     def request_called_with(self, method, url, data=None, params=None):
         (req_method, req_url), others = self.request_mock.call_args()
-        self.assertEquals(method, req_method, req_method)
-        self.assertEquals(url, req_url, req_url)
+        self.assertEqual(method, req_method, req_method)
+        self.assertEqual(url, req_url, req_url)
 
         if data:
             self.assertIn('data', others)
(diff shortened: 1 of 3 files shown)


tests/unit/api/helper.py
```diff
@@ -143,8 +143,8 @@ def get_request_data(self):
 
     def request_called_with(self, method, url, data=None, params=None):
         (req_method, req_url), others = self.request_mock.call_args()
-        self.assertEquals(method, req_method, req_method)
-        self.assertEquals(url, req_url, req_url)
+        self.assertEqual(method, req_method, req_method)
```
nmuesch: 💜 on updating this to its non deprecated alias 😄

everops-alex:
Is DataDog going to re-enable this API functionality after the EU site 400 gets fixed? Is there any chance this could be re-enabled for US users?

https://github.com/DataDog/datadogpy/issues/664

PR #386, opened 2019-06-04
Do not override API parameters with default when calling initialize if they are already set

### LINKED ISSUE #369: Calling datadog.initialize >1 times can be destructive
Given the current behaviour of `datadog.initialize`, there's potential for a subsequent call to it to wipe out previous API config, which I think is unnecessary  and could create annoying bugs.

If I do: 

```
import datadog

API_KEY="thisisakey"
datadog.initialize(api_key=API_KEY)

...
# some other code somewhere later
datadog.initialize(statsd_host=os.environ.get("DOGSTATSD_HOST_IP", "localhost"))
```

My `api_key` will no longer be set because of this code -> https://github.com/DataDog/datadogpy/blob/master/datadog/__init__.py#L73

I think it would be quite straightforward to change the above code to first check if `_api_key` is already set, and not override it with the `os.environ.get` call. Then `datadog.intialize` would be more predictable :). 

More broadly, I think the mutable state in `datadog.api` and `datadog.statsd` makes initilization/config management harder to manager. 

----

**Sidenote:** I think this issue is expressing related concerns about `datadog.initialize` https://github.com/DataDog/datadogpy/issues/347

zippolyte:
Resolves #369

### DIFF
--- a/tests/unit/api/test_api.py
+++ b/tests/unit/api/test_api.py
@@ -38,25 +38,6 @@ from tests.unit.api.helper import (
 )
 
 
-def preserve_environ_datadog(func):
-    """
-    Decorator to preserve the original environment value.
-    """
-    @wraps(func)
-    def wrapper(env_name, *args, **kwds):
-        environ_api_param = os.environ.get(env_name)
-        try:
-            return func(env_name, *args, **kwds)
-        finally:
-            # restore the original environ value
-            if environ_api_param:
-                os.environ[env_name] = environ_api_param
-            elif os.environ.get(env_name):
-                del os.environ[env_name]
-
-    return wrapper
-
-
 class TestInitialization(DatadogAPINoInitialization):
 
     def test_no_initialization_fails(self):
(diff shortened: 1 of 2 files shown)


tests/unit/api/test_api.py
```diff
@@ -147,57 +147,73 @@ def test_initialize_options(self):
...
+        self.assertEqual(api._api_key, "API_KEY_ENV")
+        self.assertEqual(api._application_key, "APP_KEY_ENV")
+        self.assertEqual(api._api_host, "HOST_ENV")
+        self.assertEqual(api._host_name, util.hostname.get_hostname())
+
+        del os.environ["DATADOG_API_KEY"]
+        del os.environ["DATADOG_APP_KEY"]
+        del os.environ["DATADOG_HOST"]
+
+    def test_function_param_value(self):
```
nmuesch: Should we add a quick test that tests the order these are used? Like provided attribute >> existing attribute >> environment variable?

PR #395, opened 2019-06-07
[dogshell] Fix dogwrap unicode option parsing on python 3

### LINKED ISSUE #394: Dogwrap command line parsing broken in Python 3 (version 0.29.0)
Version 0.29.0 introduced a bug in the dogwrap script that prevents it from parsing command line arguments on Python 3.6.5. Optparse is returning a list of strings, but the code expects bytes.

```
$ pip install dogwrap==0.29.0
$ dogwrap echo foo
  File "~/.pyenv/versions/3.6.5/bin/dogwrap", line 11, in <module>
    sys.exit(main())
  File "~/.pyenv/versions/3.6.5/lib/python3.6/site-packages/datadog/dogshell/wrap.py", line 264, in main
    cmd = b" ".join(args).decode("utf-8")
TypeError: sequence item 0: expected a bytes-like object, str found
```

Matt343:
The options parser returns a list of `str`, which was getting joined with a `bytes` string. There were also two places where `bytes` were being written to stdout and stderr.

Fixes #394

### DIFF
--- a/datadog/dogshell/wrap.py
+++ b/datadog/dogshell/wrap.py
@@ -35,6 +35,8 @@ ERROR = 'error'
 
 MAX_EVENT_BODY_LENGTH = 3000
 
+PYTHON2 = sys.version_info < (3,)
+
 
 class Timeout(Exception):
     pass
@@ -110,8 +112,10 @@ def execute(cmd, cmd_timeout, sigterm_timeout, sigkill_timeout,
     try:
         # Let's that the threads collecting the output from the command in the
         # background
-        out_reader = OutputReader(proc.stdout, sys.stdout if not buffer_outs else None)
-        err_reader = OutputReader(proc.stderr, sys.stderr if not buffer_outs else None)
+        stdout = sys.stdout if PYTHON2 else sys.stdout.buffer
+        stderr = sys.stderr if PYTHON2 else sys.stderr.buffer
+        out_reader = OutputReader(proc.stdout, stdout if not buffer_outs else None)
+        err_reader = OutputReader(proc.stderr, stderr if not buffer_outs else None)
         out_reader.start()
         err_reader.start()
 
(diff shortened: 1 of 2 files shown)


datadog/dogshell/wrap.py
```diff
@@ -110,8 +112,10 @@ def execute(cmd, cmd_timeout, sigterm_timeout, sigkill_timeout,
     try:
         # Let's that the threads collecting the output from the command in the
         # background
-        out_reader = OutputReader(proc.stdout, sys.stdout if not buffer_outs else None)
-        err_reader = OutputReader(proc.stderr, sys.stderr if not buffer_outs else None)
+        stdout = sys.stdout if PYTHON2 else sys.stdout.buffer
+        stderr = sys.stderr if PYTHON2 else sys.stderr.buffer
```
zippolyte: Let's use the existing `is_py3k()` method from `util.compat`

datadog/dogshell/wrap.py
```diff
@@ -259,9 +266,19 @@ def main():
     parser.add_option('--tags', action='store', type='string', dest='tags', default='',
                       help="comma separated list of tags")
 
-    options, args = parser.parse_args()
+    options, args = parser.parse_args(args=raw_args)
+
+    if PYTHON2:
```
zippolyte: same here

datadog/dogshell/wrap.py
```diff
@@ -310,6 +327,10 @@ def main():
     }
 
     if options.buffer_outs:
+        if not PYTHON2:
```
zippolyte: and here

tests/unit/dogwrap/test_dogwrap.py
```diff
@@ -40,3 +41,50 @@ def test_build_event_body(self):
 
         event_body = build_event_body(cmd, returncode, stdout, stderr, notifications)
         self.assertEqual(expected_body, event_body)
+
+    def test_parse_options(self):
+        options, cmd = parse_options([])
+        self.assertEqual(cmd, '')
+
+        # The output of parse_args is already unicode in python 3, so don't encode the input
+        if sys.version_info < (3,):
```
zippolyte: here as well

datadog/dogshell/wrap.py
```diff
@@ -35,6 +35,8 @@
 
 MAX_EVENT_BODY_LENGTH = 3000
 
+PYTHON2 = sys.version_info < (3,)
```
zippolyte: let's remove this

datadog/dogshell/wrap.py
Matt343: Ah I didn't see this, fixed

PR #396, opened 2019-06-10
Convert CaseInsensitiveDict response headers to dict

Tenzer:
Requests by default makes the response headers into it's own custom type,
`CaseInsensitiveDict`, which can't be JSON serialized without extra work.
This affects the `dog timeboard show ...` and `dog screenboard show ...`
commands, as they JSON encode the Python data structures and they will give
the following error message without this change:

    TypeError: Object of type CaseInsensitiveDict is not JSON serializable

This fixes it by explictly converting the response headers object to a dict
before being added to the response object.

### DIFF
--- a/datadog/api/api_client.py
+++ b/datadog/api/api_client.py
@@ -159,7 +159,8 @@ class APIClient(object):
 
             if response_obj:
                 for response in response_obj:
-                    response['response_headers'] = response_headers
+                    # Convert response headers from CaseInsensitiveDict to dict so it can be JSON serialized
+                    response['response_headers'] = dict(response_headers)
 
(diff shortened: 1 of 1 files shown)


Tenzer:
Rebased on the latest master branch.

agottlie:
Hey @Tenzer looks like this PR that was just merged should resolve this issue: https://github.com/DataDog/datadogpy/pull/397 - can you take a look and see if that does the trick?

Tenzer:
I can see this PR is no longer necessary as the addition of the response headers to the response object was reverted again in dac5c409c4fe14a2c341b47e595926a050f6cf1f.

PR #407, opened 2019-06-12
Fixes Integration Tests and fix dogshell screenboard creation

nmuesch:
This PR does the following:

* Adds __init__ files to the integration test folders
**The integrations tests weren't being run previously due to there not being any __init__ files in the integration/performance/dogshell folders so the test framework couldn't load the tests - https://docs.python.org/3/library/unittest.html#unittest-test-discovery

* Fixes the loading of environment variables and overall setUp/tearDown logic of these tests
* Copies the EnvVars class from datadog_checks_dev so we have a context manager for safely manipulating environment variables. 
* Fixes the integration tests that were broken, various fixes based on what the API returns/and logic issues within the tests
* Dogshell tests are also fixed up and should pass

Note: We aren't actually running these integration tests in Travis quite yet, but the integration tests should now pass if run locally. 

Pillow support - https://pillow.readthedocs.io/en/latest/installation.html (dropped for Python 3.4)

### DIFF
--- a/datadog/dogshell/screenboard.py
+++ b/datadog/dogshell/screenboard.py
@@ -199,8 +199,8 @@ class ScreenboardClient(object):
             raise Exception('bad json parameter')
 
         res = api.Screenboard.update(
-            args.screenboard_id, title=args.title, description=args.description,
-            graphs=graphs, template_variables=args.template_variables,
+            args.screenboard_id, board_title=args.title, description=args.description,
+            widgets=graphs, template_variables=args.template_variables,
             width=args.width, height=args.height)
         report_warnings(res)
         report_errors(res)
@@ -273,9 +273,9 @@ class ScreenboardClient(object):
             graphs = json.loads(graphs)
         except:
             raise Exception('bad json parameter')
-        res = api.Screenboard.create(title=args.filename,
+        res = api.Screenboard.create(board_title=args.filename,
                                      description="Description for {0}".format(args.filename),
-                                     graphs=[graphs])
+                                     widgets=[graphs])
         report_warnings(res)
         report_errors(res)
 
--- a/datadog/dogshell/timeboard.py
+++ b/datadog/dogshell/timeboard.py
@@ -147,6 +147,7 @@ class TimeboardClient(object):
             title=args.filename,
             description="Description for {0}".format(args.filename),
             graphs=[graphs])
+
         report_warnings(res)
         report_errors(res)
 
@@ -173,6 +174,9 @@ class TimeboardClient(object):
 
             if string_ids:
                 dash_obj["id"] = str(dash_obj["id"])
+
+            if not dash_obj.get('template_variables'):
+                dash_obj.pop('template_variables', None)
 
             json.dump(dash_obj, f, indent=2)
 
--- a/tests/integration/api/test_api.py
+++ b/tests/integration/api/test_api.py
@@ -86,8 +86,8 @@ class TestDatadog(unittest.TestCase):
                                            date_happened=before_ts)['event']['id']
         time.sleep(self.wait_time)
 
-        now_event = dog.Event.get(now_event_id)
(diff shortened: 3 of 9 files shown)


datadog/dogshell/timeboard.py
```diff
@@ -174,6 +175,9 @@ def _write_dash_to_file(cls, dash_id, filename, timeout, format='raw', string_id
             if string_ids:
                 dash_obj["id"] = str(dash_obj["id"])
 
+            if not dash_obj.get('template_variables'):
+                dash_obj.pop('template_variables', None)
```
zippolyte: What's this for ?

datadog/dogshell/screenboard.py
```diff
@@ -199,8 +199,8 @@ def _update(cls, args):
             raise Exception('bad json parameter')
 
         res = api.Screenboard.update(
-            args.screenboard_id, title=args.title, description=args.description,
-            graphs=graphs, template_variables=args.template_variables,
+            args.screenboard_id, board_title=args.title, description=args.description,
+            widgets=graphs, template_variables=args.template_variables,
```
zippolyte: Was it not working before ?

tests/integration/api/test_api.py
```diff
@@ -203,62 +193,36 @@ def test_timeboard_validation(self):
...
 
         # No title
-        try:
-            dog.Timeboard.create(title=None, description='my api timeboard', graphs=[graph])
-            assert False, "Should report an api error"
-        except ApiError as e:
-            exception_msg = e.args[0]['errors'][0]
-            eq(exception_msg, "The parameter 'title' is required")
-
-        # No description
```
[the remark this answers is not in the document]
nmuesch: Description is optional - https://docs.datadoghq.com/api/?lang=python#create-a-dashboard

tests/integration/api/test_api.py
```diff
@@ -333,7 +297,14 @@ def test_metrics(self):
         time.sleep(self.wait_time)
 
         metric_query = dog.Metric.query(start=now_ts - 3600, end=now_ts + 3600,
-                                        query="avg:%s{host:%s}" % (metric_name, host_name))
+                                        query="%s{host:%s}" % (metric_name, host_name))
```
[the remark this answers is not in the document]
nmuesch: Based on our docs, it looks like the aggregator (avg) shouldn't be in the query - https://docs.datadoghq.com/api/?lang=python#query-timeseries-points

tests/integration/dogshell/test_dogshell.py
```diff
@@ -98,8 +92,6 @@ def test_comment(self):
         time.sleep(self.wait_time)
         cmd = ["comment", "show", post_data['id']]
         out, err, return_code = self.dogshell(cmd, check_return_code=False)
-        self.assertEquals(out, '')
-        self.assertEquals(return_code, 1)
```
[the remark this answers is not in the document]
nmuesch: There are some issues with the Comment API in terms of deletion.

datadog/dogshell/timeboard.py
nmuesch: So, the way dogshell works is that you tell it to create a new file. Dogshell then calls the API and places the response inside the file on disk. 

You make whatever changes to the file you want, then tell dogshell to update the Datadog timeboard based on the contents of that file. The API read seems to return` template_variables: null` if none are defined. But when you make the update request, its not allowed to be None, the API will return an error. So the easiest solution, IMO is just not to save it to the file if its empty.

datadog/dogshell/screenboard.py
nmuesch: Nope, not from what I can tell. `Graphs` are a timeboard thing. `Widgets` are for screenboards (same for title vs board_title) - https://docs.datadoghq.com/graphing/guide/screenboard-api-doc/?tab=python#arguments

PR #413, opened 2019-06-26
Add project URLs to package metadata

Tenzer:
This will add more links in the "Project links" section in the left hand
sidebar on the PyPI page, liking to the bug tracker, documentation and source
code. It will make it easier for users of the library and automated package
updating tools to figure out what's changed, where to report issues and where
to find the documentation for the package.


Tenzer:
An example of how this would look can for instance be found on Stripe's library: https://pypi.org/project/stripe/.

Issue #408, opened 2019-06-16
http status codes not passed back to caller


mzizzi:
In some cases it's useful for the caller to know the underlying http status code returned by requests. As of now, status codes never make their way back to the caller. This makes it difficult to decide how to proceed after a failed request. Just about every error that the API could throw is wrapped in something like `{'errors': [...]}`. This is fine for some use cases but unless we know every possible message that might end up in that datastruct we'll have a hard time knowing _exactly_ how to proceed.

There is the option to set `datadog.api._mute = False` which will cause exceptions to bubble up to the caller. But in this case, things like [ApiError](https://github.com/DataDog/datadogpy/blob/d4ce8ee3dd551db1a5d0b04b2cfacefc8f2c5077/datadog/api/exceptions.py#L58) are too broad to take make decisions on. e.g. I may wish to never  retry a failed request due to a 403 or 404 but I certainly want to retry a request with that has failed with a 429.

https://github.com/DataDog/datadogpy/pull/378 is related in that it's asking for a little more detail on failed requests. Maybe we pass both status codes and relevant headers back to the caller?

nmuesch:
Hey, we just merged in https://github.com/DataDog/datadogpy/pull/412 which allows you to retrieve the full response object from the previous request. This way you can access the status_code as well as the response headers in whichever cases make sense. Going to close this issue as this will be out in the next release. Thanks for the report!

mzizzi:
I'd like to re-open this. The method of fetching the entire http response in #412 isn't sufficient in multi-threaded programs.

nmuesch:
Posting an update here for completeness: we continued this discussion on the original PR. This was reimplemented in a more threadsafe way https://github.com/DataDog/datadogpy/pull/414 that allows you to retrieve the full raw response object.

nmuesch:
Reopening based on the note from #414

zippolyte:
As discussed in #414, this is likely not something that we'll be able to do without and significant rework of the library while keeping backward compatibility.
However we are currently working on a new and fresh python library for interacting with Datadog's API that wouldn't have that issue. I can't share an exact ETA at the moment.

PR #414, opened 2019-06-26
Add init option to add raw response to return values

nmuesch:
Reverts https://github.com/DataDog/datadogpy/pull/412 and re implements in a different way. 

This PR makes the approach threadsafe while keeping backwards compatibility by using a configuration option passed through the global init. `return_raw_response` is set to false by default, but when its explicitly enabled, we return the decoded response as well as the raw response object from the `submit` method.

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -31,7 +31,7 @@ logging.getLogger('datadog.threadstats').addHandler(NullHandler())
 
 def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
                statsd_host=None, statsd_port=None, statsd_use_default_route=False,
-               statsd_socket_path=None, statsd_namespace=None, **kwargs):
+               statsd_socket_path=None, statsd_namespace=None, return_raw_response=False, **kwargs):
     """
     Initialize and configure Datadog.api and Datadog.statsd modules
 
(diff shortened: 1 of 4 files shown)


mzizzi:
This is an improvement but we should it should be noted that the underlying response isn't always returned. The SDK's underlying `HTTPClient` could still cause exceptions to be thrown in some cases without returning the underlying response. e.g. 401 responses.

mzizzi:
Explicitly notifying @nmuesch again.

nmuesch:
Hey @mzizzi Thats a valid point! I was taking a look but I don't see a way to keep this backwards compatible, thread safe, and return the raw response always. The HTTP Request wrapper, as it currently stands, wouldn't be able to return the response object and maintain the error codes its returning.

I'm going to reopen https://github.com/DataDog/datadogpy/issues/408 but I don't think this is something we'll be able to tackle in the short term.

VidhushiniSrinivasan16:
currently I get this issue :
ImportError: cannot import name '_return_raw_response' from 'datadog.api' (/var/task/datadog/api/__init__.py)
How do we enable return_raw_response?

zippolyte:
Hey @VidhushiniSrinivasan16 !
You need to call `initialize` with the `return_raw_response` argument to enable it.

I recently merged a PR (#461) that would add its default value to the module in order to solve that import error, but it is not yet released.

Hope this helps !

VidhushiniSrinivasan16:
@zippolyte Thanks Henry. That worked for me!

PR #415, opened 2019-06-27
[dogshell] enable proxy

ian28223:
### What does this PR do

Allows for configuring a proxy for Dogshell.

sample ~/.dogrc that utilizes a proxy

```
[Connection]
apikey = xxx
appkey = yyy
api_host = https://api.datadoghq.com

[Proxy]
http = http://localhost:3128/
https = http://localhost:3128/
```

### Motivation

Someone asked in public slack how to use dogshell with a proxy.
Told him he could set `api_host` to a reverse proxy. Then, I looked at the code and found that dogshell uses `datadog.api.initialize` [here](https://github.com/DataDog/datadogpy/blob/master/datadog/dogshell/__init__.py#L72-L73)

...and that `initialize` already takes in proxies as a parameter [here](https://github.com/DataDog/datadogpy/blob/master/datadog/__init__.py#L44-L46)

... so I figured I could just find a way pass the information from dogrc

### DIFF
--- a/datadog/dogshell/common.py
+++ b/datadog/dogshell/common.py
@@ -78,7 +78,7 @@ class DogshellConfig(IterableUserDict):
             self['api_key'] = config.get('Connection', 'apikey')
             self['app_key'] = config.get('Connection', 'appkey')
             if config.has_section('Proxy'):
-                self['proxies'] = config._sections['Proxy']
+                self['proxies'] = dict(config.items('Proxy'))
             if config.has_option('Connection', 'host_name'):
                 self['host_name'] = config.get('Connection', 'host_name')
             if config.has_option('Connection', 'api_host'):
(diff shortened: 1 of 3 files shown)


ian28223:
> Seems legit to me. However could you add the motivation for this in the PR description

Done! And thanks for the approval! 🙇

datadog/dogshell/common.py
```diff
@@ -77,6 +77,8 @@ def load(self, config_file, api_key, app_key):
 
             self['api_key'] = config.get('Connection', 'apikey')
             self['app_key'] = config.get('Connection', 'appkey')
+            if config.has_section('Proxy'):
+                self['proxies'] = config._sections['Proxy']
```
zippolyte:
```suggestion
                self['proxies'] = dict(config.items('Proxy'))
```
Let's do like this to not use the private attribute directly

PR #417, opened 2019-07-26
Add deprecation warning to old aws lambda threadstats integration

DarcyRaynerDD:
The lambda wrapper in the thread stats library is now deprecated for the new [datadog lambda layer](https://github.com/DataDog/datadog-lambda-layer-python), which has the same functionality. We want to introduce a warning so people will know to upgrade.

### DIFF
--- a/datadog/threadstats/aws_lambda.py
+++ b/datadog/threadstats/aws_lambda.py
@@ -5,7 +5,7 @@ import os
 import warnings
 
 """
-DEPRECATED use datadog-lambda package instead https://github.com/DataDog/datadog-lambda-layer-python
+DEPRECATED use datadog-lambda package instead https://git.io/fjy8o
 Usage:
 
 from datadog import datadog_lambda_wrapper, lambda_metric
--- a/datadog/threadstats/aws_lambda.py
+++ b/datadog/threadstats/aws_lambda.py
@@ -5,7 +5,7 @@ import os
 import warnings
 
 """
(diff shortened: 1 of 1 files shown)


datadog/threadstats/aws_lambda.py
```diff
@@ -61,6 +63,8 @@ def _close(cls):
                     _lambda_stats.flush(float("inf"))
 
     def __call__(self, *args, **kw):
+        warnings.warn(
+            "datadog_lambda_wrapper() is deprecated; use datadog-lambda package instead https://github.com/DataDog/datadog-lambda-layer-python", DeprecationWarning)
```
tianchu: "datadog_lambda_wrapper() is deprecated" can potentially cause confusion. For users, it just has been relocated, e.g., user still need to import `datadog_lambda_wrapper`, but from a different location. Perhaps `datadog_lambda_wrapper() is relocated to https://github.com/DataDog/datadog-lambda-layer-python`?

tianchu: BTW, is it better to add this warning in the `__init__` method instead?

DarcyRaynerDD: I think if it's in init it won't get caught in cloud watch logs, because it's outside the invocation.

DarcyRaynerDD: I updated the url to use githubs link shortener, otherwise the line trips up the linter for being too long.

PR #419, opened 2019-08-01
Treat API_HOST as URL

### LINKED ISSUE #411: API_HOST should be manipulated as an URL and not as a string
This creates errors when changing the host to something with a trailing `/`

https://github.com/DataDog/datadogpy/blob/7fb486f53e6992e088d27cc5d644428681affdd4/datadog/api/api_client.py#L121

dabcoder:
Attempts to fix #411.

### DIFF
--- a/datadog/api/api_client.py
+++ b/datadog/api/api_client.py
@@ -14,6 +14,11 @@ from datadog.api.exceptions import (
 )
 from datadog.api.http_client import resolve_http_client
 from datadog.util.compat import is_p3k
+try:
+    from urllib.parse import urljoin
+except ImportError:
+    from urlparse import urljoin
+
 
 log = logging.getLogger('datadog.api')
 
@@ -92,10 +97,6 @@ class APIClient(object):
             if _application_key:
                 params['application_key'] = _application_key
 
-            # Remove trailing backslashes from the _api_host
-            if _api_host[-1] == "/":
-                _api_host = _api_host[:-1]
-
             # Attach host name to body
             if attach_host_name and body:
                 # Is it a 'series' list of objects ?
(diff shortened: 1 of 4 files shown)


datadog/api/api_client.py
```diff
@@ -14,6 +14,11 @@
 )
 from datadog.api.http_client import resolve_http_client
 from datadog.util.compat import is_p3k
+try:
+    from urllib.parse import urljoin
+except ImportError:
+    from urlparse import urljoin
```
zippolyte: Let's move this bit into `util/compat.py`

datadog/api/api_client.py
```diff
@@ -117,12 +122,10 @@ def submit(cls, method, path, api_version=None, body=None, attach_host_name=Fals
...
                 headers['Content-Type'] = 'application/json'
+
             # Construct the URL
-            url = "{api_host}/api/{api_version}/{path}".format(
-                  api_host=_api_host,
-                  api_version=api_version,
-                  path=path.lstrip("/"),
-            )
+            start_url = urljoin(_api_host, 'api/{}'.format(api_version))
+            url = urljoin(start_url, path.lstrip("/"))
```
zippolyte: No need to `lstrip`, `urljoin` does it for you

tests/unit/api/test_api.py
```diff
@@ -240,10 +240,10 @@ def test_creatable(self):
         Creatable resource logic.
         """
         MyCreatable.create(mydata="val")
-        self.request_called_with('POST', "host/api/v1/creatables", data={'mydata': "val"})
+        self.request_called_with('POST', API_HOST + "/api/v1/creatables", data={'mydata': "val"})
 
         MyCreatable.create(mydata="val", attach_host_name=True)
-        self.request_called_with('POST', "host/api/v1/creatables",
+        self.request_called_with('POST', API_HOST + "api/v1/creatables",
```
zippolyte: You're missing a forward slash in a bunch of those tests

tests/unit/api/helper.py
```diff
@@ -28,7 +28,7 @@
 
 API_KEY = "apikey"
 APP_KEY = "applicationkey"
-API_HOST = "host"
+API_HOST = "https://api.datadoghq.com"
```
nmuesch: Nit: Since these are all unit tests and shouldn't actually interact with the API, can we just make this `"https://example.com"` or something similarly obvious that it isn't a real integration test?

dabcoder: Updated

PR #425, opened 2019-08-13
config: do not leak file descriptor when reading config

jd:
Makes sure that the config file descriptor is closed once read.


jd:
Any reason for this not to be merged already?

PR #427, opened 2019-08-13
Replace nose by pytest

jd:
Nose is not maintained for a few years now; it's not a really good idea to
stick to it at this point.

Uses pytest instead to run tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,6 @@ pip-log.txt
 # Unit test / coverage reports
 .coverage
 .tox
-nosetests.xml
 
 #Doc
 ###doc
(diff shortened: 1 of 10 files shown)


jd:
I did not find how to run integrations test. Any hint?

dabcoder:
Sorry for the delay, thanks for working on this.

> I did not find how to run integrations test. Any hint?

Should be with `python setup.py test` I think. Additional context: https://github.com/DataDog/datadogpy/pull/407.

jd:
> Should be with python setup.py test I think.

Alright, I thought they should run with Travis but I see in #407 it's not the case (yet).

I've updated the PR to fix a `mark` issue and a few other things. It should work now. Anyone that knows how to run this  with an API key can double check :)

jd:
Rebased.

zippolyte:
Thanks, merging !

Issue #416, opened 2019-07-03
Constant error: "datadog.api: No agent or invalid configuration file found"


apenney:
I'm getting a lot of errors like:

```
2019-07-03 16:20:12,233 [INFO:140620589405376:26627] datadog.api: No agent or invalid configuration file found
```

After digging in a bit I realized in `datadog/util/config.py` we have:

```def _unix_config_path():
    path = os.path.join('/etc/dd-agent', DATADOG_CONF)
    if os.path.exists(path):
        return path
    raise PathNotFound(path)
```

However, on my machines (AWS Linux 2):

```
# ls -lad /etc/datadog-agent/
drwxr-xr-x 4 dd-agent dd-agent 187 Jun 10 19:04 /etc/datadog-agent/
```

Does it make sense to check for both paths, change this, etc? Something definitely feels wrong here.

dabcoder:
Thanks for reporting this @apenney, we'll be looking into this and get back to you, thanks for the details.

dabcoder:
We've taken care of this in #428 (downgrade to warning and add an option hostname_from_config=False) to turn this off completely - it will be available in the next release, and we aim to deprecate this feature in the future.

bkabrda:
Closing this issue now, see @dabcoder's comment above on how to get rid of these errors.

PR #428, opened 2019-08-16
Hostname lookup in debug and on/off setup

dabcoder:
Attempts to address #416.

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -47,9 +47,6 @@ def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
     :param api_host: Datadog API endpoint
     :type api_host: url
 
-    :param config_lookup: Hostname resolution using the Datadog agent configuration file
-    :type config_lookup: boolean
-
     :param statsd_host: Host of DogStatsd server or statsd daemon
     :type statsd_host: address
 
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -29,9 +29,10 @@ logging.getLogger('datadog.dogstatsd').addHandler(NullHandler())
 logging.getLogger('datadog.threadstats').addHandler(NullHandler())
 
 
-def initialize(api_key=None, app_key=None, host_name=None, hostname_from_config=True, api_host=None,
+def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
                statsd_host=None, statsd_port=None, statsd_use_default_route=False,
-               statsd_socket_path=None, statsd_namespace=None, return_raw_response=False, **kwargs):
+               statsd_socket_path=None, statsd_namespace=None, return_raw_response=False,
+               hostname_from_config=True, **kwargs):
     """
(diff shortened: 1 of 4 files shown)


bkabrda:
I'm wondering whether it might be better to add an argument to `get_hostname`, e.g. `def get_hostname(from_config=True)`, so that people using this from outside of the library could use it better (while preserving current behavior). Additionally, I see you added the argument to the docstring of `initialize`, but I don't actually see the argument there - perhaps you forgot about it?

dabcoder:
> I'm wondering whether it might be better to add an argument to get_hostname, e.g. def get_hostname(from_config=True), so that people using this from outside of the library could use it better (while preserving current behavior).

That is a possibility, good point. Will look into that.

> Additionally, I see you added the argument to the docstring of initialize, but I don't actually see the argument there - perhaps you forgot about it?

No I was looking at other optional arguments such as `cacert`: https://github.com/DataDog/datadogpy/blob/42b578c9543d99861b225ee1965f69d65947058a/datadog/__init__.py#L63-L66 but it's not used anywhere in the `initialize` method either, so was not sure how to reference it.

bkabrda:
I'm not sure about the cacert specifically, but we should either:

* Make this a global configuration variable and then not pass it to `initialize`.
* Make this an argument to `initialize`, thus passing it to `get_hostname`, which is called inside `initialize`.

apenney:
Just a bump to ask how this one is going?

I was reminded this morning when adding a new log source and realizing I see over 250,000 of these errors a day in our logs.

apenney:
You're the best!

datadog/__init__.py
```diff
@@ -29,7 +29,7 @@
 logging.getLogger('datadog.threadstats').addHandler(NullHandler())
 
 
-def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
+def initialize(api_key=None, app_key=None, host_name=None, hostname_from_config=True, api_host=None,
```
bkabrda: I'd honestly put the new argument at the very end of the argument list, because it someone is doing right now `initialize("xxx", "yyy", None, "some_api_host")`, this would break their code. Optional arguments should always be added to the end.

datadog/__init__.py
```diff
@@ -80,7 +87,11 @@ def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
         api._application_key or
         os.environ.get('DATADOG_APP_KEY', os.environ.get('DD_APP_KEY'))
     )
-    api._host_name = host_name or api._host_name or get_hostname()
```
bkabrda: I actually wanted to keep this line that you removed, but it would call `get_hostname(hostname_from_config)` instead of just `get_hostname()`. We can't just ignore the `get_hostname` invocation completely, as it still does different things when the config is not found.

datadog/__init__.py
dabcoder: 👍 Done in the last commit

PR #429, opened 2019-08-23
Add support for Cloud integration endpoints

Ricky-Thomas:
This PR adds the existing Cloud integration endpoints to the python client.


dabcoder:
@Ricky-Thomas thanks for working on this! Looks like flake8 tests are failing, can you have a look at that then?

Ricky-Thomas:
Still a WIP. Working to make most recent failing tests pass.

tests/integration/api/test_aws_integration.py
```diff
@@ -0,0 +1,92 @@
...
+        ]
+        assert all(k in output['accounts'][0].keys() for k in expected_fields)
+
+    def test_delete(self):
+        output = dog.Aws.delete(account_id="123456789101", role_name="DatadogApiTestRole")
+        assert output == {}
+
+    def test_generate_new_external_id(self):
+        output = dog.Aws.generate_new_external_id(
+            account_id="123456789102",
```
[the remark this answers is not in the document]
nmuesch: Same for some of the other fields.

datadog/api/aws_log_integration.py
```diff
@@ -0,0 +1,111 @@
+from datadog.api.resources import GetableAPIResource, CreateableAPIResource, \
+    SearchableAPIResource, DeletableAPISubResource, \
+    UpdatableAPISubResource, ListableAPISubResource, AddableAPISubResource
+
+
+class AwsLogs(GetableAPIResource, CreateableAPIResource, SearchableAPIResource,
+              DeletableAPISubResource, ListableAPISubResource, UpdatableAPISubResource,
+              AddableAPISubResource):
+    """
+    A wrapper around Event HTTP API.
```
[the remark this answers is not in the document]
Ricky-Thomas: 🙇

datadog/api/azure_integration.py
```diff
@@ -0,0 +1,85 @@
+from datadog.api.resources import GetableAPIResource, CreateableAPIResource, \
+    SearchableAPIResource, DeletableAPIResource, \
+    UpdatableAPIResource, ListableAPISubResource, AddableAPISubResource
+
+
+class Azure(GetableAPIResource, CreateableAPIResource, SearchableAPIResource,
+            DeletableAPIResource, ListableAPISubResource, UpdatableAPIResource,
+            AddableAPISubResource):
+    """
+    A wrapper around Event HTTP API.
```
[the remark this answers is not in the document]
Ricky-Thomas: 🙇

tests/integration/api/test_aws_integration.py
```diff
@@ -0,0 +1,92 @@
...
+APP_KEY = os.environ.get("DD_TEST_CLIENT_APP_KEY", "a" * 40)
+API_HOST = os.environ.get("DATADOG_HOST")
+FAKE_PROXY = {"https": "http://user:pass@10.10.1.10:3128/"}
+
+
+class TestAwsIntegration:
+
+    @classmethod
+    def setup_class(cls):
+        initialize(api_key=API_KEY, app_key=APP_KEY, api_host=API_HOST)
```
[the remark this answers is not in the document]
Ricky-Thomas: Yeah I can revisit that later on. Pressed for time today. I need to circle back in the near future and add a few thus far undocumented filtering endpoints for AWS that are totally separate from all this. So I can circle back on that then.

tests/integration/api/test_aws_integration.py
```diff
@@ -0,0 +1,92 @@
+import os
+from datadog import api as dog
+from datadog import initialize
+
+TEST_USER = os.environ.get("DD_TEST_CLIENT_USER")
+API_KEY = os.environ.get("DD_TEST_CLIENT_API_KEY", "a" * 32)
+APP_KEY = os.environ.get("DD_TEST_CLIENT_APP_KEY", "a" * 40)
+API_HOST = os.environ.get("DATADOG_HOST")
+FAKE_PROXY = {"https": "http://user:pass@10.10.1.10:3128/"}
```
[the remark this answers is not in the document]
Ricky-Thomas: 🙇

tests/integration/api/test_aws_integration.py
```diff
@@ -0,0 +1,92 @@
...
+        call.
+        """
+        dog.Aws.delete(account_id="123456789101", role_name="DatadogApiTestRole")
+        dog.Aws.delete(account_id="123456789106", role_name="DatadogApiTestRolo")
+
+    def test_create(self):
+        output = dog.Aws.create(
+            account_id="123456789103",
+            role_name="DatadogApiTestRole",
+            host_tags=["api:test"]
```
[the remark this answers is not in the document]
Ricky-Thomas: Adjusted with all the options. Thanks 🙇

tests/integration/api/test_aws_integration.py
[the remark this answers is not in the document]
Ricky-Thomas: Made changes in both this file and AwsLogs to make it a bit cleaner.

tests/integration/api/test_aws_integration.py
```diff
@@ -0,0 +1,99 @@
+import os
+from datadog import api as dog
+from datadog import initialize
+
+API_KEY = os.environ.get("DD_TEST_CLIENT_API_KEY", "a" * 32)
+APP_KEY = os.environ.get("DD_TEST_CLIENT_APP_KEY", "a" * 40)
+API_HOST = os.environ.get("DATADOG_HOST")
```
zippolyte: Let's not duplicate these, and create a `constants.py` from which to import all the constants we need

tests/integration/api/test_aws_integration.py
```diff
@@ -0,0 +1,99 @@
...
+            role_name=TEST_ROLE_NAME
+        )
+        assert "external_id" in output
+
+    def test_list_namespace_rules(self):
+        output = dog.Aws.list_namespace_rules(
+            account_id=TEST_ACCOUNT_ID_2,
+            role_name=TEST_ROLE_NAME
+        )
+        assert len(output) >= 76
```
zippolyte: What is this 76 ?

tests/integration/api/test_aws_integration.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        initialize(api_key=API_KEY, app_key=APP_KEY, api_host=API_HOST)
+        dog.Aws.create(
+            account_id=TEST_ACCOUNT_ID_1,
+            role_name=TEST_ROLE_NAME
+        )
+        dog.Aws.create(
+            account_id=TEST_ACCOUNT_ID_2,
+            role_name=TEST_ROLE_NAME
+        )
```
zippolyte: Why create these as part of a `setup_class` ? we could create it only in the test that requires. 
Right now, you're modifying those as part of the test, so the `teardown_class` doesn't delete the same things that you created, which is counter-intuitive. Plus if one of the tests modifying that resource don't pass, you won't be deleting the correct thing in the teardown.

tests/integration/api/test_aws_logs.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @classmethod
+    def setup_class(cls):
+        """ setup any state specific to the execution of the given class.
+        """
+        initialize(api_key=API_KEY, app_key=APP_KEY, api_host=API_HOST)
+        dog.Aws.create(
+            account_id=TEST_ACCOUNT_ID,
+            role_name=TEST_ROLE_NAME
+        )
```
zippolyte: In this test the setup part is legit 😄

tests/integration/api/test_aws_logs.py
```diff
@@ -0,0 +1,78 @@
...
+    @classmethod
+    def teardown_class(cls):
+        """ teardown any state that was previously setup with a setup_method
+        call.
+        """
+        dog.Aws.delete(account_id=TEST_ACCOUNT_ID, role_name=TEST_ROLE_NAME)
+
+    def test_list_log_services(self):
+        output = dog.AwsLogs.list_log_services()
+        assert len(output) >= 6
```
zippolyte: What is that 6 ?

tests/integration/api/test_aws_logs.py
```diff
@@ -0,0 +1,78 @@
...
+            lambda_arn=TEST_LAMBDA_ARN
+        )
+        assert 'status' in output.keys()
+
+    def test_check_services(self):
+        output = dog.AwsLogs.check_services(
+            account_id=TEST_ACCOUNT_ID,
+            services=["s3", "elb", "elbv2", "cloudfront", "redshift", "lambda"]
+        )
+        assert 'status' in output.keys()
```
zippolyte: Are those test assuming that log configs were added before running ? If so, let's just explicitly add those in the tests themselves. We shouldn't rely on the execution order of the individual tests. It is also fine to combine those separate tests into one, so that we can rely on some order and not have to repeat a bunch of setup (or use a `setup_method` that will be used before every test method)

tests/integration/api/test_azure_integration.py
```diff
@@ -0,0 +1,81 @@
...
+    test_client_secret = "testingx./Sw*g/Y33t..R1cH+hScMDt"
+    test_new_tenant_name = "1234abcd-1234-5678-9101-abcd1234abcd"
+    test_new_client_id = "abcd1234-5678-1234-5678-1234abcd5678"
+    not_yet_installed_error = 'Azure Integration not yet installed.'
+
+    @classmethod
+    def setup_class(cls):
+        initialize(api_key=API_KEY, app_key=APP_KEY, api_host=API_HOST)
+
+    def test_azure_create(self):
```
zippolyte: Same comment here. It's fine to combine all tests under one `test_azure_crud` so that we can rely on the order of execution of operations.

tests/integration/api/test_azure_integration.py
```diff
@@ -0,0 +1,81 @@
...
+            tenant_name=self.test_new_tenant_name,
+            client_id=self.test_new_client_id
+        )
+        tests_pass = True
+        list_output = dog.Azure.list()
+        if type(list_output) == list:
+            for i in dog.Azure.list():
+                if i['tenant_name'] == self.test_new_tenant_name:
+                    tests_pass = False
+        elif self.not_yet_installed_error in list_output['errors'][0]:
```
zippolyte: Why would we get a `not installed error` here ?

tests/integration/api/test_gcp_integration.py
```diff
@@ -0,0 +1,64 @@
...
+class TestGcpIntegration:
+
+    test_project_id = "datadog-apitest"
+    test_client_email = "api-dev@datadog-sandbox.iam.gserviceaccount.com"
+
+    @classmethod
+    def setup_class(cls):
+        initialize(api_key=API_KEY, app_key=APP_KEY, api_host=API_HOST)
+
+    def test_gcp_create(self):
```
zippolyte: Same here regarding assuming order of execution

datadog/api/aws_integration.py
```diff
@@ -0,0 +1,203 @@
...
+        :param role_name: Your Datadog role delegation name. \
+        For more information about you AWS account Role name, \
+        see the Datadog AWS integration configuration info.
+        :type role_name: string
+
+        :param existing_role_name: Your existing Datadog role delegation name. \
+        For more information about you AWS account Role name, \
+        see the Datadog AWS integration configuration info. \
+        This is the role_name to be updated.
+        :type role_name: string
```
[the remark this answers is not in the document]
Ricky-Thomas: This was done because the endpoint in dogweb expects query parameters of account_id, role_name, etc. AND a JSON body containing account_id, role_name, etc. Those fields being sent as query params being the ones representing the existing fields. 

I'm unsure on the best way to manipulate the body object being passed in with the `new` as they need to remain unchanged in the body object being passed in to line up with expected fields. 

i.e. if we just call the existing_account_id, account_id, we still need to get this out of the body object to be passed in as query params, and we would also then need to introduce logic that will manipulate the `new_account_id` field in the body to change into `account_id`. This would also require additional logic. i.e. my early draft of how this would need to look is as follows:

```
       if body.get('account_id') and body.get('role_name'):
            params['account_id'] = body.pop('account_id')
            params['role_name'] = body.pop('role_name')
            if body.get('new_account_id'):
                body['account_id'] = body.pop('new_account_id')
            if body.get('new_role_name'):
                body['role_name'] = body.pop('new_role_name')
        if body.get('access_key_id'):
            params['access_key_id'] = body.pop('access_key_id')
            params['secret_access_key'] = body.pop('secret_access_key')
            if body.get('new_access_key_id'):
                body['access_key_id'] = body.pop('new_access_key_id')
            if body.get('new_secret_access_key'):
                body['secret_access_key'] = body.pop('new_secret_access_key')
```

That said, I'm having some trouble getting this update endpoint working now with the access_key and secret_key setup. I had that working before I had thought, so revisiting this entirely to make sure it works in all cases.

tests/integration/api/utils.py
```diff
@@ -0,0 +1,30 @@
+# datadog
```
gzussa: 👍

tests/integration/api/test_gcp_integration.py
```diff
@@ -0,0 +1,64 @@
...
+        output = dog.Gcp.create(
+            type="service_account",
+            project_id=self.test_project_id,
+            private_key_id="123456789abcdefghi123456789abcdefghijklm",
```
gzussa: is this a sensitive key, If not could we make it more obvious it is not?

tests/integration/api/test_azure_integration.py
```diff
@@ -0,0 +1,81 @@
+import os
+from datadog import api as dog
+from datadog import initialize
+
+API_KEY = os.environ.get("DD_TEST_CLIENT_API_KEY", "a" * 32)
```
gzussa: Since they are in all files, let's move them to a common place.

datadog/api/aws_integration.py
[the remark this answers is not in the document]
zippolyte: Yeah this sound good to me. I don’t mind having a bit more logic if it means the user experience will be better and more consistent. The update endpoint for azure works like that, so better to have things consistent here.

tests/integration/api/test_gcp_integration.py
Ricky-Thomas: nope that's just numbers and letters in order twice. i.e. 123456789 twice, and abcdefghi twice.  I can change it to something more obvious.

datadog/api/azure_integration.py
```diff
@@ -0,0 +1,85 @@
...
+        >>> client_secret = "<AZURE_CLIENT_SECRET>"
+        >>> host_filters = "<KEY>:<VALUE>"
+
+        >>> api.Azure.update(tenant_name=tenant_name, client_id=client_id, \
+        new_tenant_name=new_tenant_name, new_client_id=new_client_id,\
+        client_secret=client_secret, host_filters=host_filters)
+        """
+        params = {}
+        params['new_tenant_name'] = body.get('new_tenant_name')
+        params['new_client_id'] = body.get('new_client_id')
```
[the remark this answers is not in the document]
Ricky-Thomas: ah yep that's my bad. Passing those in here doesn't actually do anything. We can remove them.

tests/integration/api/test_aws_integration.py
Ricky-Thomas: I did this to avoid having to create the same resource over and over in each method that would use it. I see what you mean though. I've added it to every method individually instead.

tests/integration/api/test_aws_integration.py
Ricky-Thomas: This is the number of available namespaces that will be output. It does not change unless we add or remove namespaces on our end. We're very unlikely to reduce this number, so I decided on checking the output against it having 76 or more available namespaces.

tests/integration/api/test_aws_logs.py
Ricky-Thomas: That's the number of available services. This is the same for everyone and always returns this. We may add more in the future, hence the "greater than or equal to", but we're very unlikely to reduce that. I'll make this a variable instead for clarity.

tests/integration/api/test_aws_logs.py
Ricky-Thomas: The final two tests here do not require there to be a Lambda config in place to pass. I'll combine the others into one so as to not rely on execution order of individual tests.

tests/integration/api/test_azure_integration.py
Ricky-Thomas: I'm likely misunderstanding something here. But this test essentially deletes the config we've created above (refactored in my upcoming PR in order to make sure we're not relying on order of execution of individual tests). It's my understanding that these tests will not always be run with our Datadog credentials. If the user who is running these tests does not have Azure integrated, then when we use the List output to verify the delete worked, it will return the 'Azure Integration not yet installed.' error message. In that scenario, that output is expected, and the test should still pass. 

There's probably a better way to handle that, but I'm not positive on what that would be.

datadog/api/aws_integration.py
```diff
@@ -0,0 +1,203 @@
...
+          DeletableAPIResource, ListableAPISubResource, UpdatableAPIResource,
+          UpdatableAPISubResource, AddableAPISubResource):
+    """
+    A wrapper around AWS Integration API.
+    """
+    _resource_name = 'integration'
+    _resource_id = 'aws'
+
+    @classmethod
+    def list(cls, id=_resource_id, **params):
```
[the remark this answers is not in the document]
Ricky-Thomas: I've updated this in the latest commit, thanks!

datadog/api/aws_integration.py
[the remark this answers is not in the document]
Ricky-Thomas: This has been updated now to work the way you'd mentioned. I agree it's easier to use. That said, this also surfaced some potential bugs in the AWS Update endpoint in dogweb. I'm looking into that further tomorrow and will potentially be changing this again.

tests/integration/api/test_gcp_integration.py
Ricky-Thomas: I've changed this to `fake_private_key_id` for clarity.

tests/integration/api/test_azure_integration.py
Ricky-Thomas: Thanks!

Ricky-Thomas:
The most recent refactoring work has surfaced a potential bug in the AWS Update endpoint on our side. I'm digging into this tomorrow and may still change the behavior for that here should I have to modify the endpoint in a way that warrants that. This PR can be considered a WIP still at this time. Thanks!

datadog/api/aws_integration.py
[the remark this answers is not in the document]
Ricky-Thomas: This is now working properly with the updated method you'd suggested. I've fixed the bug that was surfaced by this work and should now be good to go!

Ricky-Thomas:
This PR is now ready for another review. The most recent changes address comments from both @zippolyte and @gzussa. The aforementioned potential bug in the AWS Update endpoint on our side has been fixed.

datadog/api/__init__.py
```diff
@@ -35,3 +35,7 @@
 from datadog.api.service_checks import ServiceCheck
 from datadog.api.tags import Tag
 from datadog.api.users import User
+from datadog.api.aws_integration import Aws
```
gzussa: Shouldn't we call classes `AWSIntegration` and Co instead. This kind of shortcuts can bit us later on.

datadog/api/__init__.py
```diff
@@ -35,3 +35,7 @@
...
+from datadog.api.aws_log_integration import AwsLogs
```
gzussa: Shouldn't this be part of the AWSIntegration?

datadog/api/__init__.py
Ricky-Thomas: Thanks for the review! I'm happy to modify all of these new classes to have `Integration` suffixed on the end. Will update!

datadog/api/__init__.py
Ricky-Thomas: I'd prefer to keep `AwsIntegration` and `AwsLogsIntegration` separate if possible, as these encompass pretty different functionality that is owned by two separate teams.

We've also split these into their own sections in a coming docs update for clarity from a user perspective. 

For me, it also just feels less cluttered having them separate as they each have their own respective GET, PUT, POST, DELETE endpoints for different routes.

datadog/api/aws_integration.py
```diff
@@ -0,0 +1,222 @@
...
+        >>> host_tags = ["<KEY>:<VALUE>"]
+        >>> account_specific_namespace_rules = {"namespace1":true/false, "namespace2":true/false}
+
+        >>> api.AwsIntegration.update(account_id=account_id, role_name=role_name, \
+        new_account_id=new_account_id, new_role_name=new_role_name, \
+        filter_tags=filter_tags,host_tags=host_tags,\
+        account_specific_namespace_rules=account_specific_namespace_rules)
+        """
+        params = {}
+        if body.get('account_id') and body.get('role_name'):
```
jd: If somebody wants to ever pass `None` as a value for e.g `role_name` for whatever reason, that wouldn't work.

As rule of thumb, it's safer to go with:
```python
if 'role_name' in body:
    params['role_name'] = body.pop('role_name')
```

datadog/api/aws_integration.py
```diff
@@ -0,0 +1,222 @@
...
+            params['account_id'] = body.pop('account_id')
+            params['role_name'] = body.pop('role_name')
+            if body.get('new_account_id'):
+                body['account_id'] = body.pop('new_account_id')
+            if body.get('new_role_name'):
+                body['role_name'] = body.pop('new_role_name')
```
jd: If I read this correctly, there a few cases where I'm not sure what will happen. If body is:

```python
body = {"account_id": "foobar", "new_role_name": "foobaz"}
```
Then it'll be pass as is to the API. Is that expected?
(I don't know the API :))

datadog/api/aws_integration.py
Ricky-Thomas: If they they were to try and pass in `role_name` as `None`, they would be met with the following error from the API:

`'errors': [u'AWS Account does not exist']`

That's to be expected. It's not currently possible to have `None` saved as the role_name so this will be the case. 

That said, I had thought that `.get` will return `None` if the `role_name` field isn't present, so wouldn't it essentially be the same as if they didn't submit a `role_name` at all?

datadog/api/aws_integration.py
Ricky-Thomas: In this case: 

```
from datadog import initialize, api

options = {
    'api_key': '<DD_API_KEY>',
    'app_key': '<DD_APP_KEY>'
}

initialize(**options)

print(api.AwsIntegration.update(
    account_id="<Existing_account_id>",
    new_role_name="Testing"
))
```

Would return an error from the API:
`{u'errors': [u"Additional properties are not allowed (u'new_role_name' was unexpected)"]}`

This is expected as well.

jd: Cool, as long as that's what we want, fine with me :)

datadog/api/aws_integration.py
jd: My point is that you either 1. implement a full logic in this method, or you 2. do the bare minimum and pass the data to the API and let it validate.
I think you're trying to do 2., which is I think the best solution in this case, but you're just preventing a (probably useless) case to be possible.

It's really a detail and I don't think it'll be an issue for any user, so feel free to ignore. :)

Ricky-Thomas: Okay cool. Yeah we have fairly extensive validation on our end in the API, so that's kinda what I was going for here. Thanks @jd! I really appreciate the review 🙇!

datadog/api/aws_integration.py
Ricky-Thomas: Thanks @jd !

datadog/api/__init__.py
```diff
@@ -36,5 +36,9 @@
 from datadog.api.service_checks import ServiceCheck
 from datadog.api.tags import Tag
 from datadog.api.users import User
+from datadog.api.aws_integration import AwsIntegration
```
gzussa: :+1:

PR #433, opened 2019-09-11
Add Synthetics support

dabcoder:
Add Synthetics support based on https://docs.datadoghq.com/api/?lang=bash#synthetics.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ pip-log.txt
 .eggs/
 .env/
 .idea/
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -42,4 +42,3 @@ pip-log.txt
 .eggs/
 .env/
 .idea/
-.vscode/
--- a/datadog/api/api_client.py
+++ b/datadog/api/api_client.py
@@ -174,14 +174,13 @@ class APIClient(object):
                 except ValueError:
                     raise ValueError('Invalid JSON response: {0}'.format(content))
 
-                if response_obj and 'errors' in response_obj:
-                    # suppress ApiError when specified and just return the response
-                    if not (suppress_response_errors_on_codes and
-                            result.status_code in suppress_response_errors_on_codes):
                 # response_obj can be a bool and not a dict
                 if not isinstance(response_obj, bool):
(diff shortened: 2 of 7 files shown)


Kerl1310:
Conflicts with master are the only things I can pick on

.gitignore
```diff
@@ -42,3 +42,4 @@ pip-log.txt
 .eggs/
 .env/
 .idea/
+.vscode/
```
jd: That should go into _your_ `~/.gitignore`. There's nothing generating `.vscode` in this project, so it'd be better if we don't start adding everyone's favorite editor in `.gitignore`. :)

https://help.github.com/en/github/using-git/ignoring-files#create-a-global-gitignore

datadog/api/api_client.py
```diff
@@ -174,11 +174,13 @@ def submit(cls, method, path, api_version=None, body=None, attach_host_name=Fals
                 except ValueError:
                     raise ValueError('Invalid JSON response: {0}'.format(content))
 
-                if response_obj and 'errors' in response_obj:
-                    # suppress ApiError when specified and just return the response
-                    if not (suppress_response_errors_on_codes and
-                            result.status_code in suppress_response_errors_on_codes):
-                        raise ApiError(response_obj)
+                # response_obj can be a bool and not a dict
+                if not isinstance(response_obj, bool):
```
jd: just go with `isinstance(response_obj, dict)`?

This is going to crash if  `response_obj` ends up being a e.g. `int`.

datadog/api/resources.py
```diff
@@ -99,7 +99,7 @@ class UpdatableAPIResource(object):
     Updatable API Resource
     """
     @classmethod
-    def update(cls, id, params=None, **body):
+    def update(cls, id, synthetics=False, params=None, **body):
```
jd: Is this the best design or wouldn't it be better to have an `UpdatableAPISyntheticsResource` class that is being used instead?
I'm afraid that special casing here with keyword arguments never ends up with some new API coming up.

dabcoder:
Can't reproduce the failure locally. Re-running just in case to see if we get the same value for the number of tests.

PR #437, opened 2019-09-13
Make sure we distribute wheels in addition to sdists

bkabrda:


bkabrda:
Thanks for the review, merging.

PR #440, opened 2019-09-23
switch to azure pipelines

zippolyte:


.azure-pipelines/all.yml
```diff
@@ -0,0 +1,47 @@
...
+    - master
+
+jobs:
+- job: UnitTests
+  pool:
+    vmImage: "Ubuntu-16.04"
+  strategy:
+    matrix:
+      Py27:
+        PYTHON_VERSION: "2.7"
```
gzussa: Do we need all those versions?

zippolyte: It's what travis tested, figured I'd leave it like that

PR #443, opened 2019-09-24
Add DOGSTATSD_HOST_IP as env var

dabcoder:
`DOGSTATSD_HOST_IP` can also be used.


dabcoder:
Closing. All clients support `DD_AGENT_HOST` and we should keep using this env var.

PR #444, opened 2019-09-24
Rename creds env vars for tests

zippolyte:
Don't want to accidentally use the same credentials you would use normally, since the integration tests can be destructive

### DIFF
--- a/tests/integration/dogshell/test_dogshell.py
+++ b/tests/integration/dogshell/test_dogshell.py
@@ -40,7 +40,7 @@ class TestDogshell(unittest.TestCase):
         config = ConfigParser()
         config.add_section('Connection')
         config.set('Connection', 'apikey', os.environ['DD_TEST_CLIENT_API_KEY'])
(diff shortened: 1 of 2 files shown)


tests/integration/dogshell/test_dogshell.py
```diff
@@ -39,8 +39,8 @@ def setUp(self):
         self.config_fn, self.config_file = get_temp_file()
         config = ConfigParser()
         config.add_section('Connection')
-        config.set('Connection', 'apikey', os.environ['DATADOG_API_KEY'])
-        config.set('Connection', 'appkey', os.environ['DATADOG_APP_KEY'])
+        config.set('Connection', 'apikey', os.environ['DD_TEST_CLIENT_API_KEY'])
+        config.set('Connection', 'appkey', os.environ['DD_TEST_CLIENT_API_KEY'])
```
bkabrda: Typo: this should be APP key (you have API).

PR #426, opened 2019-08-13
Replace deprecated assertEquals by assertEqual

jd:

PR #445, opened 2019-09-24
assertNotEquals is deprecated too

dabcoder:
Similar to #426


dabcoder:
Closing are we're going to do plain `assert`s instead.

PR #448, opened 2019-09-26
Fix graph snapshot status endpoint

zippolyte:
Using `urljoin` led to some unexpected results depending on whether there are leading `/`.
Now uses a well defined and tested utility method, so no surprises


dabcoder:
Thanks for updating this!

PR #449, opened 2019-09-26
Fix error printing to stderr char by char

dabcoder:
The current error handling would result in the errors and warnings from the API JSON response to be printed to `stderr` character by character (when the API would return a string and not a list).

### DIFF
--- a/datadog/dogshell/common.py
+++ b/datadog/dogshell/common.py
@@ -20,9 +20,9 @@ def report_errors(res):
         errors = res['errors']
         if isinstance(errors, list):
             for error in errors:
-                print_err("ERROR: " + error)
+                print_err("ERROR: {}".format(error))
         else:
-            print_err("ERROR: " + errors)
+            print_err("ERROR: {}".format(errors))
         sys.exit(1)
     return False
 
@@ -32,9 +32,9 @@ def report_warnings(res):
         warnings = res['warnings']
         if isinstance(warnings, list):
             for warning in warnings:
-                print_err("WARNING: " + warning)
+                print_err("WARNING: {}".format(warning))
         else:
-            print_err("WARNING: " + warnings)
+            print_err("WARNING: {}".format(warnings))
         return True
     return False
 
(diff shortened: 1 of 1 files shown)


datadog/dogshell/common.py
```diff
@@ -17,16 +17,14 @@ def print_err(msg):
 
 def report_errors(res):
     if 'errors' in res:
-        for e in res['errors']:
-            print_err('ERROR: ' + e)
+        print_err('ERROR: ' + res['errors'])
```
zippolyte: `res["errors"]` can actually be a list: e.g. `{'errors': ['No dashboard matches that dash_id.']}`
let's check for the type first, then iterate or not

datadog/dogshell/common.py
```diff
@@ -10,23 +10,33 @@
...
     if 'errors' in res:
-        for e in res['errors']:
-            print_err('ERROR: ' + e)
+        errors = res['errors']
+        if isinstance(errors, str):
+            print_err('ERROR: ' + errors)
+        elif isinstance(errors, list):
+            print('ERROR:')
+            for error in errors:
+                print_err(error)
```
zippolyte: make it `print_err("ERROR: " + error)` and remove the  print('ERROR:')`. Same for warnings

datadog/dogshell/common.py
```diff
@@ -10,23 +10,33 @@
...
     else:
         sys.stderr.write(msg + '\n')
 
 
 def report_errors(res):
     if 'errors' in res:
-        for e in res['errors']:
-            print_err('ERROR: ' + e)
+        errors = res['errors']
+        if isinstance(errors, str):
```
zippolyte: Let's do the following instead, so that we always print, even if the type is not `list` or `str`
```python
if isinstance(errors, list):
    for ...
else:
    print_err(...)
```

datadog/dogshell/common.py
```diff
@@ -10,23 +10,31 @@
...
 
 
 def report_errors(res):
     if 'errors' in res:
-        for e in res['errors']:
-            print_err('ERROR: ' + e)
+        errors = res['errors']
+        if isinstance(errors, list):
+            for error in errors:
+                print_err("ERROR: " + error)
```
zippolyte: My bad, let's use `format` everywhere as well, so that there are no problems with types when trying to do the `+` operation

PR #451, opened 2019-09-26
Fix integration tests

zippolyte:
we need these


tests/integration/api/test_api.py
```diff
@@ -521,221 +357,156 @@ def _compare_screenboard(apiBoard, expectedBoard):
...
-        eq(mute['action'], "Muted")
-        eq(mute['end'], end2)
+        assert mute["hostname"] == hostname
+        assert mute["action"] == "Muted"
+        assert mute["end"] == end2
 
-        dog.Host.unmute(hostname)
+        unmute = dog.Host.unmute(hostname)
+        assert unmute["hostname"] == hostname
+        assert unmute["action"] == "Unmuted"
```
nmuesch: Should we get with retry and assert the host is muted here, or should this be enough?

tests/integration/api/test_api.py
```diff
@@ -521,221 +357,156 @@ def _compare_screenboard(apiBoard, expectedBoard):
...
+            message="PG is WARNING",
+            tags=["db:prod_data"]
+        ) == {"status": "ok"}
 
-    @pytest.mark.host
     def test_host_muting(self):
-        hostname = "my.test.host"
-        message = "Muting this host for a test."
         end = int(time.time()) + 60 * 60
+        hostname = "my.test.host" + str(end)
```
nmuesch: Similar to whats done for the user, would it make sense to put a random token in here so concurrent test runs don't impact one another?

tests/integration/api/test_api.py
```diff
@@ -521,221 +357,156 @@ def _compare_screenboard(apiBoard, expectedBoard):
...
         # Update downtime
         message = "Doing some testing on staging."
         end = int(time.time()) + 60000
-        dog.Downtime.update(downtime_id, scope='env:staging',
-                            end=end, message=message)
-        dt = dog.Downtime.get(downtime_id)
-        eq(dt['end'], end)
-        eq(dt['message'], message)
-        eq(dt['disabled'], False)
+        downtime = dog.Downtime.update(downtime["id"], scope="env:staging", end=end, message=message)
```
nmuesch: Can we not send the scope since its not updated?

tests/integration/api/test_api.py
```diff
@@ -521,221 +357,156 @@ def _compare_screenboard(apiBoard, expectedBoard):
...
+        # We shouldn"t be able to mute a simple alert on a scope.
+        assert "errors" in dog.Monitor.mute(monitor1["id"], scope="env:staging")
+
+        monitor2 = dog.Monitor.mute(monitor2["id"], scope="host:foo")
+        assert monitor2["options"]["silenced"] == {"host:foo": None}
+
+        monitor2 = dog.Monitor.unmute(monitor2["id"], scope="host:foo")
+        assert monitor2["options"]["silenced"] == {}
+
+        dog.Monitor.delete(monitor2["id"])
```
nmuesch: worth asserting the downtime on this monitor actually is disabled?

tests/integration/api/test_api.py
```diff
@@ -521,221 +357,156 @@ def _compare_screenboard(apiBoard, expectedBoard):
...
-        assert monitor1['query'] == query1, monitor1
-        assert monitor2['query'] == query2, monitor2
+        dt = dog.Monitor.mute_all()
+        assert dt["active"] is True
+        assert dt["scope"] == ["*"]
+
+        assert dog.Monitor.unmute_all() is None  # No response expected
+
+        # We shouldn"t be able to mute a simple alert on a scope.
+        assert "errors" in dog.Monitor.mute(monitor1["id"], scope="env:staging")
```
nmuesch: Can we mute the monitor overall and assert on that?

tests/integration/api/test_api.py
```diff
@@ -521,221 +357,156 @@ def _compare_screenboard(apiBoard, expectedBoard):
...
+        assert monitor["query"] == query2
+        assert monitor["options"]["notify_no_data"] is True
 
-        name = 'test_monitors'
-        monitor_id = dog.Monitor.update(monitor_id, query=query2, name=name,
-                                        options={'notify_no_data': True})['id']
-        monitor = dog.Monitor.get(monitor_id)
-        eq(monitor['name'], name)
-        eq(monitor['options']['notify_no_data'], True)
+        monitors = [m for m in dog.Monitor.get_all() if m["id"] == monitor["id"]]
```
nmuesch: What is this testing? IDs should be unique so would this ever not be `1`? Is this more testing that the `get_all` function actually returns a monitor object?

tests/integration/api/test_api.py
```diff
@@ -407,31 +252,26 @@ def test_graph_snapshot(self):
...
-        while dog.Graph.status(snapshot_url)['status_code'] != 200:
-            time.sleep(self.wait_time)
-        if 'localhost' in dog._api_host:
-            snapshot_url = 'http://%s%s' % (dog.api_host, snapshot_url)
+        assert "event_query" not in snap
+        assert snap["metric_query"] == metric_query
+        snapshot_url = snap["snapshot_url"]
+        retry_count = 0
+        dog.Graph.status(snapshot_url)
+        while dog.Graph.status(snapshot_url)["status_code"] != 200 and retry_count < 10:
```
nmuesch: if we setup `get_with_retry` to take in a failure scenario, we could de duplicate this logic

tests/integration/api/test_api.py
```diff
@@ -258,147 +157,93 @@ def test_timeboard(self):
...
-        dog.Metric.send(metric="test.metric", points=[(time.time() - 3600, 1.0)])
-        dog.Metric.send(metric="test.metric", points=1.0)
-        dog.Metric.send(metric="test.metric", points=(time.time(), 1.0))
-
-    @pytest.mark.monitor
-    def test_monitors(self):
-        query = "avg(last_1h):sum:system.net.bytes_rcvd{host:host0} > 100"
+        assert dog.Metric.send(metric="test.metric", points=[(time.time() - 3600, 1.0)])["status"] == "ok"
+        assert dog.Metric.send(metric="test.metric", points=1.0)["status"] == "ok"
+        assert dog.Metric.send(metric="test.metric", points=(time.time(), 1.0))["status"] == "ok"
```
nmuesch: What are we testing here? Should we try to query as well?

tests/integration/api/test_api.py
```diff
@@ -258,147 +157,93 @@ def test_timeboard(self):
...
+                start=now_ts - 600,
+                end=now_ts + 600,
+                query="{}{{host:{}}}".format(metric_name_single, host_name)
+            )
+            metric_query_multi = dog.Metric.query(
+                start=now_ts - 600,
+                end=now_ts + 600,
+                query="{}{{host:{}}}".format(metric_name_multi, host_name)
+            )
+        print(retry_time)
```
nmuesch: Leftover `print`s 😉

tests/integration/api/test_api.py
```diff
@@ -258,147 +157,93 @@ def test_timeboard(self):
...
-            assert False
+        assert dog.Timeboard.get(timeboard["dash"]["id"])["dash"]["id"] == timeboard["dash"]["id"]
+        dog.Timeboard.delete(timeboard["dash"]["id"])
+        assert "errors" in dog.Timeboard.get(timeboard["dash"]["id"])
 
     def test_search(self):
-        results = dog.Infrastructure.search(q='e')
-        assert len(results['results']['hosts']) > 0
-        assert len(results['results']['metrics']) > 0
+        results = dog.Infrastructure.search(q="e")
```
nmuesch: Is this going to be flaky? Assumes we always have a host up with `e`?

tests/integration/api/test_api.py
```diff
@@ -258,147 +157,93 @@ def test_timeboard(self):
...
+            end=now_ts + 600,
+            query="{}{{host:{}}}".format(metric_name_multi, host_name)
+        )
         # There can be a slight delay before the metric is ready to be queried
         retry_time = 0
-        while not metric_query['series'] and retry_time < 10:
-            metric_query = dog.Metric.query(start=now_ts - 3600, end=now_ts + 3600,
-                query="%s{host:%s}" % (metric_name, host_name))
-            time.sleep(self.wait_time)
+        while (not metric_query_single["series"]) and (not metric_query_multi["series"]) and retry_time < 20:
```
nmuesch: same comment for re using `get_with_retry` here?

tests/integration/api/test_api.py
```diff
@@ -6,156 +6,113 @@
...
+        event_id = dog.Event.create(
+            title="test source",
+            text="test source",
+            source_type_name="vsphere",
+            priority="low",
+        )["event"]["id"]
+        get_with_retry("Event", event_id)
+        events = dog.Event.query(start=now_ts-100, end=now_ts+100, priority="low", sources="vsphere")
+        assert events["events"], "No events found in stream"
+        assert event_id in [event["id"] for event in events["events"]]
```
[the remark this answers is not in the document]
zippolyte: We make sure the query function works as it should by querying the event stream and verifying that the event we expect is here

tests/integration/api/test_api.py
```diff
@@ -6,156 +6,113 @@
...
         event_id = dog.Event.create(
-            title='test no hostname',
-            text='test no hostname',
-            attach_host_name=False
-        )['event']['id']
-        time.sleep(self.wait_time)
-        event = self.get_event_with_retry(event_id)
-        assert not event['event']['host']
-
-    def test_aggregate_events(self):
```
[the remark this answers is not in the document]
zippolyte: Exactly, it wasn't testing much. From what the API returns to us, we have no way of asserting on aggregation, so this test is just useless.

tests/integration/api/test_api.py
zippolyte: This is enough. We make a call and the API returns something to us that we can assert on. trying to get again would be testing that the backend correctly does its job, which is not the goal here

tests/integration/api/test_api.py
zippolyte: The timestamp is the "random" key. Having two tests run at the exact same second would be really unlucky I assume, and shouldn't happen.

tests/integration/api/test_api.py
zippolyte: I don't think so. The monitor CRUD is tested in another test.
I added the deletion of monitor1 though

tests/integration/api/test_api.py
zippolyte: yup, testing the `get_all`

tests/integration/api/test_api.py
zippolyte: done

tests/integration/api/test_api.py
zippolyte: testing the different types you can pass to the `points`. Doing that as part of the test_metrics instead, and removing that one

tests/integration/api/test_api.py
zippolyte: passing empty string instead. Seems to search for everything

tests/integration/api/test_api.py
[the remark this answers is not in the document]
nmuesch: Misread, makes sense 👍

tests/integration/api/test_api.py
nmuesch: Ok thats fair. The `get` itself is tested elsewhere, so that makes sense.

tests/integration/api/test_api.py
nmuesch: missed the `+ str(end)`. Thanks for the confirmation

tests/integration/api/test_api.py
```diff
@@ -1,478 +1,301 @@
...
+        assert dog.Timeboard.get(timeboard["dash"]["id"])["dash"]["id"] == timeboard["dash"]["id"]
+        dog.Timeboard.delete(timeboard["dash"]["id"])
+        assert "errors" in dog.Timeboard.get(timeboard["dash"]["id"])
 
     def test_search(self):
-        results = dog.Infrastructure.search(q='e')
-        assert len(results['results']['hosts']) > 0
-        assert len(results['results']['metrics']) > 0
+        results = dog.Infrastructure.search(q="")
+        print(results)
```
nmuesch: One more :)

tests/integration/api/test_api.py
```diff
@@ -1,478 +1,301 @@
...
+    else:
+        resource = getattr(getattr(dog, resource_type), operation)(resource_id, **kwargs)
+    retry_counter = 0
+    while retry_condition(resource) and retry_counter < 10:
+        if resource_id is None:
+            resource = getattr(getattr(dog, resource_type), operation)(**kwargs)
+        else:
+            resource = getattr(getattr(dog, resource_type), operation)(resource_id, **kwargs)
+        retry_counter += 1
+        time.sleep(WAIT_TIME)
```
nmuesch: should we raise an exception here if our retry condition is still what we got back? Currently it'll fail at the test itself when an assertion happens on this data but seems like we could fail faster here.

tests/integration/api/test_api.py
zippolyte: crap

tests/integration/api/test_api.py
zippolyte: good point

tests/integration/util/snapshot_test_utils.py
```diff
@@ -1,39 +0,0 @@
-import io
-
-from PIL import Image
```
nmuesch: If we're removing this, lets also kill it in setup.py https://github.com/DataDog/datadogpy/blob/master/setup.py#L25

PR #452, opened 2019-09-30
Rework dogshell integration tests

zippolyte:
also fixes one bug with the api tests

### DIFF
--- a/tests/integration/dogshell/test_dogshell.py
+++ b/tests/integration/dogshell/test_dogshell.py
@@ -53,24 +53,22 @@ class TestDogshell:
         cmd = ["comment", "post", TEST_USER]
         comment_msg = "yo dudes"
         post_data = {}
-        out, err, return_code = self.dogshell(cmd, stdin=comment_msg)
+        out, _, _ = self.dogshell(cmd, stdin=comment_msg)
         post_data = self.parse_response(out)
         assert "id" in post_data
         assert "url" in post_data
-        assert "message" in post_data
         assert comment_msg in post_data["message"]
 
         # Read that comment from its id
-        time.sleep(WAIT_TIME)
         cmd = ["comment", "show", post_data["id"]]
-        out, err, return_code = self.dogshell(cmd)
+        out, _, _ = self.dogshell_with_retry(cmd)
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -7,7 +7,7 @@ envlist =
     integration
 
 [testenv]
-passenv = *
+passenv = DD_TEST_CLIENT*
```
nmuesch: To avoid accidentally using things like API/APP key env vars?

zippolyte: Yes. Because we invoke `dog` as a subprocess directly, and it uses the regular env variables by default

tests/integration/dogshell/test_dogshell.py
```diff
@@ -109,20 +99,13 @@ def match_permalink(out):
         out, err, return_code = self.dogshell(cmd, stdin=body)
 
         event_id = match_permalink(out)
-        assert event_id, out
-
-        # Add a bit of latency for the event to appear
-        time.sleep(self.wait_time)
```
nmuesch: no need for the latency anymore?

tests/integration/dogshell/test_dogshell.py
```diff
@@ -134,115 +113,83 @@ def match_permalink(out):
...
-                    }]
+        new_dash = [
+            {
+                "title": "blerg",
+                "definition": {"viz": "timeseries", "requests": [{"q": "avg:system.load.15{web,env:prod}"}]},
+            }
+        ]
 
         # Update a dash directly on the server
-        self.dogshell(["timeboard", "update", str(dash["id"]), new_title, new_desc],
```
nmuesch: not updating anymore here?

tests/integration/dogshell/test_dogshell.py
zippolyte: I use the new `dogshell_with_retry`, so yeah it's not needed here anymore

tests/integration/dogshell/test_dogshell.py
zippolyte: yes we are. It's the same command, just wrapped to respect the 120 chars per line

PR #453, opened 2019-10-01
add support for SLOs

platinummonkey:
Adds support for Service Level Objectives

MA-434

### DIFF
--- a/datadog/api/api_client.py
+++ b/datadog/api/api_client.py
@@ -46,7 +46,7 @@ class APIClient(object):
 
     @classmethod
     def submit(cls, method, path, api_version=None, body=None, attach_host_name=False,
-               response_formatter=None, error_formatter=None, **params):
+               response_formatter=None, error_formatter=None, suppress_response_errors_on_codes=None, **params):
         """
         Make an HTTP API request
 
@@ -69,6 +69,10 @@ class APIClient(object):
 
         :param attach_host_name: link the new resource object to the host name
         :type attach_host_name: bool
+
+        :param suppress_response_errors_on_codes: suppress ApiError on `errors` key in the response for the given HTTP
+                                                  status codes
+        :type suppress_response_errors_on_codes: None|list(int)
 
         :param params: dictionary to be sent in the query string of the request
         :type params: dictionary
@@ -151,7 +155,10 @@ class APIClient(object):
                     raise ValueError('Invalid JSON response: {0}'.format(content))
 
                 if response_obj and 'errors' in response_obj:
-                    raise ApiError(response_obj)
+                    # suppress ApiError when specified and just return the response
+                    if not (suppress_response_errors_on_codes and
+                            result.status_code in suppress_response_errors_on_codes):
+                        raise ApiError(response_obj)
             else:
                 response_obj = None
 
@@ -177,7 +184,7 @@ class APIClient(object):
                 raise
         except ApiError as e:
             if _mute:
-                for error in e.args[0]['errors']:
+                for error in (e.args[0].get('errors') or []):
                     log.error(error)
                 if error_formatter is None:
                     return e.args[0]
--- a/datadog/dogshell/service_level_objective.py
+++ b/datadog/dogshell/service_level_objective.py
@@ -3,6 +3,7 @@ import argparse
 import json
 
 # 3p
+from datadog.util.cli import set_of_ints, comma_set, comma_list_or_empty
 from datadog.util.format import pretty_json
 
 # datadog
@@ -33,11 +34,14 @@ class MonitorClient(object):
             "--description", help="description of the SLO", default=None
         )
         create_parser.add_argument(
-            "--tags", help="comma-separated list of tags", default=None
+            "--tags",
+            help="comma-separated list of tags",
+            default=None,
+            type=comma_list_or_empty,
         )
         create_parser.add_argument(
             "--thresholds",
-            help="comma separated list of <timeframe>:<target>[:<warning>]",
+            help="comma separated list of <timeframe>:<target>[:<warning>[:<target_display>[:<warning_display>]]",
             required=True,
         )
         create_parser.add_argument(
@@ -51,10 +55,19 @@ class MonitorClient(object):
             default=None,
         )
         create_parser.add_argument(
-            "--monitor_ids", help="explicit monitor_ids to use (CSV)", default=None
+            "--monitor_ids",
+            help="explicit monitor_ids to use (CSV)",
+            default=None,
+            type=set_of_ints,
         )
         create_parser.add_argument(
             "--monitor_search", help="monitor search terms to use", default=None
+        )
+        create_parser.add_argument(
+            "--groups",
+            help="for a single monitor you can specify the specific groups as a pipe (|) delimited string",
+            default=None,
+            type=comma_list_or_empty,
         )
         create_parser.set_defaults(func=cls._create)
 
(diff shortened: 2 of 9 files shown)


datadog/dogshell/service_level_objective.py
```diff
@@ -0,0 +1,418 @@
...
+            "--denominator",
+            help="denominator metric query (sum of total events)",
+            default=None,
+        )
+        create_parser.add_argument(
+            "--monitor_ids", help="explicit monitor_ids to use (CSV)", default=None
+        )
+        create_parser.add_argument(
+            "--monitor_search", help="monitor search terms to use", default=None
+        )
```
zippolyte: Is it missing the `groups` parameter for monitor SLOs ?

datadog/dogshell/service_level_objective.py
```diff
@@ -0,0 +1,418 @@
...
+        create_parser.add_argument("--name", help="name of the SLO", default=None)
+        create_parser.add_argument(
+            "--description", help="description of the SLO", default=None
+        )
+        create_parser.add_argument(
+            "--tags", help="comma-separated list of tags", default=None
+        )
+        create_parser.add_argument(
+            "--thresholds",
+            help="comma separated list of <timeframe>:<target>[:<warning>]",
```
zippolyte: no `target_display` or `warning_display` ?

datadog/dogshell/service_level_objective.py
```diff
@@ -0,0 +1,418 @@
...
+        update_parser = verb_parsers.add_parser("update", help="Update existing SLO")
+        update_parser.add_argument(
+            "slo_id", help="SLO to replace with the new definition"
+        )
+        update_parser.add_argument(
+            "--type",
+            required=True,
+            help="type of the SLO (must specify it's original type)",
+            choices=["metric", "monitor"],
+        )
```
zippolyte: Why is this required if it mustn't change ?

datadog/dogshell/service_level_objective.py
```diff
@@ -0,0 +1,418 @@
...
+
+        if args.type == "metric":
+            params["query"] = {
+                "numerator": args.numerator,
+                "denominator": args.denominator,
+            }
+        elif args.monitor_search:
+            params["monitor_search"] = args.monitor_search
+        else:
+            params["monitor_ids"] = args.monitor_ids
```
zippolyte: Should we add some logic to make sure we use the proper arguments with the proper SLO type ?

datadog/dogshell/service_level_objective.py
```diff
@@ -0,0 +1,418 @@
...
+
+        if args.tags:
+            params["tags"] = args.tags.split(",")
+
+        res = api.ServiceLevelObjective.create(**params)
+        report_warnings(res)
+        report_errors(res)
```
zippolyte: `report_warnings` and `report_errors` won't work here, since the key in the response is `error`, not `errors`. And I don't know if you even return `warnings`

datadog/dogshell/service_level_objective.py
```diff
@@ -0,0 +1,418 @@
...
+        else:
+            print(json.dumps(res))
+
+    @classmethod
+    def _file_create(cls, args):
+        api._timeout = args.timeout
+        format = args.format
+        slo = json.load(args.file)
+
+        params = {
```
zippolyte: Shouldn't we simply pass the loaded JSON ?

datadog/dogshell/service_level_objective.py
```diff
@@ -0,0 +1,418 @@
...
+                "numerator": slo["numerator"],
+                "denominator": slo["denominator"],
+            }
+        elif slo.get("monitor_search"):
+            params["monitor_search"] = slo["monitor_search"]
+        else:
+            params["monitor_ids"] = slo["monitor_ids"]
+
+        if slo.get("tags"):
+            tags = sorted(set([t.strip() for t in slo["tags"].split(",") if t.strip()]))
```
zippolyte: In a file, this should already be a JSON list no ?

datadog/dogshell/service_level_objective.py
```diff
@@ -0,0 +1,418 @@
...
+    def _file_update(cls, args):
+        api._timeout = args.timeout
+        format = args.format
+        slo = json.load(args.file)
+
+        params = {"type": slo["type"], "name": slo["name"]}
+
+        if slo.get("thresholds"):
+            thresholds = []
+            for threshold_str in args.thresholds.split(","):
```
zippolyte: same here, thresholds should already be in a proper JSON format, like you did in `file_create`

datadog/dogshell/service_level_objective.py
platinummonkey: adding

datadog/dogshell/service_level_objective.py
platinummonkey: adding 👍

datadog/dogshell/service_level_objective.py
platinummonkey: this is required but the dogshell doesn't currently know what the current type is (though I could make it query and find out). this is more so it knows how to handle the update correctly.

datadog/dogshell/service_level_objective.py
platinummonkey: adding some typing to argument parser

datadog/dogshell/service_level_objective.py
platinummonkey: `warnings` is not currently returned, but keeping it consistent. `errors` is what is both reported and that function uses in the key

datadog/dogshell/service_level_objective.py
zippolyte: I see, let's leave like this then

datadog/dogshell/service_level_objective.py
platinummonkey: right now it throws an exception when there are errors, so wondering if this should be an arg to the service implementation 🤔

zippolyte: Actually yeah, you're right it raises an exception. I wonder how this even works in the other resources 🤔

zippolyte: Oh I see, the `APIClient.submit` method swallows the exceptions.
But in your case, since you're raising when the `error` key in the response payload is set, directly in you resource, it will not be swallowed by the super class.
Are you saying that you can have responses which contain `errors` plural, as well as the `error` singular that you use to raise exception ?

platinummonkey: yes :) in some cases the API will return actual exceptions, in other cases we return an `errors` which can contain 0 or more errors depending on the endpoint. for bulk delete, some can succeed while others not, their reasons for not succeeding will be in `errors`

typically the responses look like `{"data": ..., "errors": ...}` even for HTTP409 for example

platinummonkey: added a workaround method

datadog/api/api_client.py
```diff
@@ -177,7 +184,7 @@ def submit(cls, method, path, api_version=None, body=None, attach_host_name=Fals
                 raise
         except ApiError as e:
             if _mute:
-                for error in e.args[0]['errors']:
+                for error in (e.args[0].get('errors') or []):
```
zippolyte:
```suggestion
                for error in (e.args[0].get('errors', [])):
```

PR #459, opened 2019-10-09
Make flake8 its own CI job

nmuesch:
Change the display name of the azure pipeline job to be the TOKENV being run

### DIFF
--- a/.azure-pipelines/unit_tests.yml
+++ b/.azure-pipelines/unit_tests.yml
@@ -36,9 +36,12 @@ jobs:
       Pypy3:
         PYTHON_VERSION: "pypy3"
         TOXENV: "pypy3"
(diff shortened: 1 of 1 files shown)


.azure-pipelines/unit_tests.yml
```diff
@@ -36,6 +36,9 @@ jobs:
       Pypy3:
         PYTHON_VERSION: "pypy3"
         TOXENV: "pypy3"
+      Lint:
+        PYTHON_VERSION: "3.5"
+        TOXENV: "flake8"
```
zippolyte: Let's also remove it from the Py35 env then

nmuesch: Done

Issue #269, opened 2018-05-08
`str` object has no attribute `get` error on metric upload


Jbithell:
After initialisation, following code returns error below. 
```
single_metric = {
    'metric': 'uploadSpeed',
    'type': 'gauge',
    'points': [speedData[0], speedData[3]]
}
api.Metric.send(metrics="uploadSpeed", **single_metric)
```
```
File "main.py", line 25, in sendData
	api.Metric.send(metrics="uploadSpeed", **single_metric)
File "venv\lib\site-packages\datadog\api\metrics.py", line 140, in send
	return super(Metric, cls).send(attach_host_name=True, **metrics_dict)
File "venv\lib\site-packages\datadog\api\resources.py", line 74, in send
	attach_host_name=attach_host_name)
File "venv\lib\site-packages\datadog\api\api_client.py", line 99, in submit
	if obj_params.get('host', "") == "":
		AttributeError: 'str' object has no attribute 'get'
```

dmangot:
Same issue on OpenBSD 6.3

dabcoder:
Apologies for the delay in updating this thread. We managed to reproduce this issue, and we're currently looking into it.

dabcoder:
We've addressed this issue here: #463. Basically `metrics` is a list of dictionaries, so this why the above code snippet resulted in this error @Jbithell (cc @dmangot).

PR #463, opened 2019-10-16
Clarify docstring for metrics

dabcoder:
Addresses https://github.com/DataDog/datadogpy/issues/269.

### DIFF
--- a/datadog/api/metrics.py
+++ b/datadog/api/metrics.py
@@ -54,6 +54,12 @@ class Metric(SearchableAPIResource, SendableAPIResource, ListableAPIResource):
         :param metric: the name of the time series
         :type metric: string
 
+        :param single_metric: a list of metric names
+        :type metric: list
+
+        :param metrics: a list of single_metric
+        :type metric: list
+
         :param points: a (timestamp, value) pair or list of (timestamp, value) pairs
         :type points: list
 
@@ -65,6 +71,14 @@ class Metric(SearchableAPIResource, SendableAPIResource, ListableAPIResource):
 
         :param type: type of the metric
         :type type: 'gauge' or 'count' or 'rate' string
+
+        >>> api.Metric.send(metric='my.series', points=[(now, 15), (future_10s, 16)])
+
+        >>> api.Metric.send(single_metric=['my.series', 'my.series2'], points=[(now, 15), (future_10s, 16)])
+
+        >>> single_metric = [{'metric': 'my.series', 'type': 'gauge', 'points': [(now, 15), (future_10s, 16)]},
+        >>> {'metric': 'my.series', 'type': 'gauge', 'points': [(now, 15), (future_10s, 16)]}]
+        >>> api.Metric.send(metrics=single_metric)
 
         :returns: Dictionary representing the API's JSON response
         """
--- a/datadog/api/metrics.py
+++ b/datadog/api/metrics.py
@@ -50,15 +50,14 @@ class Metric(SearchableAPIResource, SendableAPIResource, ListableAPIResource):
     def send(cls, metrics=None, attach_host_name=True, **single_metric):
         """
         Submit a metric or a list of metrics to the metric API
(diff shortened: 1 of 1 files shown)


datadog/api/metrics.py
```diff
@@ -54,6 +54,12 @@ def send(cls, metrics=None, attach_host_name=True, **single_metric):
         :param metric: the name of the time series
         :type metric: string
 
+        :param single_metric: a list of metric names
+        :type metric: list
```
zippolyte: That's not it. `single_metric` is just name of the kwargs argument, but it should just consist of 5 arguments (some of which are optional): `metric`, `points`, `host`, `tags` and `type`, which are documented just below.
`metrics` should be a list of dictionaries that follow the `single_metric` format, i.e. dict containing the same 5 keys than `single_metric`.

In any case, `single_metric` should not be documented as is, we already document each keyword argument that it should contain

datadog/api/metrics.py
```diff
@@ -54,6 +54,12 @@ def send(cls, metrics=None, attach_host_name=True, **single_metric):
...
+
+        :param metrics: a list of single_metric
+        :type metric: list
```
zippolyte:
```suggestion
        :type metrics: list
```

datadog/api/metrics.py
```diff
@@ -54,6 +54,12 @@ def send(cls, metrics=None, attach_host_name=True, **single_metric):
         :param metric: the name of the time series
         :type metric: string
 
+        :param single_metric: a list of metric names
+        :type metric: list
+
+        :param metrics: a list of single_metric
```
zippolyte:
```suggestion
        :param metrics: a list of dictionaries, each item being a metric to send
```

datadog/api/metrics.py
```diff
@@ -66,6 +72,14 @@ def send(cls, metrics=None, attach_host_name=True, **single_metric):
         :param type: type of the metric
         :type type: 'gauge' or 'count' or 'rate' string
 
+        >>> api.Metric.send(metric='my.series', points=[(now, 15), (future_10s, 16)])
+
+        >>> api.Metric.send(single_metric=['my.series', 'my.series2'], points=[(now, 15), (future_10s, 16)])
```
zippolyte: This is wrong too. Let's just remove it
```suggestion
```

datadog/api/metrics.py
```diff
@@ -66,6 +72,14 @@ def send(cls, metrics=None, attach_host_name=True, **single_metric):
...
+
+        >>> single_metric = [{'metric': 'my.series', 'type': 'gauge', 'points': [(now, 15), (future_10s, 16)]},
```
zippolyte:
```suggestion
        >>> metrics = [{'metric': 'my.series', 'type': 'gauge', 'points': [(now, 15), (future_10s, 16)]},
```

datadog/api/metrics.py
```diff
@@ -54,6 +54,12 @@ def send(cls, metrics=None, attach_host_name=True, **single_metric):
         :param metric: the name of the time series
```
zippolyte: let's just describe more explicitly in the main description above what a metric dict should look like

datadog/api/metrics.py
```diff
@@ -66,6 +71,12 @@ def send(cls, metrics=None, attach_host_name=True, **single_metric):
         :param type: type of the metric
         :type type: 'gauge' or 'count' or 'rate' string
 
+        >>> api.Metric.send(metric='my.series', points=[(now, 15), (future_10s, 16)])
+
+        >>> metrics = [{'metric': 'my.series', 'type': 'gauge', 'points': [(now, 15), (future_10s, 16)]},
+        >>> {'metric': 'my.series', 'type': 'gauge', 'points': [(now, 15), (future_10s, 16)]}]
```
zippolyte:
```suggestion
                {'metric': 'my.series2', 'type': 'gauge', 'points': [(now, 15), (future_10s, 16)]}]
```

PR #468, opened 2019-10-24
Fix keys in headers for some endpoints

dabcoder:
Attempts to fix integration test failures. Follow up from #446.
Also assess that keys aren't in both params and headers.

### DIFF
--- a/tests/unit/api/test_api.py
+++ b/tests/unit/api/test_api.py
@@ -148,7 +148,8 @@ class TestInitialization(DatadogAPINoInitialization):
         self.assertEqual(options['headers']['Content-Type'], 'application/json')
         self.assertEqual(options['headers']['DD-API-KEY'], API_KEY)
         self.assertEqual(options['headers']['DD-APPLICATION-KEY'], APP_KEY)
-
+        self.assertIsNone(options['params'].get('api_key'))
+        self.assertIsNone(options['params'].get('application_key'))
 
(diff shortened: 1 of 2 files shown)


tests/unit/api/test_api.py
```diff
@@ -148,7 +148,8 @@ def test_request_parameters(self):
         self.assertEqual(options['headers']['Content-Type'], 'application/json')
         self.assertEqual(options['headers']['DD-API-KEY'], API_KEY)
         self.assertEqual(options['headers']['DD-APPLICATION-KEY'], APP_KEY)
-
+        self.assertIsNone(options['params'].get('api_key'))
+        self.assertIsNone(options['params'].get('application_key'))
```
zippolyte: I would instead use `assert "api_key" not in options`

tests/unit/api/test_api.py
```diff
@@ -172,8 +173,8 @@ def test_request_parameters_api_keys_in_params(self):
         self.assertEqual(options['headers']['Content-Type'], 'application/json')
         self.assertEqual(options['params']['api_key'], API_KEY)
         self.assertEqual(options['params']['application_key'], APP_KEY)
-
-
+        self.assertIsNone(options['headers'].get('DD-API-KEY'))
```
zippolyte: same let's verify the key is not present at all with `assert not in`

PR #471, opened 2019-10-30
Add warning option for dogwrap based on exit codes

dabcoder:
Add a warning option based on exit codes provided by the user.
Warning exit codes should be a comma separated list such as `--warning_codes 123,345,678`.

### DIFF
--- a/datadog/dogshell/wrap.py
+++ b/datadog/dogshell/wrap.py
@@ -19,7 +19,8 @@ dogwrap -n test-job -k $API_KEY --timeout=1 "sleep 3"
 '''
 # stdlib
 from __future__ import print_function
-import optparse
+import argparse
+import os
 import subprocess
 import sys
 import threading
@@ -33,6 +34,7 @@ from datadog.util.compat import is_p3k
 
 SUCCESS = 'success'
 ERROR = 'error'
+WARNING = 'warning'
 
 MAX_EVENT_BODY_LENGTH = 3000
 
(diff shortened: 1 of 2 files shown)


datadog/dogshell/wrap.py
```diff
@@ -289,11 +295,25 @@ def main():
     initialize(api_key=options.api_key)
     host = api._host_name
 
+    if options.warning_codes:
+        warning_codes = [int(w.strip()) for w in options.warning_codes.split(',')]
```
jd: You should write this as a function and use it as the type for your `add_option` call. It'll automatically be called when needed.

datadog/dogshell/wrap.py
```diff
@@ -219,25 +221,42 @@ def build_event_body(cmd, returncode, stdout, stderr, notifications):
...
     Parse the raw command line options into an options object and the remaining command string
     '''
     parser = optparse.OptionParser(usage="%prog -n [event_name] -k [api_key] --submit_mode \
-[ all | errors ] [options] \"command\". \n\nNote that you need to enclose your command in \
+[ all | errors | warnings] [options] \"command\". \n\nNote that you need to enclose your command in \
 quotes to prevent python executing as soon as there is a space in your command. \n \nNOTICE: In \
 normal mode, the whole stderr is printed before stdout, in flush_live mode they will be mixed but \
 there is not guarantee that messages sent by the command on both stderr and stdout are printed in \
-the order they were sent.", version="%prog {0}".format(get_version()))
+the order they were sent.", version="%prog {0}".format(get_version()), option_class=WarningOption)
```
jd: I doubt you want to replace the whole option class for all options here, right?

datadog/dogshell/wrap.py
```diff
@@ -219,25 +221,42 @@ def build_event_body(cmd, returncode, stdout, stderr, notifications):
...
 as it should appear on your Datadog stream")
     parser.add_option('-k', '--api_key', action='store', type='string',
                       help="your DataDog API Key")
     parser.add_option('-m', '--submit_mode', action='store', type='choice',
-                      default='errors', choices=['errors', 'all'], help="[ all | errors ] if set \
+                      default='errors', choices=['errors', 'warnings', 'all'], help="[ all | errors | warnings ] if set \
 to error, an event will be sent only of the command exits with a non zero exit status or if it \
-times out.")
+times out. If set to warning, a list of exit codes need to be provided")
+    parser.add_option('--warning_codes', action='store', type='warning_codes', dest='warning_codes', default='',
```
jd: That's where you could use your `WarningOption` class, but I'm not sure it's needed.

Using `type=warning_codes` should be enough.

(ok I admit I forgot optparse but I expect it to be close to argparse :))

datadog/dogshell/wrap.py
dabcoder: Yeah that's incorrect, need to revert to how it previously was.

dabcoder: Actually, I just tested this and the `WarningOption` class should take the default options and just add the `warning_codes` one to the whole option parser, so it looks like this is required here. Otherwise it'll throw:

```
optparse.OptionError: option --warning_codes: invalid option type: 'warning_codes'
```

datadog/dogshell/wrap.py
dabcoder: > Using type=warning_codes should be enough.

Yes `WarningOption` should be defined at the parser global config level.

datadog/dogshell/wrap.py
jd: Oh ok, that might be an `optparse` specific thing, damn. I guess that's fine then if that's the way you're supposed to add option. :)

datadog/dogshell/wrap.py
```diff
@@ -219,25 +221,44 @@ def build_event_body(cmd, returncode, stdout, stderr, notifications):
...
+    try:
+        # options_warning is a string e.g.: --warning_codes 123,456,789
+        # we need to create a list from it
+        warning_codes = options_warning.split(",")
+        return warning_codes
+    except ValueError:
+        raise optparse.OptionValueError("option %s: invalid warning codes value(s): %r" % (opt, options_warning))
+
+
+class WarningOption(optparse.Option):
```
jd: The type name might not be that good since it's going to be your option type with all your types.
`DogshellOption` might be better?

dabcoder: Since it's for `dogwrap` only, we may even want to call this `DogwrapOption`.

PR #475, opened 2019-11-04
Convert setup.py to setup.cfg

jd:


jd:
@ofek So I took a look at doing so, and it might not be a good time yet.

My next PR was going to leverage `[setuptools_scm](https://pypi.org/project/setuptools-scm/)` which is pretty powerful and that we already leverage in https://github.com/DataDog/dd-trace-py with great success.

Poetry does not allow that — worse, it has 0 plugin support (it's still WIP) so there's no way to add such functionality for now.

The https://github.com/psf/black project switched back from Poetry to setuptools for that exact reason — they also leverage `setuptools_scm` so they reverted from using Poetry.

I'll update this PR to include a `pyproject.toml` with basic information, but for the time being I'd advise keeping setuptools.

jd:
I'm merging this. If it ever breaks somebody workflow in the future, we can still revert anyway 🤞

PR #476, opened 2019-11-04
Upgrade flake8 to 3.7.9

jd:


datadog/dogshell/dashboard.py
```diff
@@ -81,7 +81,7 @@ def _post(cls, args):
             widgets = sys.stdin.read()
         try:
             widgets = json.loads(widgets)
-        except:
+        except Exception:
```
dabcoder: Thanks for this. 
Should we raise specific exceptions there (and below too) as opposed to plain ones?

jd: Don't start me on this 😆 

The way this is written and done is indeed pretty bad. I've chosen to ignore it for now as it's not the purpose of this patch, but those exception handlings need serious cleanups indeed.

I'll make a second PR later to clean this up if that's good with you.

dabcoder: :). Fair enough, yes I agree it'd need to be a different PR. 

> I'll make a second PR later to clean this up if that's good with you.

Can also look into this if you don't have time. Thanks.

tox.ini
```diff
@@ -28,7 +28,7 @@ commands =
 [testenv:flake8]
 skip_install = true
 deps =
-    flake8==3.4.1
+    flake8==3.7.7
```
dabcoder: `flake8` is now at version `3.7.9`. Should we leave it at 3.7.7?

jd: No, my bad. Updating.

jd:
Update flake8 to 3.7.9

Issue #479, opened 2019-11-13
403 Forbidden on Threadstats.distribution submit


whisust:
Currently on `datadog` version 0.31.0, you can't send distributions through `ThreadStats`.

```python
from datadog import initialize, ThreadStats
import time
import logging

logging.basicConfig(level='DEBUG')
logger = logging.getLogger(__name__)

DATADOG_API_KEY = "<api_key>"
DATADOG_APP_KEY = "<app_key>"

initialize(api_key=DATADOG_API_KEY, app_key=DATADOG_APP_KEY)

client = ThreadStats()

logger.info('Starting datadog threadstats reporter')
client.start(flush_interval=5, roll_up_interval=5)

# 403 on submit
client.distribution('my-distrib', 1)
time.sleep(10)

# 202 On submit
client.histogram('my-histo', 1)
time.sleep(10)
```

Outputs:
```
WARNING:datadog.api:No agent or invalid configuration file found
INFO:__main__:Starting datadog threadstats reporter
INFO:datadog.threadstats:Starting flush thread with interval 5.
DEBUG:datadog.threadstats:Flushing metrics in thread
DEBUG:datadog.threadstats:No metrics to flush. Continuing.
DEBUG:datadog.threadstats:Flush #1 sending 1 distributions
DEBUG:datadog.api:Use `requests` based HTTP client.
DEBUG:urllib3.util.retry:Converted retries value: 3 -> Retry(total=3, connect=None, read=None, redirect=None, status=None)
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): api.datadoghq.com
DEBUG:urllib3.connectionpool:https://api.datadoghq.com:443 "POST /api/v1/distribution_points HTTP/1.1" 403 20
INFO:datadog.api:403 POST https://api.datadoghq.com/api/v1/distribution_points (416.9381ms)
ERROR:datadog.threadstats:Error flushing metrics and events
Traceback (most recent call last):
  File "/Users/antoinel/venvs/jellynote/lib/python2.7/site-packages/datadog/threadstats/base.py", line 324, in flush
    self.reporter.flush_distributions(dists)
  File "/Users/antoinel/venvs/jellynote/lib/python2.7/site-packages/datadog/threadstats/reporters.py", line 21, in flush_distributions
    api.Distribution.send(distributions, compress_payload=self.compress_payload)
[...]

PR #480, opened 2019-11-14
Set API Key in URL when submit distribution metrics

tianchu:
`api.Distribution.send` no longer works after https://github.com/DataDog/datadogpy/pull/446, as endpoint `v1/distribution_points` still expects the api key in the URL parameter than header.

To reproduce:
```python
from datadog import initialize, api
options = {
    'api_key':'<API-KEY>',
    'app_key':'<APP-KEY>'
}
initialize(**options)

# submitting a regular metric point works
api.Metric.send(metric='my.test.metric', points=1)

# submitting a distribution metric point gets error 'No api key specified'
api.Distribution.send(distributions=[{
    'metric': 'my.dist.test',
    'points': [(1573757950, [1.0])],
    'host': None,
}])
```

### DIFF
--- a/datadog/api/api_client.py
+++ b/datadog/api/api_client.py
@@ -273,6 +273,7 @@ class APIClient(object):
         constructed_path = construct_path(api_version, path)
 
         set_of_paths = {
+            "v1/distribution_points",
             "v1/series",
             "v1/check_run",
             "v1/events",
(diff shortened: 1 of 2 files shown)


zippolyte:
Looks like this should solve https://github.com/DataDog/datadogpy/issues/479. Do you confirm @tianchu

zippolyte:
The change looks good to me. This wasn't caught because we don't have integration tests for distributions. Would you mind adding some ?

tianchu:
> Looks like this should solve #479. Do you confirm @tianchu

Correct, it should solve $479.

tianchu:
> The change looks good to me. This wasn't caught because we don't have integration tests for distributions. Would you mind adding some ?

Makes sense. Let me look into it.

PR #483, opened 2019-11-19
Remove __init__.py files from tests/

jd:
This makes sure the tests directory is not considered as a Python package.


jd:
The default mechanism for setup.cfg to find package will make it find `tests` as a package because of the `__init__` file. So it'll try to install it as a package.
Alternatively, we can just exclude those.
(credit @jirikuncar)

PR #487, opened 2019-11-27
[MA-665]  Add support for the `Monitor.validate` endpoint.

armcburney:
@DataDog/monitor-app 

___

The `validate` endpoint can be used to check if a monitor definition is valid.

### DIFF
--- a/datadog/api/resources.py
+++ b/datadog/api/resources.py
@@ -384,15 +384,15 @@ class ActionAPIResource(object):
     Actionable API Resource
     """
     @classmethod
-    def _trigger_class_action(cls, method, name, id=None, params=None, **body):
+    def _trigger_class_action(cls, method, action_name, id=None, params=None, **body):
         """
         Trigger an action
 
         :param method: HTTP method to use to contact API endpoint
         :type method: HTTP method string
 
-        :param name: action name
-        :type name: string
+        :param action_name: action name
+        :type action_name: string
 
         :param id: trigger the action for the specified resource object
         :type id: id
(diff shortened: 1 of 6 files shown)


datadog/api/resources.py
```diff
@@ -384,15 +384,15 @@ class ActionAPIResource(object):
     Actionable API Resource
     """
     @classmethod
-    def _trigger_class_action(cls, method, name, id=None, params=None, **body):
+    def _trigger_class_action(cls, method, action_name, id=None, params=None, **body):
```
armcburney: I renamed the `name` parameter to `action_name` because when you splat the `**body` with a `name` in the dict, it causes there to be a conflict because of two `name` parameters.

datadog/dogshell/monitor.py
```diff
@@ -124,6 +124,19 @@ def setup_parser(cls, subparsers):
                                        help="monitors to check if they can be deleted")
         can_delete_parser.set_defaults(func=cls._can_delete)
 
+        validate_parser = verb_parsers.add_parser('validate',
+                                                  help="Validates if a monitor definition is correct")
+        validate_parser.add_argument('type', help="type of the monitor, e.g."
+                                     "'metric alert' 'service check'")
+        validate_parser.add_argument('query', help="query to notify on with syntax varying "
```
athap: `query to notify on with syntax varying` may be just `monitor query`?

datadog/api/resources.py
zippolyte: Test is failing for `embed` because of that name change

datadog/dogshell/monitor.py
```diff
@@ -128,7 +128,7 @@ def setup_parser(cls, subparsers):
                                                   help="Validates if a monitor definition is correct")
         validate_parser.add_argument('type', help="type of the monitor, e.g."
                                      "'metric alert' 'service check'")
-        validate_parser.add_argument('query', help="query to notify on with syntax varying "
+        validate_parser.add_argument('query', help="the monitor query"
                                      "depending on what type of monitor you are creating")
```
zippolyte: Now the sentence doesn't make much sense. Let's do something like `monitor query, the syntax varies depending on the type of the monitor`

armcburney:
@zippolyte, the integration test failure was due to an unrelated test, which has been flaky. Is it okay to merge this?

PR #488, opened 2019-11-27
[MA-669] Add support for `/v1/downtime/cancel/by_scope`.

armcburney:
@DataDog/monitor-app 

___

https://docs.datadoghq.com/api/?lang=python#cancel-monitor-downtimes-by-scope

### DIFF
--- a/datadog/dogshell/downtime.py
+++ b/datadog/dogshell/downtime.py
@@ -44,14 +44,13 @@ class DowntimeClient(object):
         show_parser.add_argument('downtime_id', help="downtime to show")
         show_parser.set_defaults(func=cls._show_downtime)
 
-        show_all_parser = verb_parsers.add_parser('show_all', help="Show a list of all downtimes")
-        show_all_parser.add_argument('--current_only', help="only return downtimes that"
-                                     " are active when the request is made", default=None)
-        show_all_parser.set_defaults(func=cls._show_all_downtime)
-
         delete_parser = verb_parsers.add_parser('delete', help="Delete a downtime")
         delete_parser.add_argument('downtime_id', help="downtime to delete")
         delete_parser.set_defaults(func=cls._cancel_downtime)
+
+        cancel_parser = verb_parsers.add_parser('cancel_by_scope', help="Cancel all downtimes with a given scope")
+        cancel_parser.add_argument('scope', help="The scope of the downtimes to cancel")
+        cancel_parser.set_defaults(func=cls._cancel_downtime_by_scope)
 
     @classmethod
     def _schedule_downtime(cls, args):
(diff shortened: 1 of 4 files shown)


datadog/dogshell/downtime.py
```diff
@@ -44,15 +44,14 @@ def setup_parser(cls, subparsers):
         show_parser.add_argument('downtime_id', help="downtime to show")
         show_parser.set_defaults(func=cls._show_downtime)
 
-        show_all_parser = verb_parsers.add_parser('show_all', help="Show a list of all downtimes")
-        show_all_parser.add_argument('--current_only', help="only return downtimes that"
-                                     " are active when the request is made", default=None)
-        show_all_parser.set_defaults(func=cls._show_all_downtime)
-
```
athap: Are these not needed anymore?

datadog/dogshell/downtime.py
```diff
@@ -100,12 +99,10 @@ def _show_downtime(cls, args):
             print(json.dumps(res))
 
     @classmethod
-    def _show_all_downtime(cls, args):
```
athap: we don't support show_all_downtime anymore?

datadog/dogshell/downtime.py
armcburney: The dogshell `downtime` command is currently broken because we don't support the `show_all` downtimes using the client. I removed this to implement the `cancel/by_scope` endpoint. I can fix the `show_all` command, but could I do that in a separate PR?

zippolyte: Sure let's fix that in a separate PR, but why remove it at all in this one ? Was it preventing you from doing anything ?

armcburney: Yeah, the dogshell `dog downtime` command won't run currently.

zippolyte: Hmm I just tried from master the `show_all` command, and it works. Maybe something wrong with your setup ?

tests/integration/api/test_api.py
```diff
@@ -649,6 +649,51 @@ def test_downtime(self):
...
+        # Cancel downtimes with scope `scope_one`
+        dog.Downtime.cancel_downtime_by_scope(scope=scope_one)
+
+        # Verify only the downtimes with scope `scope_one` were canceled
+        for downtime in downtimes_with_scope_one:
+            get_with_retry(
+                "Downtime", downtime["id"], retry_condition=lambda r: r["disabled"] is False
+            )
+        for downtime in downtimes_with_scope_two:
+            get_with_retry("Downtime", downtime["id"])
```
nmuesch: Should we assert here that the response's `disabled` field is still marked as `False`?

tests/integration/api/test_api.py
```diff
@@ -649,6 +649,51 @@ def test_downtime(self):
...
         )
 
+    def test_downtime_cancel_by_scope(self):
+        scope_one = "test:integration_one"
+        scope_two = "test:integration_two"
+        start = int(time.time())
+
+        # Create downtime with scope_one
+        end = start + 1000
+        downtime_one = dog.Downtime.create(scope=scope_one, start=start, end=end)
```
nmuesch: Thinking about subsequent test runs here. Would `creating` a downtime that already exists and is disabled just flip the `disabled` flag back to `False`?

armcburney: Ahh no, it will create a separate downtime with a different `downtime["id"]`.

armcburney: Sorry, to clarify: it creates a new one because the `start` and `end` will be different depending on the test execution - so you schedule it for a different time period.

nmuesch: Ahh ok that sounds good to me.

Issue #485, opened 2019-11-20
Tags with commas incorrectly treated as multiple tags


felixc:
It appears that commas in tags are incorrectly treated as tag separators when sent over the API. This results in a list of tags like `["foo:bar, baz", "quux:quuz, frob"]` being misinterpreted as `["foo:bar", "baz", "quux:quuz", "frob"]`.

Although I understand that the API uses commas to separate values in your serialization format, I believe that since we are passing in a clear list structure, this behaviour is unexpected and incorrect. I would hope that commas within tags would be escaped and treated as just commas; or if that's not possible, ignored/replaced with a "safe" character.

The relevant lines may be these ones (but I'm not familiar with the codebase so I can't say for sure): https://github.com/DataDog/datadogpy/blob/09c0a953d47675777ab57f3840b4b04ef378ff4d/datadog/api/api_client.py#L134-L136

I would suggest also documenting this restriction here: https://docs.datadoghq.com/tagging/#defining-tags

For what it's worth, I believe this was changed in the Ruby client some time ago: https://github.com/DataDog/dogstatsd-ruby/blob/master/CHANGELOG.md#tags

dabcoder:
Thanks for reporting this @felixc. This should be solved with #489. Let us know if this still causes issues, in which case you can reopen this thread.

PR #489, opened 2019-11-28
Tag normalization

dabcoder:
Addresses https://github.com/DataDog/datadogpy/issues/485. Manually tested too.

### DIFF
--- a/datadog/util/format.py
+++ b/datadog/util/format.py
@@ -20,3 +20,11 @@ def force_to_epoch_seconds(epoch_sec_or_dt):
     if isinstance(epoch_sec_or_dt, datetime.datetime):
         return calendar.timegm(epoch_sec_or_dt.timetuple())
     return epoch_sec_or_dt
+
+
+def normalize_tags(tag_list):
+    final_tag_list = []
+    for tag in tag_list:
+        # remove commas and whitespaces from the tag
+        final_tag_list.append(tag.replace(",", "_").replace(" ", ""))
+    return final_tag_list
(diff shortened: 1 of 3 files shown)


datadog/util/format.py
```diff
@@ -20,3 +20,11 @@ def force_to_epoch_seconds(epoch_sec_or_dt):
     if isinstance(epoch_sec_or_dt, datetime.datetime):
         return calendar.timegm(epoch_sec_or_dt.timetuple())
     return epoch_sec_or_dt
+
+
+def normalize_tags(tag_list):
+    final_tag_list = []
+    for tag in tag_list:
+        # remove commas and whitespaces from the tag
+        final_tag_list.append(tag.replace(",", "_").replace(" ", ""))
```
zippolyte: I would do the minimum client side, and let the API deal with everything. i.e. I would not touch the whitespace if not needed, just replace the comma

dabcoder: Fair enough, will do thanks.

datadog/util/format.py
```diff
@@ -20,3 +20,11 @@ def force_to_epoch_seconds(epoch_sec_or_dt):
...
+    return final_tag_list
```
ofek: For performance, please

```python
def normalize_tags(tag_list):
    return [tag.replace(',', '_') for tag in tag_list]
```

dabcoder:
@jd Actually I looked at it a bit more and it is possible to add a parameter `constant_tags`: https://github.com/DataDog/datadogpy/blob/80c9603381187e43a900e60c8ef4ac9ff019a857/datadog/dogstatsd/base.py#L60-L61 (A list of strings). Unsure if we'd need to make the same change there though.

PR #494, opened 2019-12-11
Add statsd_constant_tags kwarg to datadog.initialize()

kainswor:
Per the documentation here https://docs.datadoghq.com/developers/dogstatsd/?tab=python#client-instantiation-parameters
It seems like you should be able to pass constant tags to the default statsd via the datadog initialize method. This just adds it and merges the list already in the default datadog.dogstatsd.statsd class similar to how the other statsd_* kwargs are handled


dabcoder:
Thanks for working on this @kainswor

PR #496, opened 2019-12-12
feat(api): base class for all exceptions

hakamadare:
This makes it easier to do the following:
```python
try:
    # something
except datadog.api.exceptions.DatadogException:
    # catch all the Datadog-specific exceptions
```

### DIFF
--- a/datadog/api/exceptions.py
+++ b/datadog/api/exceptions.py
@@ -12,8 +12,6 @@ class DatadogException(Exception):
         except datadog.api.exceptions.DatadogException:
             # handle any Datadog-specific exceptions
     """
-    def __init__(self, message, *args, **kwargs):
-        super(DatadogException, self).__init__(message, *args, **kwargs)
 
 
 class ProxyError(DatadogException):
--- a/datadog/api/exceptions.py
+++ b/datadog/api/exceptions.py
@@ -3,7 +3,20 @@ API & HTTP Clients exceptions.
 """
 
 
-class ProxyError(Exception):
+class DatadogException(Exception):
+    """
+    Base class for Datadog API exceptions.  Use this for patterns like the following:
+
+        try:
+            # do something with the Datadog API
+        except datadog.api.exceptions.DatadogException:
+            # handle any Datadog-specific exceptions
+    """
+    def __init__(self, message, *args, **kwargs):
+        super(DatadogException, self).__init__(message, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


datadog/api/exceptions.py
```diff
@@ -3,7 +3,20 @@
...
+    """
+    Base class for Datadog API exceptions.  Use this for patterns like the following:
+
+        try:
+            # do something with the Datadog API
+        except datadog.api.exceptions.DatadogException:
+            # handle any Datadog-specific exceptions
+    """
+    def __init__(self, message, *args, **kwargs):
+        super(DatadogException, self).__init__(message, *args, **kwargs)
```
jd: Does not seem like `message` is useful here, I'd just remove this custom `__init__` function which does nothing.

hakamadare: @jd the point of this `__init__` is to permit invocation with keyword arg `message=<bla>`, i've added some additional tests to demonstrate this.  if you feel strongly about it i'll remove it, let me know what you think

jd: I'm fine with it if you want message to be an argument, but since you're not using it I really don't see the value.

datadog/api/exceptions.py
```diff
@@ -3,7 +3,20 @@
...
+class DatadogException(Exception):
+    """
+    Base class for Datadog API exceptions.  Use this for patterns like the following:
+
+        try:
+            # do something with the Datadog API
+        except datadog.api.exceptions.DatadogException:
+            # handle any Datadog-specific exceptions
+    """
+    def __init__(self, *args, message='', **kwargs):
```
jd: Empty message as default for an exception is not a good idea.

If you want to force having a message I'd prefer having it the way it was before then.

hakamadare: ok, no worries :) i've removed the `__init__()` (turns out i had introduced a bug in my last commit anyway)

hakamadare:
@jd it's not clear to me whether the integration test failure has anything to do with what i changed; do you have any thoughts about it?  would you be able to give me the ability to trigger integration test runs so that i can add some debugging info and try to figure it out?

dabcoder:
@hakamadare 

> it's not clear to me whether the integration test failure has anything to do with what i changed;

This test failure is indeed unrelated to your PR, we are aware of this issue.

PR #498, opened 2019-12-13
Update histogram in-line description

fanny-jiang:
Histogram metrics currently do not send the median aggregate: https://github.com/DataDog/datadogpy/blob/48a5a677c58bc7282a25061d811f445329ad1f17/datadog/threadstats/metrics.py#L86-L138

This PR corrects the in-line description of the Histogram function.

### DIFF
--- a/datadog/threadstats/base.py
+++ b/datadog/threadstats/base.py
@@ -203,7 +203,7 @@ class ThreadStats(object):
         """
         Sample a histogram value. Histograms will produce metrics that
         describe the distribution of the recorded values, namely the maximum,
-        median, average, count and the 95th percentile. Optionally, specify
+        average, count and the 95th percentile. Optionally, specify
         a list of ``tags`` to associate with the metric.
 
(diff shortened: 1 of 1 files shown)


datadog/threadstats/base.py
```diff
@@ -203,7 +203,7 @@ def histogram(self, metric_name, value, timestamp=None, tags=None, sample_rate=1
         """
         Sample a histogram value. Histograms will produce metrics that
         describe the distribution of the recorded values, namely the maximum,
-        median, average, count and the 95th percentile. Optionally, specify
+        average, count and the 95th percentile. Optionally, specify
```
nmuesch: Thanks, looks like we support a couple that aren't here as well, should we call those out - https://github.com/DataDog/datadogpy/blob/master/datadog/threadstats/metrics.py#L95-L102 ?

fanny-jiang: That sounds like a good idea. Thanks for checking!

PR #500, opened 2019-12-18
include LICENSE in MANIFEST.in

jjhelmus:
Add entry to MANIFEST.in so that the LICENSE file is included in sdist and whl artifacts.


jjhelmus:
Including the license file in the sdist is helpful for downstream packagers, such as conda-forge, who wish to include this file in their artifacts.

PR #502, opened 2019-12-26
Add EU endpoint

dabcoder:
As `dogwrap` does not currently have an option to choose the site.

### DIFF
--- a/datadog/dogshell/wrap.py
+++ b/datadog/dogshell/wrap.py
@@ -253,6 +253,8 @@ the order they were sent.", version="%prog {0}".format(get_version()), option_cl
 as it should appear on your Datadog stream")
     parser.add_option('-k', '--api_key', action='store', type='string',
                       help="your DataDog API Key")
+    parser.add_option('-s', '--site', action='store', type='choice', default='us', choices=['us', 'eu'], help="The site \
+to send data, US or EU, default: US")
     parser.add_option('-m', '--submit_mode', action='store', type='choice',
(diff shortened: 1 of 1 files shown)


datadog/dogshell/wrap.py
```diff
@@ -310,7 +312,12 @@ def main():
         options.sigterm_timeout, options.sigkill_timeout,
         options.proc_poll_interval, options.buffer_outs)
 
-    initialize(api_key=options.api_key)
+    if options.site.lower() == 'eu':
```
jd: `lower()` should not be needed if `choices` is enforced. :)

datadog/dogshell/wrap.py
```diff
@@ -253,6 +253,8 @@ def parse_options(raw_args=None):
 as it should appear on your Datadog stream")
     parser.add_option('-k', '--api_key', action='store', type='string',
                       help="your DataDog API Key")
+    parser.add_option('-s', '--site', action='store', type='choice', default='us', choices=['us', 'eu'], help="The site \
```
zippolyte: Hmm, for this I would have kept consistency with the agent config and gave the choices `datadoghq.eu` or `datadoghq.com`

dabcoder: So your suggested change would be `default='datadoghq.com', choices=['datadoghq.com', 'datadoghq.eu']`?

jd: Maybe support both syntaxes?
I mean, it's long to type the long hostname on a CLI.

PR #504, opened 2019-12-30
Add datadoghq for consistency with agent options

dabcoder:


datadog/dogshell/wrap.py
```diff
@@ -312,7 +312,7 @@ def main():
         options.sigterm_timeout, options.sigkill_timeout,
         options.proc_poll_interval, options.buffer_outs)
 
-    if options.site == 'eu':
+    if options.site == 'datadoghq.eu' or options.site == 'eu':
```
jd:
```suggestion
    if options.site in ('datadoghq.eu', 'eu'):
```

PR #505, opened 2019-12-30
Adding telemetry to dogstatsd

hush-hush:
Adding base telemetry to the Python dogstastd. This allow troubleshooting of packet drop and performance issue by users.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -17,6 +17,7 @@ from threading import Lock
 from datadog.dogstatsd.context import TimedContextManagerDecorator
 from datadog.dogstatsd.route import get_default_route
 from datadog.util.compat import text
+from datadog.util.config import get_version
 
 # Logging
 log = logging.getLogger('datadog.dogstatsd')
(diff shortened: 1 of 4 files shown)


datadog/dogstatsd/base.py
```diff
@@ -312,40 +330,60 @@ def _report(self, metric, metric_type, value, tags, sample_rate):
...
+            pass
         except (socket.error, socket.herror, socket.gaierror) as se:
             log.warning("Error submitting packet: {}, dropping the packet and closing the socket".format(se))
             self.close_socket()
         except Exception as e:
             log.error("Unexpected error: %s", str(e))
-            return
+
+        self.packets_sent += 1
+        self.bytes_sent += len(packet)
```
zippolyte: Should be added only when success is True no ?

datadog/dogstatsd/base.py
```diff
@@ -312,40 +330,60 @@ def _report(self, metric, metric_type, value, tags, sample_rate):
...
+            self.bytes_dropped += len(packet)
+            self.packets_dropped += 1
 
     def _send_to_buffer(self, packet):
         self.buffer.append(packet)
         if len(self.buffer) >= self.max_buffer_size:
             self._flush_buffer()
 
     def _flush_buffer(self):
+        self.buffer.extend(self._flush_telemetry())
```
zippolyte: If not using a buffer, there will be no telemetry since you're only doing this here right ?

PR #506, opened 2020-01-03
[dogwrap] Add duration as metric

dabcoder:
This would address #4.


datadog/dogshell/wrap.py
```diff
@@ -387,6 +389,8 @@ def main():
         print(stdout.strip(), file=sys.stdout)
 
     if options.submit_mode == 'all' or returncode != 0:
+        if options.send_metric:
+            api.Metric.send(metric='dogwrap.duration', points=duration, type="gauge")
```
zippolyte: Let's add the tags, and probably add an `event_name` tag with the value being the value of `options.name`

Issue #311, opened 2018-11-06
Add Python type hints


pior:
### Description

This library doesn't provide type hints.

### Issue

When type checking a project, the `datadog` lib must be ignored (in the Mypy config) to avoid getting 
an "import error".

Ignoring a library reduce the usefulness of type checking.

Note: once type hints are added to the project, a PEP-561 marker should be added to signal to MyPy that the package provide type hints: see https://www.python.org/dev/peps/pep-0561/

zippolyte:
This is not something we are going to do, leaving it to the open source community. Our only requirement for such thing to be merged, is to stay compatible with python 2.7 (even though it is EOL).

jairideout:
Now that #565 is merged, the package supports type hints and type checking via Mypy (including a PEP 561 marker file). I added a few type hints as a starting point, and may submit follow-up PRs with additional type hints if I have time. If anyone is interested in helping with this, please jump in!

Note: `datadog.initialize` and `datadog.util.hostname.get_hostname` are examples of functions with multi-line and single-line type annotations, respectively.

vishalkuo:
I've opened https://github.com/DataDog/datadogpy/pull/579 - would appreciate a review whenever convenient

vishalkuo:
Bump here!

caire-bear:
> When type checking a project, the datadog lib must be ignored (in the Mypy config) to avoid getting an "import error".

@pior could you post an example of how to do this? i can't for the life of me figure out the right `pyproject.toml` config block.

I'll have imports throughout my project like this:

```
# worker.py
from datadog import statsd
```
which will throw:

```
src/worker.py:6: error: Module "datadog" does not explicitly export attribute "statsd"
[attr-defined]
    from datadog import statsd
    ^
Found 1 error in 1 file (checked 32 source files)
```

i want to avoid having to add `#type: ignore` everywhere inline (including `timed` decorators). i've tried several flavors of:

```
[[tool.mypy.overrides]]
module = "datadog"
ignore_errors = true
```

including `module="datadog.*"`, `disable_error_code = ["attr-defined"]` and can't figure it out. i'm a bit new to mypy.

PR #507, opened 2020-01-05
Fix / Drop Support for < Python 3.6

bryantbiggs:
- Remove support for Python < 3.6
- Strings in Python3 are unicode by default so `u'` is dropped
- Replace `.format(x)` with f-strings (proven to be more efficient and easier to read)
 - Bump package requirements for decorator and requests; these were mostly held back by the older runtime support. Note: there are known vulnerabilities in older versions of requests (>2.21) and decorators didn't start supporting 3.7 till version 4.4.0 so better to move those minimum versions up.

Datadog agent 7 dropped support for Python 2.7 - its a great time to clean up the gnarly compatibility code and move forward with improvements and functionality offered by Python 3.6+. 

3.5 was dropped as well due to most performance improvements coming in 3.6+ and 3.5 being used by only a small portion of users (only citable source I could find that seemed [credible from 2018](https://www.jetbrains.com/research/python-developers-survey-2018) - Python 3.6 & 3.7 covered 84% of the respondents in Python 3 versions used). If there is going to be a breaking change, it makes the most sense to do it in one go rather than across multiple changes. 

If this is merged, hopefully!, I would like to follow up with PRs to 
1. Codify coding standards (automatic import statement sorting, code formatting, etc.), this will help standardize the codebase stylistically, remove the need for manually fixing linting errors, and provide a consistent experience for contributors. However it will most likely cause a lot of whitespace changes which can be annoying so I will try to do this in one go and get it over with
2. Add type hinting - there is an open issue for this https://github.com/DataDog/datadogpy/issues/311 and it is a great feature to add to a codebase for others to understand grasp the source more easily. This too will be a rather large change but with very low risk to adverse changes
3. Update documentation - currently its a single pager; we can split this up to follow the packages in the codebase here and improve on the overall documentation experience for contributors especially with changes 1 and 2

If there are any inputs, suggestions to the above please let me know! Thanks


jd:
Thanks for your contribution!
This is an awesome idea but I am not sure we're ready to drop Python 2.7 yet 🤔

dabcoder:
Thanks for doing all this work @bryantbiggs. As @jd mentioned, we need to keep backward compatibility here, so we won't be able to merge this for now. I am afraid you'll have to create a fork if you want to use these updates going forward.

PR #508, opened 2020-01-05
Feature / Poetry

bryantbiggs:
- Replace the use of `setup.py`, `MANIFEST.in` with `poetry`

Poetry offers better utilities for managing dependencies (required vs dev), and publishing to pypi (`poetry publish --build`), plus the ability to run commands from within virtualenvs for an improved development experience

https://python-poetry.org/

This PR is only a draft at this time as it depends on some assumptions in #507


jd:
This is not gonna fly, see https://github.com/DataDog/datadogpy/pull/484

PR #510, opened 2020-01-08
Handle OSError on socket.close on Python 3.6+

charettes:
### What does this PR do?

This PR handles unexpected exceptions in `DogStatsd.close_socket()`.

Under certain obscure circumstances we get a crash on Python 3.6+ if the file descriptor vanishes before close is called.

See https://bugs.python.org/issue29343

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -314,7 +314,10 @@ class DogStatsd(object):
         Closes connected socket if connected.
         """
         if self.socket:
-            self.socket.close()
+            try:
+                self.socket.close()
+            except Exception as e:
+                log.error("Unexpected error: %s", str(e))
             self.socket = None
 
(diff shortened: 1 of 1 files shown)


datadog/dogstatsd/base.py
```diff
@@ -314,7 +314,10 @@ def close_socket(self):
         Closes connected socket if connected.
         """
         if self.socket:
-            self.socket.close()
+            try:
+                self.socket.close()
+            except Exception as e:
```
dabcoder: Since we want to catch that specific `OSError` exception, wouldn't it make more sense to do:

```suggestion
            except OSError as e:
```

charettes: Sure, I assumed we'd want to silence any exception but since this is only meant to raise `OSError` instances it's likely more correct.

charettes:
Thanks for the quick merge folks.

PR #515, opened 2020-01-14
Handle EAGAIN socket error when dropping packets

### LINKED ISSUE #514: DogStatsd closing socket unnecessarily when dropping packets
**Describe the bug**
When sending data to a Datadog agent through a UDS socket and the socket becomes full, DogStatsd drops packets that would block on the socket write and closes the socket. I believe the socket close is not necessary and results in more issues because the socket gets closed and re-opened constantly if under heavy load.

**To Reproduce**

1. Following instructions on Gist: https://gist.github.com/mrknmc/06cabb2d4187ea3180a0416fa39f7c82
2. See that exception raised is a `BlockingIOError` which is a `socket.error`, not a `socket.timeout` exception

**Expected behavior**
DogStatsd should catch `BlockingIOError` and log a warning without closing the socket.

**Screenshots**
N/A

**Environment and Versions (please complete the following information):**
- Version v0.33.0 of datadogpy.
- Sending data over unix domain sockets

**Additional context**

mrknmc:
### What does this PR do?

Fixes  #514 by catching `socket.error`and checking `errno == 11` (EAGAIN) and logs a warning without closing the socket.

### Description of the Change

As mentioned in the issue, `socket.send` raises a `BlockingIOError (socket.error on py2)` exception when the socket is "full" rather than a `socket.timeout` exception. Hence, we create a new branch in the try-raise where we catch `socket.error`, check the `errno` and log a warning without closing the socket. This works across both python 2 and 3.

I left the `socket.timeout` branch in place because I wasn't sure if there is another way to set up DogStatsd where `socket.timeout` is thrown when the socket is full. This way, the change shouldn't break compatibility for people who rely on that behaviour.

### Alternate Designs

I considered not logging a message when a `socket.error` with `errno == 11` occurs but thought that people might want to be informed about this happening.

### Possible Drawbacks

N/A

### Verification Process

I created a unit test and checked that it passes.

### Additional Notes

N/A

### Release Notes

N/A

- [x] PR title must be written as a CHANGELOG entry [(see why)](/CONTRIBUTING.md#pull-request-title)

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -399,15 +399,15 @@ class DogStatsd(object):
         except socket.timeout:
             # dogstatsd is overflowing, drop the packets (mimicks the UDP behaviour)
             pass
+        except (socket.herror, socket.gaierror) as se:
+            log.warning("Error submitting packet: {}, dropping the packet and closing the socket".format(se))
+            self.close_socket()
         except socket.error as se:
             if se.errno == errno.EAGAIN:
                 log.warning("Socket send would block: {}, dropping the packet".format(se))
             else:
                 log.warning("Error submitting packet: {}, dropping the packet and closing the socket".format(se))
                 self.close_socket()
-        except (socket.herror, socket.gaierror) as se:
-            log.warning("Error submitting packet: {}, dropping the packet and closing the socket".format(se))
-            self.close_socket()
         except Exception as e:
             log.error("Unexpected error: %s", str(e))
 
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -399,8 +399,11 @@ class DogStatsd(object):
             # dogstatsd is overflowing, drop the packets (mimicks the UDP behaviour)
             pass
         except (socket.error, socket.herror, socket.gaierror) as se:
-            log.warning("Error submitting packet: {}, dropping the packet and closing the socket".format(se))
-            self.close_socket()
+            if se.errno == 11:
+                log.warning("Socket send would block: {}, dropping the packet".format(se))
+            else:
+                log.warning("Error submitting packet: {}, dropping the packet and closing the socket".format(se))
+                self.close_socket()
         except Exception as e:
             log.error("Unexpected error: %s", str(e))
 
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -11,6 +11,7 @@ from random import random
 import logging
 import os
 import socket
+import errno
 from threading import Lock
 
 # datadog
@@ -398,12 +399,15 @@ class DogStatsd(object):
         except socket.timeout:
             # dogstatsd is overflowing, drop the packets (mimicks the UDP behaviour)
             pass
-        except (socket.error, socket.herror, socket.gaierror) as se:
-            if se.errno == 11:
+        except socket.error as se:
+            if se.errno == errno.EAGAIN:
                 log.warning("Socket send would block: {}, dropping the packet".format(se))
             else:
                 log.warning("Error submitting packet: {}, dropping the packet and closing the socket".format(se))
                 self.close_socket()
+        except (socket.herror, socket.gaierror) as se:
+            log.warning("Error submitting packet: {}, dropping the packet and closing the socket".format(se))
+            self.close_socket()
         except Exception as e:
             log.error("Unexpected error: %s", str(e))
 
--- a/tests/unit/dogstatsd/test_statsd.py
+++ b/tests/unit/dogstatsd/test_statsd.py
@@ -66,6 +66,14 @@ class BrokenSocket(FakeSocket):
         raise socket.error("Socket error")
 
 
+class OverflownSocket(FakeSocket):
+
+    def send(self, payload):
+        error = socker.error("Socker error")
+        error.errno = 11
+        raise error
+
+
 def telemetry_metrics(metrics=1, events=0, service_checks=0, bytes_sent=0, bytes_dropped=0, packets_sent=0, packets_dropped=0, transport="udp", tags="", namespace=""):
     version = get_version()
     if tags:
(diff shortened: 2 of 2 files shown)


datadog/dogstatsd/base.py
```diff
@@ -399,8 +399,11 @@ def _send_to_server(self, packet):
             # dogstatsd is overflowing, drop the packets (mimicks the UDP behaviour)
             pass
         except (socket.error, socket.herror, socket.gaierror) as se:
-            log.warning("Error submitting packet: {}, dropping the packet and closing the socket".format(se))
-            self.close_socket()
+            if se.errno == 11:
```
jd: This should be `errno.EAGAIN`.
The number is different depending on the platform.
I'd also prefer to see that under a specific `socket.error` `except` statement to be clearer. Catching everything like this is done is not a good thing, we should make clear things a bit.

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -66,6 +66,14 @@ def send(self, payload):
         raise socket.error("Socket error")
 
 
+class OverflownSocket(FakeSocket):
+
+    def send(self, payload):
+        error = socker.error("Socker error")
+        error.errno = 11
```
jd: ditto use `errno.EAGAIN`

datadog/dogstatsd/base.py
mrknmc: Good point on `errno.EAGAIN`.

What do you mean by a specific `socket.error` `except` statement? I can create a separate statement but if we want to keep the logic the same for non-EAGAIN errors, I'd have to add the `close_socket` logic there as well.

jd: I mean:

```python
except socket.error:
   if not EAGAIN:
     close()
except (socket.herror, socket.gaierror):
     close()
```

in summary :)

datadog/dogstatsd/base.py
```diff
@@ -398,7 +399,13 @@ def _send_to_server(self, packet):
         except socket.timeout:
             # dogstatsd is overflowing, drop the packets (mimicks the UDP behaviour)
             pass
-        except (socket.error, socket.herror, socket.gaierror) as se:
+        except socket.error as se:
```
jd: I think you need to switch the two `except` statements order here because:
```
>>> isinstance(socket.herror(), socket.error)
True
```
So
```python
>>> try:
...     raise socket.herror()
... except socket.error:
...     print("1")
... except socket.herror:
...     print("2")
...
1
```

(my bad I wrote the initial suggestion in the wrong order, sorry)

mrknmc: Re-ordered them in f4bec108947016266e98f901a0ab3bf5f407e132, thanks!

dabcoder: Correct me if I am wrong @mrknmc but since Python 3 raises `BlockingIOError` and not `socket.error`, we'd also need to take this into consideration to be fully py2/py3 compatible?

mrknmc: `BlockingIOError` raised in Python 3 is still a `socket.error` with `errno == errno.EAGAIN`:

```
>>> e = BlockingIOError(11, 'Resource temporarily unavailable')
>>> import errno
>>> e.errno == errno.EAGAIN
True
>>> import socket
>>> e.errno == errno.EAGAIN
True
```

jd: Are you sure about this?

https://docs.python.org/3/library/exceptions.html#BlockingIOError indicates it's an `OSError`, not a `socket.error`.

mrknmc: See following docs from 3.3 release:
https://docs.python.org/3/whatsnew/3.3.html#pep-3151-reworking-the-os-and-io-exception-hierarchy

Notably:
> You don’t have to worry anymore about choosing the appropriate exception type between OSError, IOError, EnvironmentError, WindowsError, mmap.error, socket.error or select.error. All these exception types are now only one: OSError. The other names are kept as aliases for compatibility reasons.

dabcoder: OK thanks! We're testing against Python 3.5 and upwards, so LGTM.

jd: Right, so you're actually pointing toward my direction @mrknmc. :)

We're catching `socket.error` here, not `OSError`. So `BlockingIOError` won't be caught by this code, they'll fall under the last `except Exception`.

mrknmc: I think you misunderstood @jd - `socket.error` is an alias for `OSError` on Python >= 3.3. If you try the gist I linked in #514 you can convince yourself of this :)

jd: Alright

jd: Alright! I misread the last sentence and I thought they were just subclasses, not _aliasses_.

So we should be good with that for Python 2 and Python 3!

dabcoder:
> So we should be good with that for Python 2 and Python 3!

@mrknmc Many thanks for following up on our comments!

mrknmc:
> > So we should be good with that for Python 2 and Python 3!
> 
> @mrknmc Many thanks for following up on our comments!

No problem, thanks for looking over it! :)

dabcoder:
@mrknmc Sorry almost ready to go, but the branch is now out of sync with the master branch. Can you update it and we should be good to go.

mrknmc:
> @mrknmc Sorry almost ready to go, but the branch is now out of sync with the master branch. Can you update it and we should be good to go.

Are  you ok with a rebase or do you want me to merge `master` into this branch?

dabcoder:
> Are you ok with a rebase or do you want me to merge `master` into this branch?

I would say the latter, merge `master` into your branch please, thanks!

mrknmc:
@dabcoder we should be good to go now, correct?

dabcoder:
Yes all good thanks

PR #533, opened 2020-02-11
Remove namespace telemetry

hush-hush:
### What does this PR do?

This PR remove the namespace for the telemetry metric to avoid counting them as custom metrics.

We also only flush the telemetry once every 10s (by default) to mirror the behavior from other Dogstatsd clients.

### DIFF
--- a/tests/unit/dogstatsd/test_statsd.py
+++ b/tests/unit/dogstatsd/test_statsd.py
@@ -70,23 +70,22 @@ class BrokenSocket(FakeSocket):
 class OverflownSocket(FakeSocket):
 
     def send(self, payload):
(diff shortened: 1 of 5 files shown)


tests/unit/dogstatsd/test_statsd.py
```diff
@@ -69,26 +70,47 @@ def send(self, payload):
 class OverflownSocket(FakeSocket):
 
     def send(self, payload):
-        error = socket.error("Socker error")
+        error = socker.error("Socker error")
```
zippolyte:
```suggestion
        error = socket.error("Socket error")
```

PR #542, opened 2020-03-04
Accept any Iterable as tags parameter

bharel:

### What does this PR do?

Allow any Iterable to be the tags parameter.

### Description of the Change

I made sure throughout the program, the tags parameter will be converted to a list.
This will allow you to enter any iterable, such as sets, and tuples to the tags parameter.

### Alternate Designs

Thought of converting it to a list in each and every API function. Chose to do it in a central place.

### Possible Drawbacks

None that I'm aware of.

### Verification Process

Datadog test suite.

### Additional Notes

Nope.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -433,12 +433,16 @@ class DogStatsd(object):
         self._send(string)
 
     def _add_constant_tags(self, tags):
+        result = []
+
+        if tags:
+            result.extend(tags)
+
         if self.constant_tags:
-            if tags:
-                return tags + self.constant_tags
-            else:
-                return self.constant_tags
-        return tags
+            result.extend(self.constant_tags)
+
+        return result
+
 
 
 statsd = DogStatsd()
--- a/datadog/threadstats/base.py
+++ b/datadog/threadstats/base.py
@@ -150,12 +150,11 @@ class ThreadStats(object):
         """
         if not self._disabled:
             # Append all client level tags to every event
-            event_tags = tags
+            event_tags = []
+            if tags:
+                event_tags.extend(tags)
             if self.constant_tags:
-                if tags:
-                    event_tags = tags + self.constant_tags
-                else:
-                    event_tags = self.constant_tags
+                event_tags.extend(self.constant_tags)
 
             self._event_aggregator.add_event(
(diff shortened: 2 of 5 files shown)


zippolyte:
Hmm that still wouldn't work for metrics, events, ... sent through the API directly like `api.Metric.send(...)`. Only lists are supported there still.
Do you think you could add that there too ? I'm don't want to merge this unless any iterables for tags are supported everywhere, so as not to cause confusion.

bharel:
@zippolyte Added. Haven't tested these much as datadogpy doesn't seem to have full API tests which include tags, and I haven't used the API before.

bharel:
Seems to be much harder to add to API. Working on it.

bharel:
@zippolyte The tests are non-deterministic for some reason. They sometimes pass and sometimes don't. Last 2 commits had some tests pass and some fail with no code difference.

zippolyte:
Hey @bharel,

The test issues are weird, I'm looking into them. Seems like an issue with how we recorded the API responses or something, will let you know

zippolyte:
Problem comes from the fact that in the test, when transforming into a list, the order of the tags array end up different than what it was when we recorded the API requests/responses, thus preventing a match.
Looking to see if we can ignore the array order in the request matchers we use.

bharel:
Order shouldn't matter should it? Datadog doesn't even mind the order afaik.

zippolyte:
Yeah the order doesn't matter, it's just that we recorded the requests and responses in order to mock the real thing, and if the order changes in the replays, it won't match and the tests will fail.

I added a commit to your branch to make the the matcher ignore the order for tags arrays found in JSON body payloads. The tests seem to be passing now.

zippolyte:
Okay, I think we're almost there. We just need to add something for the tags in metrics in the API client here: https://github.com/bharel/datadogpy/blob/50fc480f7e85ececefd99ff81d7f3825c0e8609b/datadog/api/metrics.py#L92-L103, as well as change lists to other iterables in the metrics and service checks tests and we should be good I believe

eliasdorneles:
Hello :wave: 
Any chance of reviving this?
Context: I'd love to be able to send `set` of tags instead of a list, would make it simpler to write tests.

PR #558, opened 2020-04-18
[telemetry] optionally decouple telemetry destination from other metrics

truthbk:
### What does this PR do?

In some esoteric setups or testing it might be useful to submit telemetry statistics to a different dogstatsd server. 

### Description of the Change

This PR allows to set a different destination for telemetry metrics flushed. 

### Alternate Designs

No alternate designs were considered.

### Possible Drawbacks

Added complexity to Dogstatsd constructor, could be confusing if not well documented.

### Verification Process

Unit tests updated. Feature is being used for dogstatsd performance/feature testing currently.

### Additional Notes

None.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -70,6 +70,10 @@ class DogStatsd(object):
         telemetry_min_flush_interval=(
             DEFAULT_TELEMETRY_MIN_FLUSH_INTERVAL
         ),                              # type: int
+
+        telemetry_host=None,            # type: Text
+        telemetry_port=DEFAULT_PORT,    # type: int
+        telemetry_socket_path=None,     # type: Text
         max_buffer_len=0                # type: int
     ):  # type: (...) -> None
         """
@@ -160,6 +164,10 @@ class DogStatsd(object):
                 log.warning("Port number provided in DD_DOGSTATSD_PORT env var is not an integer: \
                 %s, using %s as port number", dogstatsd_port, port)
 
+        # Assuming environment variables always override
+        telemetry_host = os.environ.get('DD_TELEMETRY_HOST', telemetry_host)
+        telemetry_port = os.environ.get('DD_TELEMETRY_PORT', telemetry_port)
+
         # Check enabled
         if os.environ.get('DD_DOGSTATSD_DISABLE') not in {'True', 'true', 'yes', '1'}:
             self._enabled = True
@@ -183,8 +191,21 @@ class DogStatsd(object):
             if not self._max_payload_size:
                 self._max_payload_size = UDP_OPTIMAL_PAYLOAD_LENGTH
 
+        self.telemetry_destination = True
+        if telemetry_socket_path is not None:
+            self.telemetry_socket_path = telemetry_socket_path
+            self.telemetry_host = None
+            self.telemetry_port = None
+        elif telemetry_host:
+            self.telemetry_socket_path = None
+            self.telemetry_host = self.resolve_host(telemetry_host, use_default_route)
+            self.telemetry_port = int(telemetry_port)
+        else:
+            self.telemetry_destination = False
+
         # Socket
         self.socket = None
+        self.telemetry_socket = None
         self._send = self._send_to_server
         self.encoding = 'utf-8'
 
@@ -242,7 +263,7 @@ class DogStatsd(object):
 
         return get_default_route()
 
-    def get_socket(self):
+    def get_socket(self, telemetry=False):
         """
         Return a connected socket.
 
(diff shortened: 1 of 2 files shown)


datadog/dogstatsd/base.py
```diff
@@ -50,7 +50,8 @@ class DogStatsd(object):
     def __init__(self, host=DEFAULT_HOST, port=DEFAULT_PORT, max_buffer_size=50, namespace=None,
                  constant_tags=None, use_ms=False, use_default_route=False,
                  socket_path=None, default_sample_rate=1, disable_telemetry=False,
-                 telemetry_min_flush_interval=DEFAULT_TELEMETRY_MIN_FLUSH_INTERVAL):
+                 telemetry_min_flush_interval=DEFAULT_TELEMETRY_MIN_FLUSH_INTERVAL,
+                 telemetry_host=None, telemetry_port=DEFAULT_PORT, telemetry_socket_path=None):
```
zippolyte, line 54: Can we document the added params in the docstring please ?

truthbk: Absolutely, my bad.

datadog/dogstatsd/base.py
```diff
@@ -112,6 +120,26 @@ def __init__(self, host=DEFAULT_HOST, port=DEFAULT_PORT, max_buffer_size=50, nam
 
         :param default_sample_rate: Sample rate to use by default for all metrics
         :type default_sample_rate: float
+
+        :param disable_telemetry: Should client telemetry be disabled
+        :type disable_telemetry: boolean
+
+        :param telemetry_min_flush_interval: Minimum flush interval for telemtry in seconds
```
zippolyte, line 127:
```suggestion
        :param telemetry_min_flush_interval: Minimum flush interval for telemetry in seconds
```

truthbk: Thank you! 🙇

datadog/dogstatsd/base.py
hush-hush: We should use None too for the port, for the case where users change the port and want the telemetry to follow.

datadog/dogstatsd/base.py
```diff
@@ -204,24 +250,39 @@ def resolve_host(host, use_default_route):
 
         return get_default_route()
 
-    def get_socket(self):
+    def get_socket(self, telemetry=False):
```
hush-hush, line 253: I think we should split this function to avoid code duplication. We could have `get_socket()` and `get_telemetry_socket()`. And then 2 helper fuctions `get_udp_socket(host, port)` and `get_uds_socket(socket_path)`

datadog/dogstatsd/base.py
```diff
@@ -140,8 +173,21 @@ def __init__(self, host=DEFAULT_HOST, port=DEFAULT_PORT, max_buffer_size=50, nam
             self.port = int(port)
             transport = "udp"
 
+        self.telemetry_destination = True
```
hush-hush, line 176: I'm not a fan of having 2 boolean to reflect the state of the telemetry (like `if self.telemetry and self.telemetry_destination`. I think we always compute the `telemetry_destination` and only check `self.telemetry` in the rest of the code.

By default `telemetry_destination` would be equal to host+port/socket used for metrics.

datadog/dogstatsd/base.py
truthbk: Coolio.

datadog/dogstatsd/base.py
truthbk: I can try to just use `telemetry_socket`.

datadog/dogstatsd/base.py
```diff
@@ -488,23 +488,24 @@ def _is_telemetry_flush_time(self):
...
+            flush_telemetry = self._is_telemetry_flush_time()
+            if flush_telemetry:
+                telemetry_packet = self._flush_telemetry()
+                if not self._dedicated_telemetry_destination():
+                    packet += "\n"+telemetry_packet
+                else:
+                    (self.telemetry_socket or self.get_socket(telemetry=True)).send(
+                        telemetry_packet.encode(self.encoding))
+
+                self._reset_telemetry()
```
hush-hush, line 501: In case we don't have a _dedicated_telemetry_destination we need to wait for the `send` line 504 before reseting the telemetry or we're going to lose data (if we have network error twice in a row).

datadog/dogstatsd/base.py
```diff
@@ -82,6 +83,14 @@ def __init__(self, host=DEFAULT_HOST, port=DEFAULT_PORT, max_buffer_size=50, nam
         If set, it is appended to the constant (global) tags of the statsd client.
         :type DD_VERSION: string
 
+        :envvar DD_TELEMETRY_HOST: the host for the dogstatsd server we wish to submit
+        telemetry stats to. If set, it overrides default value.
+        :type DD_TELEMETRY_HOST: string
+
+        :envvar DD_TELEMETRY_PORT: the port for the dogstatsd server we wish to submit
+        telemetry stats to. If set, it overrides default value.
+        :type DD_TELEMETRY_HOST: string
```
rodo, line 92: Should be `DD_TELEMETRY_PORT` in place of `DD_TELEMETRY_HOST` here, no ?

truthbk: You're 100% right, thanks for pointing it out. We're bringing this PR back to life, hopefully we'll merge it soon.

datadog/dogstatsd/base.py
```diff
@@ -512,13 +584,20 @@ def _send_to_server(self, packet):
                 self.bytes_dropped += len(telemetry)
                 self.packets_dropped += 1
 
-    def _xmit_packet(self, packet, telemetry):
+    def _xmit_packet(self, packet, account, is_telemetry):
```
hush-hush, line 587: I don't get what `account` is and what ` self._xmit_packet(packet, self._telemetry, False)` means then

truthbk: So, this what each means: 
- `account` is to tell `_xmit_packet()` if we want to do accounting of the packet getting sent.
- `telemetry` is to tell `_xmit_packet()` if we're sending an a telemetry packet or not. 

But I see what you're saying, we don't need two parameters since we'll never want to `account` for telemetry packets. OK, I'll fix this.

hush-hush (approved):

hush-hush (approved):

PR #559, opened 2020-04-29
Enable buffer for all submission

prognant:
### What does this PR do?
Enable buffering for all metric submission to prevent high PPS.

### Description of the Change
All metric submission are buffer. A background loop submit metric at periodic interval.

### Alternate Designs
A more elaborate design could disable the background loop if metric burst occur or if submission occurs at a reasonable frequency.

### Possible Drawbacks
It uses Thread.

### Additional Notes
N/A

### Release Notes
API is kept compatible.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -13,7 +13,7 @@ import os
 import socket
 import errno
 import time
-from threading import Lock
+from threading import Lock, Timer
 
 # datadog
 from datadog.dogstatsd.context import TimedContextManagerDecorator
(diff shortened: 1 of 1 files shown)


prognant:
Thanks for having a look.
The PR is really in an early stage of development, and I'm working on benchmarking the buffering logic. Regarding early benchmark, with the exact same submission code (I'm using this stub : https://github.com/DataDog/datadog-agent-benchmarks/blob/master/dogstatsd/python-client-fakeapp/main.py), buffering allows much more metric from to be submitted :
image

The whole idea of the client refactoring is:
- to lower the PPS when submitting lots of metrics (right now, outside of the batch submission pattern, sending one metric triggers one packet)
- to have best possible default settings (buffer depth, submission period, etc.)

Bouncing on the PR you mention, improve the thread-safetiness should also be part of this refactoring the redis query pipeline logic is indeed a nice replacement for the current batch submission mechanism (unsure if we can keep the existing batch API compatible with such a change). 

From an implementation standpoint this class should definitely stay simple, the buffering logic is likely to be extracted in a new file, whether we should keep the existing submission pattern (no buffering, one metric per packet) is yet to be planed.

jd:
I'm glad you're doing benchmarks, that's awesome.

Yeah to me the first big thing to do before any of that is to have a proper thread-safe buffer… and then you can build on top of it. :)

sgnn7:
Superseded by #670. Closing.

Issue #439, opened 2019-09-20
DogStatsd is not thread-safe when using buffer


jd:
If I'm reading the code correctly, the class `DogStatsd` is not thread-safe when using a buffer.

`open_buffer` does the following: `self.buffer = []`
Which means that if you have 2 threads opening a buffer at the same time, they will override each other.

The `open_buffer` class should return a different object that should be used as the buffer object.

jd:
Just to add an example of what I mean, `redis-py` has a nice design where they return a `Pipeline` object when you call `client.pipeline()`. That's something that'd work fine in that case.
https://redis-py.readthedocs.io/en/latest/#redis.Redis.pipeline

hush-hush:
Hi @jd,

Thanks for opening this issue. We're currently planning some ground work on the Python DogStatsD client. This is definitely on the list of things to improve/change. I can't give you any estimation about when this will be available but I'll be sure to update this issue !

komuw:
could use `collections.deque` which are documented as thread-safe and can act as a drop in replacement for list

https://docs.python.org/3.7/library/collections.html#collections.deque

PR #562, opened 2020-05-07
Size-based packetization for dogstatsd batched metrics 

prognant:
### What does this PR do?
It improves the packetization while sending batched metrics with packet size adjusted to prevent fragmentation while maximizing payload usage. Telemetry is sent in separate packets, once every 10 seconds won't impact overall performance. API is backward compatible, `max_buffer_size` has been deprecated but not removed, warning are logged if the parameter is used when building `Dogstatsd` objects.

### Description of the Change
It adds sending heuristic based on current buffer length. Some approximation regarding the actual bytes sent on wire include unicode character encoding (note that the telemetry is not accounting for that either).

### Alternate Designs
Better optimal packet size detection could have been considered but it seems overkill, moreover the settings can be overridden.

### Verification Process
Unit tests, integration tests, manual tests, packet capture.

### Additional Notes
Future rework of the client is planned and will include better multithreading awareness and optimal packet size management without the need to use batched metrics pattern.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -427,10 +427,7 @@ class DogStatsd(object):
         ))
 
     def _is_telemetry_flush_time(self):
-        if self._telemetry \
-           and self._last_flush_time + self._telemetry_flush_interval < time.time():
-            return True
-        return False
+        return self._last_flush_time + self._telemetry_flush_interval < time.time()
 
     def _send_to_server(self, packet):
         self._xmit_packet(packet, self._telemetry)
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -32,6 +32,10 @@ DEFAULT_PORT = 8125
 # Tag name of entity_id
 ENTITY_ID_TAG_NAME = "dd.internal.entity_id"
 
+# Default buffer settings
+UDP_OPTIMAL_PAYLOAD_LENGTH = 1432
+UDS_OPTIMAL_PAYLOAD_LENGTH = 8192
+
 # Mapping of each "DD_" prefixed environment variable to a specific tag name
 DD_ENV_TAGS_MAPPING = {
     'DD_ENTITY_ID': ENTITY_ID_TAG_NAME,
@@ -47,10 +51,11 @@ DEFAULT_TELEMETRY_MIN_FLUSH_INTERVAL = 10
 class DogStatsd(object):
     OK, WARNING, CRITICAL, UNKNOWN = (0, 1, 2, 3)
 
-    def __init__(self, host=DEFAULT_HOST, port=DEFAULT_PORT, max_buffer_size=50, namespace=None,
+    def __init__(self, host=DEFAULT_HOST, port=DEFAULT_PORT, max_buffer_size=None, namespace=None,
                  constant_tags=None, use_ms=False, use_default_route=False,
                  socket_path=None, default_sample_rate=1, disable_telemetry=False,
-                 telemetry_min_flush_interval=DEFAULT_TELEMETRY_MIN_FLUSH_INTERVAL):
+                 telemetry_min_flush_interval=DEFAULT_TELEMETRY_MIN_FLUSH_INTERVAL,
+                 max_buffer_len=0):
         """
         Initialize a DogStatsd object.
 
@@ -88,9 +93,8 @@ class DogStatsd(object):
         :param port: the port of the DogStatsd server.
         :type port: integer
 
-        :param max_buffer_size: Maximum number of metrics to buffer before sending to the server
-        if sending metrics in batch
-        :type max_buffer_size: integer
+        :max_buffer_size: Deprecated option, do not use it anymore.
+        :type max_buffer_type: None
 
         :param namespace: Namespace to prefix all metric names
         :type namespace: string
@@ -111,9 +115,18 @@ class DogStatsd(object):
 
         :param default_sample_rate: Sample rate to use by default for all metrics
         :type default_sample_rate: float
+
+        :param max_buffer_len: Maximum number of bytes to buffer before sending to the server
+        if sending metrics in batch. If not specified it will be adjusted to a optimal value
+        depending on the connection type.
+        :type max_buffer_len: integer
         """
 
         self.lock = Lock()
+
+        # Check for deprecated option
+        if max_buffer_size is not None:
+            log.warning("The parameter max_buffer_size is now deprecated and is not used anymore")
 
         # Check host and port env vars
         agent_host = os.environ.get('DD_AGENT_HOST')
@@ -129,20 +142,24 @@ class DogStatsd(object):
                 %s, using %s as port number", dogstatsd_port, port)
 
         # Connection
+        self._max_payload_size = max_buffer_len
         if socket_path is not None:
             self.socket_path = socket_path
             self.host = None
             self.port = None
             transport = "uds"
+            if not self._max_payload_size:
+                self._max_payload_size = UDP_OPTIMAL_PAYLOAD_LENGTH
         else:
             self.socket_path = None
             self.host = self.resolve_host(host, use_default_route)
             self.port = int(port)
             transport = "udp"
+            if not self._max_payload_size:
+                self._max_payload_size = UDS_OPTIMAL_PAYLOAD_LENGTH
 
         # Socket
         self.socket = None
-        self.max_buffer_size = max_buffer_size
         self._send = self._send_to_server
         self.encoding = 'utf-8'
 
@@ -168,7 +185,7 @@ class DogStatsd(object):
                 "client_version:{}".format(__version__),
                 "client_transport:{}".format(transport),
                 ]
-        self._reset_telementry()
+        self._reset_telemetry()
         self._telemetry_flush_interval = telemetry_min_flush_interval
         self._telemetry = not disable_telemetry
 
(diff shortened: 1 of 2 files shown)


datadog/dogstatsd/base.py
```diff
@@ -47,8 +53,8 @@
 class DogStatsd(object):
     OK, WARNING, CRITICAL, UNKNOWN = (0, 1, 2, 3)
 
-    def __init__(self, host=DEFAULT_HOST, port=DEFAULT_PORT, max_buffer_size=50, namespace=None,
-                 constant_tags=None, use_ms=False, use_default_route=False,
+    def __init__(self, host=DEFAULT_HOST, port=DEFAULT_PORT, max_buffer_size=MAX_BUFFER_ITEM, max_buffer_bytes_size=0,
+                 namespace=None, constant_tags=None, use_ms=False, use_default_route=False,
```
hush-hush, line 57: We should add new parameter at the end of the parameter list, so we don't break customer code who pass arguments without naming them.

ie:
```python
DogStatsd("localhost", 1234, 40, ["const_tag1", "const_tag2"])
```

datadog/dogstatsd/base.py
```diff
@@ -32,6 +33,11 @@
 # Tag name of entity_id
 ENTITY_ID_TAG_NAME = "dd.internal.entity_id"
 
+# Default buffer settings
+UDP_OPTIMAL_PAYLOAD_LENGTH = 1432
+UDS_OPTIMAL_PAYLOAD_LENGTH = 8192
+MAX_BUFFER_ITEM = sys.maxsize
```
hush-hush, line 39: `sys.maxsize` is way to big to be used as a default value. Also what the point of still using `MAX_BUFFER_ITEM` since we now computing the actual byte size ?

I think we should drop the feature of setting a manual limit on the number of metrics per packet (it's arbitrary and never work properly).

datadog/dogstatsd/base.py
```diff
@@ -420,14 +437,21 @@ def _is_telemetry_flush_time(self):
     def _send_to_server(self, packet):
         flush_telemetry = self._is_telemetry_flush_time()
         if flush_telemetry:
-            packet += "\n"+self._flush_telemetry()
-
+            telemetry = self._flush_telemetry()
+            same_packet = (len(telemetry) + len(packet) < self._max_payload_size)
+            if same_packet:
+                packet += "\n"+telemetry
```
hush-hush, line 443: Why not compute the length of the telemetry at startup and decrease `_max_payload_size` by it (we need the tags first and can set a high estimate for the metrics values). This would simplify the logic but more importantly avoid sending multiple packet when flushing the telemetry. It's what we do in the ruby client.

When looking at the size of the telemetry payload compare to a regular metric is will never fit in the remaining space of a buffer. This means that we will always have two `socket.send` in the next `try/except` that can be raise an exception. Also, currently the metrics will be wrong if we succeed in sending the metrics but fail for the telemetry packet (we would not increase `packets_sent` and `bytes_sent` line 453.

datadog/dogstatsd/base.py
```diff
@@ -451,12 +475,21 @@ def _send_to_server(self, packet):
...
             self._flush_buffer()
+        self.buffer.append(packet)
+        self._current_buffer_total_size += (len(packet) + 1)
+
+    def _should_flush(self, length_to_be_added):
+        if len(self.buffer) >= self._max_buffer_items:
+            return True
+        if self._current_buffer_total_size + length_to_be_added > self._max_payload_size:
+            return True
+        return False
```
hush-hush, line 488: This will break behavior if the user set a high number for `_max_buffer_items`.

datadog/dogstatsd/base.py
prognant: It makes sense.

datadog/dogstatsd/base.py
prognant: It was mostly for consistency purpose between clients (https://github.com/DataDog/datadog-go/blob/master/statsd/options.go#L17). I agree that settings an arbitrary limit is not a good idea and I don't mind dropping the feature.

datadog/dogstatsd/base.py
prognant: Agree, will do as detailed above.

datadog/dogstatsd/base.py
prognant: If we systematically lower the max buffer size by the telemetry size, we may increase the overall packet rate because packet without telemetry won't be filled (almost half the payload for UDP). I guess it depends on the submission pattern and as discussed before I don't have any number regarding packet size. I think anyhow that reworking the byte accounting will make the code much readable.

datadog/dogstatsd/base.py
```diff
@@ -445,18 +456,24 @@ def _send_to_server(self, packet):
...
             self.bytes_dropped += len(packet)
             self.packets_dropped += 1
 
     def _send_to_buffer(self, packet):
-        self.buffer.append(packet)
-        if len(self.buffer) >= self.max_buffer_size:
+        if self._should_flush(len(packet)):
             self._flush_buffer()
+        self.buffer.append(packet)
+        self._current_buffer_total_size += (len(packet) + 1)
```
hush-hush, line 467: Can we had a quick comment that the `+ 1` is for the line break between payload ?

datadog/dogstatsd/base.py
```diff
@@ -418,16 +428,17 @@ def _is_telemetry_flush_time(self):
...
 
     def _send_to_server(self, packet):
-        flush_telemetry = self._is_telemetry_flush_time()
-        if flush_telemetry:
-            packet += "\n"+self._flush_telemetry()
+        self._xmit_packet(packet)
+        if self._is_telemetry_flush_time():
+            telemetry = self._flush_telemetry()
+            self._reset_telemetry()
+            self._xmit_packet(telemetry)
```
hush-hush, line 435: We're losing telemetry info here in case we fail to send the telemetry over the network since we reset it in any case. `_xmit_packet` should return a value to indicate if it was successful or not.

datadog/dogstatsd/base.py
```diff
@@ -418,16 +428,17 @@ def _is_telemetry_flush_time(self):
...
 
+    def _xmit_packet(self, packet):
```
hush-hush, line 437: Why `_xmit_packet` ? Is this a typo for `_emit_packet` or does it means something ?

prognant: Most of the things I worked before used the wording `xmit` as a short version of transmit when sending a packet, [linux kernel does too](https://elixir.bootlin.com/linux/latest/source/include/linux/netdevice.h#L97), I have no strong opinion about `emit` vs. `xmit` vs. `transmit` in this context.

datadog/dogstatsd/base.py
prognant: Indeed I will adjust to keep telemetry if a failure occur while sending the packet.

datadog/dogstatsd/base.py
prognant: Will do.

prognant:
> Overall it looks good. Since the API for `DogStatsd` changed we should bump the major, but I'm not sure if we can do it since the whole repo contains multiple project. We should sync with ing-tool-libs on this.
> 
> If bumping the major is not possible, we could still receive `max_buffer_size` in the API but not use it and print a warning ?
I will sync & assess the situation with ing-tool-libs

datadog/dogstatsd/base.py
```diff
@@ -445,18 +463,26 @@ def _send_to_server(self, packet):
...
+        if self._should_flush(len(packet)):
             self._flush_buffer()
+        self.buffer.append(packet)
+        # Update the current buffer lenght, including line break to anticipate final packet size
+        self._current_buffer_total_size += (len(packet) + 1)
+
+    def _should_flush(self, length_to_be_added):
+        if self._current_buffer_total_size + length_to_be_added > self._max_payload_size:
+            return True
+        return False
```
jd, line 481:
```suggestion
        return self._current_buffer_total_size + length_to_be_added > self._max_payload_size
```

datadog/dogstatsd/base.py
```diff
@@ -225,7 +235,7 @@ def get_socket(self):
 
         return self.socket
 
-    def open_buffer(self, max_buffer_size=50):
+    def open_buffer(self):
```
jd, line 238: This breaks the current API FWIW.
I'm not against it, but if you're going to break the API, it'd be better to provide a new one that e.g. fixes #439

datadog/dogstatsd/base.py
```diff
@@ -445,18 +463,26 @@ def _send_to_server(self, packet):
...
             self.packets_dropped += 1
+        return False
 
     def _send_to_buffer(self, packet):
-        self.buffer.append(packet)
-        if len(self.buffer) >= self.max_buffer_size:
+        if self._should_flush(len(packet)):
             self._flush_buffer()
+        self.buffer.append(packet)
+        # Update the current buffer lenght, including line break to anticipate final packet size
```
jd, line 475:
```suggestion
        # Update the current buffer length, including line break to anticipate the final packet size
```

datadog/dogstatsd/base.py
prognant: Suggestion include in latest change.

datadog/dogstatsd/base.py
prognant: I had a quick discussion about that, we agreed not to break the API, pending upcoming rework (I keep in mind the redis pipeline idea thanks again BTW).

PR #565, opened 2020-05-23
Initial support for Python type hints and Mypy type checking

jairideout:
### What does this PR do?

This PR begins to address #311 by adding initial support for Python type hints and type checking via Mypy.

### Description of the Change

This PR updates the codebase to support Mypy type checking and includes type hints for a couple of functions. Before adding type hints to the entire codebase, I wanted to share these initial changes and see if this is a direction that maintainers want to move in. This PR can be merged on its own (with follow-up PRs to add additional type hints), or I'm also happy to continue adding type hints here before merge.

Type hints use the comment-based syntax described in [PEP 484](https://www.python.org/dev/peps/pep-0484/) to be compatible with both Python 2 and 3. Other changes to the codebase were necessary in order to fix a few Mypy errors (commented in the code and noted below).

The following changes were made:

- The functions `datadog.initialize` and `datadog.util.hostname.get_hostname` have been updated with type hints. These functions serve as examples of multi-line and single-line annotations, respectively.
- The package declares [PEP 561](https://www.python.org/dev/peps/pep-0561/) type hint compatibility via a `py.typed` marker file.
- There is a new tox `mypy` environment for running the Mypy type checker in Python 2 and Python 3 modes (`tox -e mypy`). This is hooked up to the CI job matrix.
- A few updates were necessary to fix Mypy errors related to version-specific code paths (most of the changes happened in `datadog.util.compat`). Mypy requires using `sys.version_info` directly instead of helper functions or variables to detect version-specific code paths. Also, mypy doesn't support try/except imports for version-specific code paths (more details [here](https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks)).
- I added a hack to `datadog.dogstatsd.context_async` to avoid Mypy syntax errors (see code comment for details and an alternative solution).

### Alternate Designs

Comment-based annotations are used over function annotations because the codebase must be Python 2/3 compatible. If support for Python 2 is dropped in the future, it may be desirable to use Python 3 function annotations instead. Also, the `sys.version_info` checks could be removed, and the hack in `datadog.dogstatsd.context_async` could be reverted.

If the hack in `datadog.dogstatsd.context_async` is undesirable, another option is to specify the source files to process in `mypy.ini` (using `glob` inclusion patterns), and omit `datadog.dogstatsd.context_async` from the list. I chose the current hack because manually listing the source files to include (even with recursive `glob`) could be error-prone (e.g. if a new module or subpackage is added, the developer would have to remember to update `mypy.ini` to include the file or folder).

### Possible Drawbacks

The changes introduce some complexity in a couple of places (mainly the hack in `datadog.dogstatsd.context_async` and requiring the direct usage of `sys.version_info` for version-specific code). Also, the type hints may be redundant in cases where a function's docstring already describes parameter types. Mypy may complain about invalid assignment statements in type-hinted functions, and typing annotations may need to be applied to fix these errors.

There shouldn't be noticeable performance issues to users.

### Verification Process

I tested the changes locally with tox and added a new `mypy` environment to ensure future changes are tested automatically in CI/CD. I used the new `mypy` environment to find and fix all errors, and verified that new Mypy errors are reported as tox failures.

To run the new `mypy` environment: `tox -e mypy`

### Additional Notes

N/A

### Release Notes

N/A

- [x] PR title must be written as a CHANGELOG entry [(see why)](/CONTRIBUTING.md#pull-request-title)

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -60,6 +60,10 @@ def initialize(api_key=None,                    # type: Optional[str]
     :param host_name: Set a specific hostname
     :type host_name: string
 
+    :param proxies: Proxy to use to connect to Datadog API;
+                    for example, 'proxies': {'http': "http:<user>:<pass>@<ip>:<port>/"}
+    :type proxies: dictionary mapping protocol to the URL of the proxy.
+
     :param api_host: Datadog API endpoint
     :type api_host: url
 
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -14,6 +14,7 @@ without hindering performance.
 import logging
 import os
 import os.path
+from typing import Any, List, Optional
 
 # datadog
 from datadog import api
@@ -32,10 +33,21 @@ logging.getLogger('datadog.dogstatsd').addHandler(NullHandler())
 logging.getLogger('datadog.threadstats').addHandler(NullHandler())
 
 
-def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
-               statsd_host=None, statsd_port=None, statsd_use_default_route=False,
-               statsd_socket_path=None, statsd_namespace=None, statsd_constant_tags=None,
-               return_raw_response=False, hostname_from_config=True, **kwargs):
+def initialize(api_key=None,                    # type: Optional[str]
+               app_key=None,                    # type: Optional[str]
+               host_name=None,                  # type: Optional[str]
+               api_host=None,                   # type: Optional[str]
+               statsd_host=None,                # type: Optional[str]
+               statsd_port=None,                # type: Optional[int]
+               statsd_use_default_route=False,  # type: bool
+               statsd_socket_path=None,         # type: Optional[str]
+               statsd_namespace=None,           # type: Optional[str]
+               statsd_constant_tags=None,       # type: Optional[List[str]]
+               return_raw_response=False,       # type: bool
+               hostname_from_config=True,       # type: bool
+               **kwargs                         # type: Any
+               ):
+    # type: (...) -> None
     """
     Initialize and configure Datadog.api and Datadog.statsd modules
 
@@ -47,10 +59,6 @@ def initialize(api_key=None, app_key=None, host_name=None, api_host=None,
 
     :param host_name: Set a specific hostname
     :type host_name: string
-
-    :param proxies: Proxy to use to connect to Datadog API;
-                    for example, 'proxies': {'http': "http:<user>:<pass>@<ip>:<port>/"}
-    :type proxies: dictionary mapping protocol to the URL of the proxy.
 
     :param api_host: Datadog API endpoint
     :type api_host: url
--- a/datadog/api/http_client.py
+++ b/datadog/api/http_client.py
@@ -18,7 +18,7 @@ try:
     import requests
     import requests.adapters
 except ImportError:
-    requests = None
+    requests = None  # type: ignore
 
 try:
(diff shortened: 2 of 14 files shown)


datadog/api/http_client.py
```diff
@@ -18,7 +18,7 @@
     import requests
     import requests.adapters
 except ImportError:
-    requests = None
+    requests = None  # type: ignore
```
jairideout, line 21: This is one of the recommended patterns for fixing Mypy errors when setting a missing module to `None`.

mypy.ini
```diff
@@ -0,0 +1,10 @@
+[mypy]
+
+[mypy-boto.*]
```
jairideout, line 3: These config sections tell Mypy to ignore third-party packages that are missing type hints.

tox.ini
```diff
@@ -40,6 +41,15 @@ deps =
     flake8==3.7.9
 commands = flake8 datadog
 
+[testenv:mypy]
+basepython = python3
```
jairideout, line 45: Mypy must be run in Python 3.5 or higher (but it can still type-check Python 2 code).

jairideout:
I'm looking into the integration test failures on py27 and pypy2. I must have skipped these when testing locally, sorry about that!

mypy.ini
jairideout: I added a comment to the file for future reference.

tox.ini
jairideout: I added a comment to the file for future reference.

jairideout:
Thanks @jirikuncar for triggering the integration pipeline and for reviewing! Looks like all CI tests are passing.

Would you like me to add more type hints in this PR before merge, or do you prefer follow-up PRs adding type hints to specific modules?

datadog/__init__.py
```diff
@@ -14,6 +14,7 @@
 import logging
 import os
 import os.path
+from typing import Any, List, Optional
```
justinsousa, line 17: just a heads up. this is only part of the std lib for python 3.5 and greater. so out of the  box `pip install datadog` for versions earlier than that will run successfully but unless typing is also installed the initial import will fail.

justinsousa: happy to submit an issue/fix if needed

setup.py
```diff
@@ -17,15 +17,19 @@ def get_readme_md_contents():
 
 install_reqs = ["decorator>=3.3.2", "requests>=2.6.0"]
 
-if [sys.version_info[0], sys.version_info[1]] < [2, 7]:
+if sys.version_info < (2, 7):
     install_reqs.append("argparse>=1.2")
 
+if sys.version_info < (3, 5):
```
justinsousa, line 23: oh hmm I see you added this here. doesn't seem to be ACK'd when installing via pip. I'll have to admit I forget how pypi/pip uses setup.py. I think this would work fine if you're calling `python setup.py (build|install)` but probably not via pip install directly

datadog/__init__.py
jairideout: In [setup.py](https://github.com/DataDog/datadogpy/pull/565/files#diff-2eeaed663bd0d25b7e608891384b7298), the `typing` package is included in `install_requires` if the user is running Python < 3.5.

setup.py
jairideout: Interesting, thanks for pointing out this issue! I created #571 so that package maintainers are aware of the issue.

PR #572, opened 2020-06-22
Add `typing` and `configparser` dependencies for python versions that need it

### LINKED ISSUE #571: datadog 0.37.0 wheel import error for Python < 3.5
**Describe the bug**
After installing the latest datadog 0.37.0 wheel release using a Python version < 3.5, a `typing` package import error is raised when importing `datadog`.

**To Reproduce**

1. Install a version of Python < 3.5 (for example, Python 3.4 or Python 2.7).
2. Install the wheel release of datadog 0.37.0 by running `pip install datadog`.
3. Try importing `datadog` in an interactive interpreter to see the import error:

```
>>> import datadog
  File "<stdin>", line 1, in <module>
  File "datadog/__init__.py", line 17, in <module>
    from typing import Any, List, Optional
ImportError: No module named typing
```

**Expected behavior**
The `typing` package should be included in the wheel's install requirements if Python < 3.5 is being used.

In `setup.py`, the `typing` package is added to `install_requires` if running a Python version < 3.5 (otherwise, the package is already included in the standard library for newer Python versions). This optional `typing` dependency is correctly installed when running `python setup.py install` or pip-installing from source. The wheel release that is installed via `pip install datadog` doesn't include the `typing` dependency (probably because the wheel release isn't Python version specific).

A couple of solutions come to mind:

- (Probably easiest) Always include the `typing` package in `install_requires`, regardless of Python version. According to the [`typing` package docs](https://pypi.org/project/typing/), if `typing` is already part of the standard library, installing the package has no effect.
- Create separate wheel releases for Python < 3.5 and Python >= 3.5.

**Environment and Versions:**
Version: datadog 0.37.0
Tested with Python 2.7.15 on macOS 10.15.5.

**Additional context**
The `typing` package import was added in PR #565. Thanks @justinsousa for pointing out this issue!

jairideout:
### What does this PR do?

Fixes import errors when using `datadog` wheel build with Python versions older than 3.5.

Fixes #571 

### Description of the Change

Adds explicit `typing` and `configparser` dependencies regardless of Python version, in order to allow a single `py2.py3` wheel build to work correctly.

### Alternate Designs

An alternative solution is to create multiple wheel builds for Python < 3.5 and Python >= 3.5. This solution seemed simpler and doesn't require the build/release process to change. 

### Possible Drawbacks

There are two more lightweight dependencies that aren't actually necessary for Python 3.5+ installations.

### Verification Process

I reproduced the import errors described in #571, and saw that they were fixed by installing `typing` and `configparser`. I added the two new dependencies to `setup.py` and ran the unit and integration tests in Python 2.7 and 3.8. Finally, I created a wheel build and tested it in Python 2.7 and 3.5.

### Additional Notes

N/A

### Release Notes

N/A

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,13 +18,16 @@ version = {}
 with open("datadog/version.py") as fp:
     exec(fp.read(), version)
 
-install_reqs = ["decorator>=3.3.2", "requests>=2.6.0"]
+# `typing` package is only required for Python versions older than 3.5, but we include it here so
+# that a single wheel build can be used for all supported Python versions. Installing `typing` on
+# Python 3.5+ has no effect.
+#
+# `configparser` package is only required for Python versions older than 3 (it is included here for
+# the same reason as the `typing` package).
+install_reqs = ["decorator>=3.3.2", "requests>=2.6.0", "typing", "configparser"]
 
 if sys.version_info < (2, 7):
     install_reqs.append("argparse>=1.2")
-
-if sys.version_info < (3, 5):
-    install_reqs.append("typing")
 
 setup(
     name="datadog",
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,12 @@ with open("datadog/version.py") as fp:
 #
 # `configparser` package is only required for Python versions older than 3 (it is included here for
 # the same reason as the `typing` package).
-install_reqs = ["decorator>=3.3.2", "requests>=2.6.0", "typing", "configparser"]
+install_reqs = [
+    "decorator>=3.3.2",
+    "requests>=2.6.0",
+    'typing;python_version<"3.5"',
+    'configparser;python_version<"3.0"',
+]
 
 if sys.version_info < (2, 7):
     install_reqs.append("argparse>=1.2")


setup.py
```diff
@@ -18,14 +18,17 @@ def get_readme_md_contents():
...
     exec(fp.read(), version)
 
-install_reqs = ["decorator>=3.3.2", "requests>=2.6.0"]
+# `typing` package is only required for Python versions older than 3.5, but we include it here so
+# that a single wheel build can be used for all supported Python versions. Installing `typing` on
+# Python 3.5+ has no effect.
+#
+# `configparser` package is only required for Python versions older than 3 (it is included here for
+# the same reason as the `typing` package).
+install_reqs = ["decorator>=3.3.2", "requests>=2.6.0", "typing", "configparser"]
```
jirikuncar, line 27: Would following code work (see https://www.python.org/dev/peps/pep-0508/)?
```suggestion
install_reqs = [
    "decorator>=3.3.2",
    "requests>=2.6.0",
    'typing;python_version<"3.5"',
    'configparser;python_version<"3.0"',
]
```

zippolyte: I tested, and it does. Commiting.

jairideout: Thanks @jirikuncar and @zippolyte, I wasn't aware of Python version specific dependencies in `install_requires`. This is a much better fix! 🎉

PR #583, opened 2020-07-01
Improve user-agent header to include telemetry information

bkabrda:
### What does this PR do?

Makes the user-agent header more useful.

### DIFF
--- a/tests/integration/api/test_api.py
+++ b/tests/integration/api/test_api.py
@@ -966,4 +966,3 @@ class TestDatadog:
     def test_user_agent(self, dog):
         _, resp = dog.api_client.APIClient.submit('GET', 'validate')
         assert re.match(r'^datadogpy\/[^\s]+ \(python [^\s]+; os [^\s]+; arch [^\s]+\)$', resp.request.headers['User-Agent'])
-        resp.request.headers['User-Agent']
--- a/tests/integration/api/test_api.py
+++ b/tests/integration/api/test_api.py
@@ -4,7 +4,9 @@
 # python
 import datetime
 import json
+import mock
 import os
+import re
 import time
 
 import requests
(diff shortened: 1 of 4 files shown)


tests/integration/api/test_api.py
```diff
@@ -959,3 +961,9 @@ def test_roles_crud(self, dog):
         # check if new role is deleted successfully
         res = dog.Roles.get(role_uuid)
         assert "errors" in res
+
+    @mock.patch('datadog.api._return_raw_response', True)
+    def test_user_agent(self, dog):
+        _, resp = dog.api_client.APIClient.submit('GET', 'validate')
+        assert re.match(r'^datadogpy\/[^\s]+ \(python [^\s]+; os [^\s]+; arch [^\s]+\)$', resp.request.headers['User-Agent'])
+        resp.request.headers['User-Agent']
```
nmuesch, line 969: Is this line needed?

tox.ini
```diff
@@ -11,9 +11,9 @@ envlist =
 passenv = DD_TEST_CLIENT*
 usedevelop = true
 deps =
-    !integration: mock
```
nmuesch, line 14: I think if we move the user agent test to tests/unit instead of tests/integration we wouldn't need to change deps.

tests/integration/api/test_api.py
bkabrda: Ah, forgotten debugging line, sorry. Will remove.

tox.ini
bkabrda: Yeah, but the whole unittest suite as it is set up now replaces the `RequestClient._session`, so we'd have to create a completely different setup just because of this one specific test. I don't mind doing that, but this is just simpler (plus it tests the real thing).

bkabrda:
Hmm, I have also realized that there's a theoretical possibility that someone uses the `URLFetchClient`, so I'll also make sure to correct the User-Agent header for it.

PR #589, opened 2020-07-15
Add environment variable to disable statsd metric collection

### LINKED ISSUE #576: Environment variable to disable statsd in development
**Is your feature request related to a problem? Please describe.**

Hi! We use the `statsd` part of this library in a django app to send metrics to Datadog when making API calls to a third party service (tracking things like status code and response times). In development, those API calls still happen, but we don't have the local datadog agent running. 

As a result we get errors such as `Error submitting packet: [Errno 61] Connection refused, dropping the packet and closing the socket"` which tend to clutter the command line output. 

**Describe the solution you'd like**
Would love if there was an environment variable we could set in development that would disable any `statsd` metric collection, perhaps `DATADOG_DISABLE_STATSD`?

**Describe alternatives you've considered**
Some alternative global "disable" option that isn't necessarily an environment variable would also work. For example with `ddtrace` we can disable it in development by putting the following in the django `settings.py` file:

```
DATADOG_TRACE = {
    "ENABLED": os.environ["DDTRACE_ENABLED"] == "True",
}
```

Another option we considered was to manually add some conditionals around each of our uses of `statsd` in the codebase to check if the app is running in development, but since we use it in quite a few places we felt this would be quite a messy approach. Similarly, trying to manually override the imported `statsd` module when the app is running in development felt like a not-so-great solution.

**Additional context**
It looks like there was a previous issue about this: https://github.com/DataDog/datadogpy/issues/296

Thanks a lot for making this! We're big fans of Datadog ✌️ 

cc @ThomasSydney

dgzlopes:
#### What does this PR do?

Closes #576

Adds an environment variable to disable `statsd` metric collection in certain cases (e.g development).

#### Description of the Change

Added an environment variable called `DD_DOGSTATSD_ENABLE` that is set as `True` by default.

#### Release Notes

Should be backward compatible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ tags = ['version:1', 'application:web']
 api.Event.create(title=title, text=text, tags=tags)
 ```
 
-In development, you can disable any `statsd` metric collection using `DD_DOGSTATSD_ENABLE=False`.
+In development, you can disable any `statsd` metric collection using `DD_DOGSTATSD_DISABLE=True` (or any not-empty value).
 
 ## DogStatsD
 
--- a/README.md
+++ b/README.md
@@ -69,6 +69,8 @@ tags = ['version:1', 'application:web']
 api.Event.create(title=title, text=text, tags=tags)
 ```
 
+In development, you can disable any `statsd` metric collection using `DD_DOGSTATSD_ENABLE=False`.
+
 ## DogStatsD
 
 In order to use DogStatsD metrics, the Agent must be [running and available](https://docs.datadoghq.com/developers/dogstatsd/?tab=python).
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -421,7 +421,7 @@ class DogStatsd(object):
         """
         if value is None:
             return
-        
+
         if self._enabled is not True:
             return
 
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -158,7 +158,7 @@ class DogStatsd(object):
                 %s, using %s as port number", dogstatsd_port, port)
 
         # Check enabled
-        if os.environ.get('DD_DOGSTATSD_ENABLE', 'True') == 'True':
+        if os.environ.get('DD_DOGSTATSD_ENABLE', 'True') in {'True', 'true', 'yes', '1'}:
             self._enabled = True
         else:
             self._enabled = False
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -158,7 +158,7 @@ class DogStatsd(object):
                 %s, using %s as port number", dogstatsd_port, port)
 
         # Check enabled
-        if os.environ.get('DD_DOGSTATSD_ENABLE', 'True') in {'True', 'true', 'yes', '1'}:
+        if os.environ.get('DD_DOGSTATSD_DISABLE') not in {'True', 'true', 'yes', '1'}:
             self._enabled = True
         else:
             self._enabled = False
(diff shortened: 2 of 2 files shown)


datadog/dogstatsd/base.py
```diff
@@ -154,6 +157,12 @@ def __init__(
                 log.warning("Port number provided in DD_DOGSTATSD_PORT env var is not an integer: \
                 %s, using %s as port number", dogstatsd_port, port)
 
+        # Check enabled
+        if os.environ.get('DD_DOGSTATSD_ENABLE', 'True') == 'True':
```
jirikuncar, line 161:
```suggestion
        if os.environ.get('DD_DOGSTATSD_ENABLE', 'True') in {'True', 'true', 'yes', '1'}:
```

jirikuncar: or maybe better inverse the logic since it's enabled by default.

README.md
```diff
@@ -69,6 +69,8 @@ tags = ['version:1', 'application:web']
 api.Event.create(title=title, text=text, tags=tags)
 ```
 
+In development, you can disable any `statsd` metric collection using `DD_DOGSTATSD_ENABLE=False`.
```
jirikuncar, line 72:
```suggestion
In development, you can disable any `statsd` metric collection using `DD_DOGSTATSD_DISABLE=True` (or any not-empty value).
```

dgzlopes:
Thanks for the quick review :)

dgzlopes:
Is there anything more I can do? I would love to start using this feature :)

datadog/dogstatsd/base.py
```diff
@@ -100,6 +100,9 @@ def __init__(
         If set, it is appended to the constant (global) tags of the statsd client.
         :type DD_VERSION: string
 
+        :envvar DD_DOGSTATSD_ENABLE: Enable any statsd metric collection (default True)
+        :type DD_DOGSTATSD_ENABLE: boolean
```
jirikuncar, line 104: If you have accepted the change in README then we need to change it everywhere.
```suggestion
        :envvar DD_DOGSTATSD_DISABLE: Disable any statsd metric collection (default False)
        :type DD_DOGSTATSD_DISABLE: boolean
```

datadog/dogstatsd/base.py
```diff
@@ -154,6 +157,12 @@ def __init__(
                 log.warning("Port number provided in DD_DOGSTATSD_PORT env var is not an integer: \
                 %s, using %s as port number", dogstatsd_port, port)
 
+        # Check enabled
+        if os.environ.get('DD_DOGSTATSD_ENABLE', 'True') in {'True', 'true', 'yes', '1'}:
```
jirikuncar, line 161:
```suggestion
        if os.environ.get('DD_DOGSTATSD_DISABLE') not in {'True', 'true', 'yes', '1'}:
```

dgzlopes:
I completely forgot about changing it on the rest of the code :sweat_smile:

dgzlopes:
Removed some whitespace too (was making the CI fail).

dgzlopes:
Hey @jirikuncar. The PR is missing something? Can I help in any way?

wilhelmklopp:
Exciting stuff!

Thanks for building and shipping this ❤️

dgzlopes:
Hey @therve, when it's the next release planned? I would love to start using this feature :)

peterdeme:
When is this getting released?

therve:
We'll schedule a release for next week, thanks for your patience.

soulshake:
> Added an environment variable called `DD_DOGSTATSD_ENABLE` that is set as `True` by default.

For anyone stumbling across this via Google, the envvar is `DD_DOGSTATSD_DISABLE`, not `DD_DOGSTATSD_ENABLE`. [ref](https://github.com/DataDog/datadogpy/blob/12c06dbc2d432eb15a873ae6e2434a0082c9945a/README.md?plain=1#L76)

PR #590, opened 2020-07-15
[dogstatsd] sock.setblocking(0) for UDP socket

prognant:
### What does this PR do?
Add sock.setblocking(0) for UDP socket when connecting to a dogstastd server.
It resolve a blocking situation that was happening with internal test, making half test job stuck and ultimately failing after hitting the 1h timeout.

### Description of the Change
cf. above.

### Alternate Designs
N/A

### Possible Drawbacks
None identified so far.

### Verification Process

Manual tests.
Test pipelines are all green with this change.

### Additional Notes
IPv6 seems unsupported for dogstatsd communication, should be fixed.

### Release Notes
cf. PR title

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -241,8 +241,8 @@ class DogStatsd(object):
             if not self.socket:
                 if self.socket_path is not None:
                     sock = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
-                    sock.connect(self.socket_path)
                     sock.setblocking(0)
+                    sock.connect(self.socket_path)
                     self.socket = sock
                 else:
                     sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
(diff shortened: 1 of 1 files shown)


datadog/dogstatsd/base.py
```diff
@@ -246,6 +246,7 @@ def get_socket(self):
                     self.socket = sock
                 else:
                     sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+                    sock.setblocking(0)
```
jd, line 249: The order is different compared to the block above, where the blocking is set after `connect`. 
I think your order is fine here, I'm more worried about the one above. But they should probably not be different?

prognant: Indeed, I harmonised both case, I did some test to ensure it does not break anything.

prognant:
> I _think_ the effect of making the UDP socket non-blocking is that on a `send`, if the socket send buffer is full, an exception will be raised instead of waiting (blocking) for the send buffer to not be full.
> 
> Are we catching these errors correctly in UDP mode if they to be raised by a `send`? If so, I think this change is fine 👍

In theory I agree on the effect of the call, as far as checked cpython it should be consistent across platforms including windows. Regarding the error exception with errno set to `EAGAIN` are already properly caught, so we should be covered.

olivielpeau:
ok great! Thanks for confirming

PR #602, opened 2020-09-21
[dogstatsd] Fix race condition in _flush_buffer

### LINKED ISSUE #601: Possible race condition in _flush_buffer
The internal buffer is not cleared until after a yield to io, it is possible for another thread to attempt to flush the same buffer.

```py
    def _flush_buffer(self):
        self._send_to_server("\n".join(self.buffer)) # <-- do io, might yield greenlet/thread
        self._current_buffer_total_size = 0
        self.buffer = []
```

**Expected behavior**
no race conditions plz!

**Environment and Versions (please complete the following information):**
0.39.0

NorthIsUp:
### What does this PR do?

fixes bug #439

### Description of the Change

copy the global buffer to a local buffer so we can't accidentally flush it twice.

### Alternate Designs

`self.buffer` could be a queue, but that has other issues, this approach should be good enough without the complexity

### Verification Process

existing tests should cover this

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -552,7 +552,7 @@ class DogStatsd(object):
         return False
 
     def _flush_buffer(self):
-        with Lock():
+        with self.lock:
             to_send, self.buffer = self.buffer, []
             self._current_buffer_total_size = 0
 
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -552,8 +552,10 @@ class DogStatsd(object):
         return False
 
     def _flush_buffer(self):
-        to_send, self.buffer = self.buffer, []
-        self._current_buffer_total_size = 0
+        with Lock():
+            to_send, self.buffer = self.buffer, []
+            self._current_buffer_total_size = 0
+
         self._send_to_server("\n".join(to_send))
 
     def _escape_event_content(self, string):
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -552,9 +552,9 @@ class DogStatsd(object):
         return False
 
     def _flush_buffer(self):
-        self._send_to_server("\n".join(self.buffer))
+        to_send, self.buffer = self.buffer, []
         self._current_buffer_total_size = 0
-        self.buffer = []
+        self._send_to_server("\n".join(to_send))
 
     def _escape_event_content(self, string):
         return string.replace('\n', '\\n')


datadog/dogstatsd/base.py
```diff
@@ -552,9 +552,9 @@ def _should_flush(self, length_to_be_added):
         return False
 
     def _flush_buffer(self):
-        self._send_to_server("\n".join(self.buffer))
+        to_send, self.buffer = self.buffer, []
         self._current_buffer_total_size = 0
-        self.buffer = []
+        self._send_to_server("\n".join(to_send))
```
jd, line 557: There's still a chance another thread will write to the buffer reference, depending on where it was interrupted.

This improves the timing but does not fix the root cause.

datadog/dogstatsd/base.py
```diff
@@ -552,9 +552,11 @@ def _should_flush(self, length_to_be_added):
         return False
 
     def _flush_buffer(self):
-        self._send_to_server("\n".join(self.buffer))
-        self._current_buffer_total_size = 0
-        self.buffer = []
+        with Lock():
```
jd, line 555: Just in case, that definitely will not work. I'll leave you as an exercise to guess why? 😉

jd (changes requested):

datadog/dogstatsd/base.py
NorthIsUp: what, every thread deserves its own lock! ;)

NorthIsUp:
@jd does this look better?

datadog/dogstatsd/base.py
```diff
@@ -552,9 +552,11 @@ def _should_flush(self, length_to_be_added):
         return False
 
     def _flush_buffer(self):
-        self._send_to_server("\n".join(self.buffer))
-        self._current_buffer_total_size = 0
-        self.buffer = []
+        with self.lock:
```
jd, line 555: This lock is used for socket handling, it's not a great idea to re-use it for another purpose.

This also only solves a potential race around `_flush_buffer`, while the original bug report mentioned `open_buffer`  as it's the first one to be the issue.

jd (changes requested):
I don't think you're going to solve the problem correctly that way, it's a little more complicated that just adding `with lock` TBH and needs redesign of the API IMHO.

jd:
@NorthIsUp are you hitting an issue in particular for working on this?

sgnn7:
Closing since this was fixed via [this](https://github.com/DataDog/datadogpy/pull/642) PR now

PR #615, opened 2020-10-29
[dogstatsd] use monotonic clock source when available for timers

jd:
This makes sure the computed time is correct by using a monotonic clock source.

### DIFF
--- a/datadog/threadstats/base.py
+++ b/datadog/threadstats/base.py
@@ -14,6 +14,10 @@ import os
 from contextlib import contextmanager
 from functools import wraps
 from time import time
+try:
+    from time import monotonic  # type: ignore[attr-defined]
+except ImportError:
+    from time import time as monotonic
 
 # datadog
 from datadog.api.exceptions import ApiNotInitialized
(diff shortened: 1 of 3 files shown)


Kyle-Verhoog (approved):

jd:
@DataDog/integrations-tools-and-libraries can I get a review on this?

datadog/threadstats/base.py
```diff
@@ -13,7 +13,10 @@
 # stdlib
 from contextlib import contextmanager
 from functools import wraps
-from time import time
+try:
+    from time import monotonic as time  # type: ignore[attr-defined]
```
zippolyte, line 17: That's kinda problematic here, since the time is used in several places to compute the actual timestamp of a metric, and not in a difference of `end - start` 

You'd need to separate the usages here

zippolyte (changes requested):

datadog/threadstats/base.py
jd: I totally missed that. Will update.

jd:
Not sure the failure are related to the PR?

zippolyte:
The python3 no indeed, as they fail on master too for some reason, but the py2 failure seems related as master works ok

datadog/threadstats/base.py
```diff
@@ -272,11 +276,11 @@ def get_user(user_id):
...
                     stats.histogram('user.query.time', time.time() - start)
         """
-        start = time()
+        start = monotonic()
         try:
             yield
         finally:
-            end = time()
+            end = monotonic()
             self.timing(metric_name, end - start, end, tags=tags,
```
zippolyte, line 284: ooh the end time is used here too as timestamp

zippolyte (approved):

Issue #611, opened 2020-10-19
Add Log-queries to api 


rynmccrmck:
**Note:** 
Currently trying to extract large number of logs that match a given query.  We need to iterate over the 1000 limit imposed and it seems as though this is the only endpoint available.

I can take a stab at implementing unless there are concerns.

**Is your feature request related to a problem? Please describe.**
Need to extract logs over 1k limit and paginate over nextLogId.  No ability to do this via events.

**Describe the solution you'd like**
Implement LogQuery class to return list of logs: https://docs.datadoghq.com/api/v1/logs/#get-a-list-of-logs

**Describe alternatives you've considered**
Currently using requests to handle this.

therve:
Hello,

Support for new APIs is implemented in the https://github.com/DataDog/datadog-api-client-python repository. I believe it already contains what you need. We haven't made the first release, but it should happen soon. Could you check if that works for you? Thanks.

rynmccrmck:
Thanks @therve! Will take a look

rynmccrmck:
Any idea on estimated time to release?

rynmccrmck:
Closing this as it is implemented in https://github.com/DataDog/datadog-api-client-python

therve:
Thanks for checking!

> Any idea on estimated time to release?

I don't want to set expectations, but I hope we'll cut a first beta in the next few weeks if nothing major is found. If this works for you this is a good data point :).

PR #622, opened 2021-01-16
Add logs listing feature

userlocalhost:

### What does this PR do?
This PR implement a feature which is mentioned at #611.

### Description of the Change

This PR implemented API wrapper of 'logs-queries/list' (c.f.
https://docs.datadoghq.com/api/v1/logs/) that returns list of logs which
are matched with specified log query.

### Alternate Designs

This PR follows implementation of other API wrappers.

### Possible Drawbacks

Yes. This PR has backward compatibility because this just add a new API wrapper.

### Verification Process

This is an example code to get logs by using added implementation. If you run this code, please replace APP_KEY, API_KEY and SEARCH_QUERY for your environment.
```
from datadog import api, initialize

options = {
    'api_key': 'APP_KEY',
    'app_key': 'API_KEY',
}

initialize(**options)

resp = api.Logs.list({
    'time': {
        "from": "2021-01-14T00:00:00Z",
        "to": "2021-01-15T00:00:00Z"
    },
    'query': 'SEARCH_QUERY',
    'limit': 10,
})
print(resp)
```

### DIFF
--- a/datadog/api/__init__.py
+++ b/datadog/api/__init__.py
@@ -49,3 +49,4 @@ from datadog.api.roles import Roles
 from datadog.api.permissions import Permissions
 from datadog.api.service_level_objectives import ServiceLevelObjective
 from datadog.api.synthetics import Synthetics
+from datadog.api.logs import Logs
--- a/datadog/api/logs.py
+++ b/datadog/api/logs.py
@@ -1,0 +1,21 @@
+# Unless explicitly stated otherwise all files in this repository are licensed under the BSD-3-Clause License.
+# This product includes software developed at Datadog (https://www.datadoghq.com/).
+# Copyright 2015-Present Datadog, Inc
+from datadog.api.resources import CreateableAPIResource
+from datadog.api.api_client import APIClient
+
+
+class Logs(CreateableAPIResource):
+    """
+    A wrapper around Log HTTP API.
+    """
+    _resource_name = 'logs-queries'
+
+    @classmethod
+    def list(cls, data):
+        path = '{resource_name}/list'.format(
+            resource_name=cls._resource_name,
+        )
+        api_version = getattr(cls, '_api_version', None)
+
+        return APIClient.submit("POST", path, api_version, data)
--- a/tests/unit/api/test_api.py
+++ b/tests/unit/api/test_api.py
@@ -17,6 +17,7 @@ from datadog import initialize, api, util
 from datadog.api import (
     Distribution,
     Event,
+    Logs,
     Metric,
     ServiceCheck,
     User
@@ -602,6 +603,16 @@ class TestEventResource(DatadogAPIWithInitialization):
                 title="test no hostname", text="test no hostname", attach_host_name=False, alert_type="wrong_type"
             )
         assert "Parameter alert_type must be either error, warning, info or success" in str(excinfo.value)
+
+
+class TestLogsResource(DatadogAPIWithInitialization):
+    def test_list_logs(self):
+        Logs.list(data={"time": {"from": "2021-01-01T11:00:00Z", "to": "2021-01-02T11:00:00Z"}})
+        self.request_called_with(
+            "POST",
+            "https://example.com/api/v1/logs-queries/list",
+            data={"time": {"from": "2021-01-01T11:00:00Z", "to": "2021-01-02T11:00:00Z"}}
+        )
 
 
 class TestMetricResource(DatadogAPIWithInitialization):


therve:
Hi,

This is already handled by our generated client: https://github.com/DataDog/datadog-api-client-python/blob/master/docs/v1/LogsApi.md#list_logs

Would that work for your purpose?

userlocalhost:
@therve
Thank you for your prompt reply. Yes, I confirmed I could do that using the datadog-api-client-python.

But I wonder why Datadog team develops two different Python libraries. Will this (datadogpy) not be extended any more and be obsoleted?

therve:
> BTW, I wonder why Datadog team develops two different Python libraries. 

It's tough to keep up manually with all the APIs we produce, so we started a new generated library.

> Will this (datadogpy) not be extended any more and be obsoleted?

datadogpy will continue to live, as it contains other capabilities (like metrics and dogstatsd) that our generated library doesn't do. We're still trying to define what our strategy is going to be, but we'll most likely won't add new endpoints to datadogpy like this one. One possibility would be to depend on datadog-api-client-python and expose the functionality that way, would that work for you?

userlocalhost:
Thank you for answering my question. May I ask some more questions?

I wonder what we should do in the use-case that datadog user wants to use both logs and metrics API simultaneously(e.g. posting metric that is made from collecting logs). Should we have to use both datadogpy and datadog-api-client-python libraries in our software?

And as you said, the datadog-api-client-python seems not to have implementation to handle some metrics APIs (e.g. [[POST] /api/v1/series](https://docs.datadoghq.com/api/v1/metrics/)).
Will you add features that handle those endpoints to datadog-api-client-python?

therve:
> Thank you for answering my question. May I ask some more questions?
> 
> I wonder what we should do in the use-case that datadog user wants to use both logs and metrics API simultaneously(e.g. posting metric that is made from collected logs). Should we have to use both datadogpy and datadog-api-client-python libraries in our software?

Today, yes. As mentioned previously, one possible implementation is to import datadog-api-client-python in datadogpy, so that you only depend and use datadogpy, but it uses datadog-api-client-python under the hood. Today you have to use both.

> And as you said, the datadog-api-client-python seems not to have implementation to handle some metrics APIs (e.g. [[POST] /api/v1/series](https://docs.datadoghq.com/api/v1/metrics/)).
> Will you add features that handle those endpoints to datadog-api-client-python?

No we don't plan to add metric submission to generated clients.

userlocalhost:
> Today, yes. As mentioned previously, one possible implementation is to import datadog-api-client-python in datadogpy, so that you only depend and use datadogpy, but it uses datadog-api-client-python under the hood. Today you have to use both.

Thank you for your reply. Sorry, I'm not sure the meaning of "Today". If you don't plan to support metrics API on the datadog-api-client-python. User permanently has to use both libraries for that use-case, doesn't it?

Or, is there possibility to consider to merge a PR when I make another PR that implement API handler of metrics on the datadog-api-client-python like this?

therve:
> Thank you for your reply. Sorry, I'm not sure the meaning of "Today". If you don't plan to support metrics API on the datadog-api-client-python. User permanently has to use both libraries for that use-case, doesn't it?

User has to install both, but we can make it so that you only explicitly depend on and import datadogpy.

> Or, is there possibility to consider to merge a PR when I make another PR that implement API handler of metrics on the datadog-api-client-python like this?

No we won't accept such a contribution. We might at some point add it, but it would be generated anyway.

userlocalhost:
Sorry, I had insufficient understanding about what the difference between datadogpy and datadog-api-client-python.

To summarize, the datadogpy is a traditional datadog python library and it has been extended in manual. On the other hand, the datadog-api-client-python is another python API client for the Datadog API. Thats code was is generated using some other tools. And it is expected to collect all public Datadog API endpoints, but it has not finished it yet, right?

I think the datadog-api-client-python is elaborate and cool solution. And I also understood the reason why it won't accept this type of in manual extended because that code would be generated automatically.

> It's tough to keep up manually with all the APIs we produce

> we'll most likely won't add new endpoints to datadogpy

I know that feeling, but I suspect whether this bold decision is good for Datadog users. I suggest to keep maintain (including extending features) datadogpy until the datadog-api-client-python has been completed. Following is reasons why I think so.

### (reason1) It causes an inconvenience for users
If the datadog-api-client-python has already cover all the Datadog API, I'll be all for that decision. But it hasn't yet. If you make a decision that datadogpy refuse any kind extension any more, user might feel that "Both official python libraries are useless. One has abandoned extending of features, some API endpoints have not been supported. And another is in the middle of development, some API endpoints also have not been supported yet".
(I don't mean to offend you, but this is an honest opinion from an user perspective)

User carefully has to distinguish which library should we use depending on what API endpoints user want to access. This is inconvenient for users I think.

### (reason2) Probably no one knows it
I've read the whole datadogpy README and [an official documentation](https://docs.datadoghq.com/ja/developers/libraries/) that describes client libraries. There is no explanation that the datadogpy intends not to develop extensions and accept any PR that extends feature any more. This is against for users expectation I think.

---

I understand how you feel that you want to dedicate to developing the datadog-api-client-python. But please at least accept contributions for the datadogpy from user if it's worth for users until development of the datadog-api-client-python is finished.

I hope Datadog team makes a decision that makes Datadog users happy.

Sincerely,

therve:
I didn't say we won't accept your patch. I said we already have a library which does what you need. If the plan isn't clear it's because it's not finalized. We'll clarify the place of both libraries when it is. Thanks.

therve (approved):

PR #634, opened 2021-02-23
Fix blocking connections in dogstatsd

therve:
This adds back non blocking to UDP and sets non blocking before connect
like it was before.


zippolyte (dismissed):

zippolyte (approved):

olivielpeau:
Thanks for this!

Trying to understand the history of this: it looks like setting non-blocking UDP and setting non blocking before connecting was fixed in https://github.com/DataDog/datadogpy/pull/590 (`0.39.0`) but then mistakenly reverted with https://github.com/DataDog/datadogpy/pull/558 (`0.40.0`).

PR #638, opened 2021-03-02
[statsd] Clean up statsd tests

sgnn7:
### What does this PR do?

In preparation for additional work for statsd client, some prep needs to be done to
the codebase to reduce chances of functionality breaking so tests of statsd have
been cleaned up in this PR

### Description of the Change

- No functional change on the client
- Unit tests were cleaned up to raise conformance score to pylint from ~6 to ~9
- Unit tests were reworked to use standard assertions instead of keyword assertions
- Some minor cleanup of the code was done (one test was fixed, some comments were outdated)

### Alternate Designs

N/A

### Possible Drawbacks

N/A

### Verification Process

Running the tests should give the exact same output as before

### Additional Notes

AC-34

### Release Notes

N/A

### DIFF
--- a/tests/unit/dogstatsd/test_statsd.py
+++ b/tests/unit/dogstatsd/test_statsd.py
@@ -6,25 +6,21 @@
 """
 Tests for dogstatsd.py
 """
-# stdlib
+# Standard libs
 from collections import deque
 import os
 import socket
 import errno
 
-import mock
 import time
 import unittest
 
-# 3p
-from mock import (
-    call,
-    mock_open,
-    patch,
-)
+# Third-party libs
+import mock
+from mock import call, mock_open, patch
 import pytest
 
-# datadog
+# Datadog libs
 from datadog import initialize, statsd
 from datadog import __version__ as version
 from datadog.dogstatsd.base import DogStatsd, UDP_OPTIMAL_PAYLOAD_LENGTH
@@ -32,10 +28,6 @@ from datadog.dogstatsd.context import TimedContextManagerDecorator
 from datadog.util.compat import is_higher_py35, is_p3k
 from tests.util.contextmanagers import preserve_environment_variable, EnvVars
 from tests.unit.dogstatsd.fixtures import load_fixtures
-
-
-def assert_equal(a, b):
-    assert a == b
 
 
 class FakeSocket(object):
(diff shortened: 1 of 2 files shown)


tests/unit/dogstatsd/test_statsd.py
```diff
@@ -95,44 +89,44 @@ def telemetry_metrics(metrics=1, events=0, service_checks=0, bytes_sent=0, bytes
...
+    def assert_equal_telemetry(self, expected_payload, actual_payload, telemetry=None):
+        if telemetry is None:
+            telemetry = telemetry_metrics(bytes_sent=len(expected_payload))
+
+        if expected_payload:
+            expected_payload = "\n".join([expected_payload, telemetry])
+        else:
+            expected_payload = telemetry
+
+        return self.assertEqual(expected_payload, actual_payload)
```
zippolyte, line 122: since we're using pytest, I'd use plain `assert` everywhere instead of the `unittest` methods

sgnn7: Hey @zippolyte - thanks for taking a look!

While I know we use `pytest` and its preferred assertions are the `assert` keyword, I opted to change it since it currently makes the test code tightly coupled with this specific test framework. By using the `unittest` assertions, the code can be run with any test framework we may want in the future (`nose2`, `unittest`, etc) without requiring any code changes here and it's still fine with `pytest`. I believe that unlike the `pytest`'s `assert`s, `unittest`s assert can be used to emit multiple errors per test.

Also architecturally, `pytest` does [a lot of magic](http://pybites.blogspot.com/2011/07/behind-scenes-of-pytests-new-assertion.html) to ensure that `assert`s are rewritten in test bytecode which makes for a much more fragile setup since that keyword behaves differently depending on what is used as the test runner, which interpreter is used, and the behavior is [only changed in test code](https://docs.pytest.org/en/stable/assert.html#assertion-introspection-details). 

Let me know your thoughts on this - I know consistency is important but it seemed like we can have our future-proofing cake and eat it with portable assertions.

zippolyte: Well my counter point would be that the more portable option is actually the `assert` keyword since it's plain python so it will work no matter what. Also changing test framework is not something we do everyday, and does not usually come for free, even if you use portable assertions.
It's true pytest does a lot of magic, but it also has a pretty large compatibility matrix, and it works with all those interpreters/versions we target.
As for stacking assertion error, since we can do asserts on entire objects, and have a good diff provided by pytest, it's usually not a limitation.

That being said, I don't have a strong opinion on this, so I'm fine with keeping it as is 🙂

hush-hush (dismissed):

hush-hush (approved):

PR #642, opened 2021-03-05
[statsd] Fix buffer operation thread-safety

sgnn7:
### What does this PR do?

Old code did not have locking around its `open_buffer()` and
`close_buffer()` operations which made those two methods not
thread-safe. This change fixes this and aligns this module with the
documentation.

Fixes https://github.com/DataDog/datadogpy/issues/439
Fixes https://github.com/DataDog/datadogpy/issues/601

### Description of the Change

We now have an `RLock` around buffer operations that prevents threads
from changing the buffer owned by another thread.

Overall performance decrease when sending 2+ metrics is within the `+/- 5%` margin of error:
```sh
$ /time_buffer_ops.py 
Using Python3
baseline: 6.19679 (baseline)
mod     : 6.02907 (97.29%, 2.78% perf increase)
baseline: 6.20096 (baseline)
mod     : 6.04097 (97.42%, 2.65% perf increase)
baseline: 6.05407 (baseline)
mod     : 6.12549 (101.18%, -1.17% perf increase)
baseline: 6.16327 (baseline)
mod     : 6.16413 (100.01%, -0.01% perf increase)
baseline: 5.94379 (baseline)
mod     : 6.18436 (104.05%, -3.89% perf increase)

$ ./time_buffer_ops.py 
Using Python2
baseline: 5.24699 (baseline)
mod     : 5.25159 (100.09%, -0.09% perf increase)
baseline: 5.17050 (baseline)
mod     : 5.27538 (102.03%, -1.99% perf increase)
baseline: 5.46608 (baseline)
mod     : 5.56813 (101.87%, -1.83% perf increase)
baseline: 5.10800 (baseline)
mod     : 5.41473 (106.01%, -5.66% perf increase)
baseline: 5.06513 (baseline)
mod     : 5.27768 (104.20%, -4.03% perf increase)

```

### Verification Process

Unit tests cover the full scope of changes

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -363,6 +363,9 @@ class DogStatsd(object):
         """
         Flush the buffer and switch back to single metric packets.
         """
+        if not hasattr(self, 'buffer'):
+            raise BufferError('Cannot close buffer that was never opened')
+
         self._send = self._send_to_server
 
         if self.buffer:
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -13,7 +13,7 @@ import os
 import socket
 import errno
 import time
-from threading import Lock
+from threading import Lock, RLock
 
 # Datadog libraries
 from datadog.dogstatsd.context import (
@@ -183,7 +183,7 @@ class DogStatsd(object):
         :type telemetry_socket_path: string
         """
 
-        self.lock = Lock()
+        self._socket_lock = Lock()
 
         # Check for deprecated option
         if max_buffer_size is not None:
@@ -273,6 +273,8 @@ class DogStatsd(object):
         self._telemetry_flush_interval = telemetry_min_flush_interval
         self._telemetry = not disable_telemetry
 
+        self._buffer_lock = RLock()
+
     def disable_telemetry(self):
         self._telemetry = False
 
@@ -311,7 +313,7 @@ class DogStatsd(object):
         Note: connect the socket before assigning it to the class instance to
         avoid bad thread race conditions.
         """
-        with self.lock:
+        with self._socket_lock:
             if telemetry and self._dedicated_telemetry_destination():
                 if not self.telemetry_socket:
                     if self.telemetry_socket_path is not None:
@@ -352,7 +354,12 @@ class DogStatsd(object):
         >>> with DogStatsd() as batch:
         >>>     batch.gauge("users.online", 123)
         >>>     batch.gauge("active.connections", 1001)
+
+        Note: This method must be called before close_buffer() matching invocation.
         """
+
+        self._buffer_lock.acquire()
+
         if max_buffer_size is not None:
             log.warning("The parameter max_buffer_size is now deprecated and is not used anymore")
         self._current_buffer_total_size = 0
@@ -362,15 +369,22 @@ class DogStatsd(object):
     def close_buffer(self):
         """
         Flush the buffer and switch back to single metric packets.
+
+        Note: This method must be called after a matching open_buffer()
+        invocation.
         """
+
         if not hasattr(self, 'buffer'):
             raise BufferError('Cannot close buffer that was never opened')
 
-        self._send = self._send_to_server
+        try:
+            self._send = self._send_to_server
 
-        if self.buffer:
-            # Only send packets if there are packets to send
-            self._flush_buffer()
+            if self.buffer:
+                # Only send packets if there are packets to send
+                self._flush_buffer()
+        finally:
+            self._buffer_lock.release()
 
     def gauge(
         self,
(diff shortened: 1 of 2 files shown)


datadog/dogstatsd/base.py
```diff
@@ -362,12 +374,25 @@ def open_buffer(self, max_buffer_size=None):
...
+        invocation.
         """
-        self._send = self._send_to_server
 
-        if self.buffer:
-            # Only send packets if there are packets to send
-            self._flush_buffer()
+        try:
+            if not hasattr(self, 'buffer'):
+                raise BufferError('Cannot close buffer that was never opened')
```
jd, line 384: That might be correct though it looks like a different change.

datadog/dogstatsd/base.py
```diff
@@ -362,12 +374,25 @@ def open_buffer(self, max_buffer_size=None):
...
+            self._send = self._send_to_server
+
+            if self.buffer:
+                # Only send packets if there are packets to send
+                self._flush_buffer()
+        finally:
+            # Release all locks that might have been held by our thread.
+            while self._buffer_lock_depth > 0:
+                self._buffer_lock_depth -= 1
+                self._buffer_lock.release()
```
jd, line 395: Does this mean you can have multiple `open_buffer` from the same thread and only one `close_buffer`?
That sounds like trying to circumvent programming errors.

datadog/dogstatsd/base.py
sgnn7: The edge case was found during the test writing. It's a separate commit but it can be split out into a separate PR if you want.

datadog/dogstatsd/base.py
sgnn7: > Does this mean you can have multiple open_buffer from the same thread and only one close_buffer?
That sounds like trying to circumvent programming errors.

There was no guidance that an `open_buffer` should match a `close_buffer` should match a `close_buffer`. Even if we did have that guidance, it's not technically a programming error - the following code gets into the tricky territory of "what's the correct behavior expected" that ends up being ambiguous as to what it should do/allow:
```python
statds.open_buffer()
# do something to the buffer

statsd.open_buffer()
# do something else
statsd.close_buffer()
```
The second `open_buffer` in the strictest of terms (which is what you were pointing out I think) seems like it should just throw an exception as it's out-of-order but given the former behavior and its user friendliness it could be interpreted as a "just give me a new buffer to write". We've discussed this a bit on slack before the PR was implemented and chose the latter as a more-appropriate behavior.

As a sidenote, both `open_buffer` and `close_buffer` are getting deprecated soon since we want buffering by default so they will effectively be `noop`s in the future.

datadog/dogstatsd/base.py
```diff
@@ -362,12 +374,25 @@ def open_buffer(self, max_buffer_size=None):
...
+                raise BufferError('Cannot close buffer that was never opened')
+
+            self._send = self._send_to_server
+
+            if self.buffer:
+                # Only send packets if there are packets to send
+                self._flush_buffer()
+        finally:
+            # Release all locks that might have been held by our thread.
+            while self._buffer_lock_depth > 0:
```
ogaca-dd, line 393: Consider the following code:
```python
d = DogStatsd()
# ...
with d as batch:
       with d as other_batch:
             other_batch.gauge("users.online", 123)
       # As the lock was released at the previous line, another thread can call `d.open_buffer()` and update buffer at the same time as `batch.gauge("users.online", 123)`
       batch.gauge("users.online", 123)
```

> As a sidenote, both open_buffer and close_buffer are getting deprecated soon since we want buffering by default so they will effectively be noops in the future

Note: I am not sure it worth handling the case in the previous code as these functions are getting deprecated.

ogaca-dd (approved):

datadog/dogstatsd/base.py
```diff
@@ -363,6 +363,9 @@ def close_buffer(self):
         """
         Flush the buffer and switch back to single metric packets.
         """
+        if not hasattr(self, 'buffer'):
```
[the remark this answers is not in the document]
sgnn7: Great point and I did consider that as an alternative but that option has the unintended effect of ignoring blatantly erroneous behavior when the user does not call `open_buffer` even a single time before trying to close the buffer. While we might not care from the logic point of view if the buffer was created or not beforehand, it made more sense here to notify the user that their code is likely very wrong.

datadog/dogstatsd/base.py
sgnn7: That's an interesting edge case for sure! Let me think about this some more when I get time and I'll get back to you on it.

sgnn7: After some thinking about this particular example and other edge use cases, I think it might be just better to make the lock not re-entrant from the same thread which should fix your and @jd's concerns.

ogaca-dd (approved):
LGTM

PR #645, opened 2021-03-26
[ISSUE-644] Ensure target socket connection is properly considered in initialize

valerian-roche:

### What does this PR do?

It was discovered during an incident that an issue is present in the current initialize method:
* If the socket was already opened with the previous configuration and is not closed
* If the peer configuration is updated (host, port or socket_path)
then:
* If going from UDS to UDP, the change is always ignored
* If updating UDS or UDP and the socket is not closed, the change is ignored

The problem can lead to bad issues if the initial socket does not work (as we are using non-blocking UDP, the library does not detect if the socket is not properly connected)
A current workaround is for the user to call the "close_socket" method after calling initialize

### Description of the Change

Two main changes are done
* Properly reset some variables to ensure users can switch from UDS to UDP
* If a change on host, port or socket_path is done, close the socket to enforce the behavior to change

### Alternate Designs

* Document user to call close_socket after calling initialize (bad user experience)
* Systematically call close_socket in initialize, even if no change is done. As it might impact performance and initialize may be call often to update other flags, this was avoided

### Possible Drawbacks

Might reset the socket more often if users are often changing the target, though that seems like a surprising usecase

### Verification Process

Started with an initial socket targeting a dropped UDP port (through iptables) and send some metrics
Then update the port to a proper port and send further metrics
Confirmed nothing was received by the listener and lsof still reported the socket opened to the initial port

Same test was redone with a UDS setup

### Release Notes

Fix update of target in initialize method when some metrics have already been sent through the dogstatsd client

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -108,15 +108,32 @@ def initialize(
 
     # Statsd configuration
     # ...overrides the default `statsd` instance attributes
+    target_updated = False
     if statsd_socket_path:
-        statsd.socket_path = statsd_socket_path
-        statsd.host = None
-        statsd.port = None
+        if statsd_socket_path != statsd.socket_path:
+            statsd.socket_path = statsd_socket_path
+            statsd.host = None
+            statsd.port = None
+            target_updated = True
     else:
         if statsd_host or statsd_use_default_route:
-            statsd.host = statsd.resolve_host(statsd_host, statsd_use_default_route)
-        if statsd_port:
+            host = statsd.resolve_host(statsd_host, statsd_use_default_route)
+            if host != statsd.host:
+                statsd.host = host
+                target_updated = True
+        if statsd_port and statsd.port != int(statsd_port):
             statsd.port = int(statsd_port)
+            target_updated = True
+
+        # Reset this var, as it is otherwise taking priority still
+        if statsd.socket_path:
+            statsd.socket_path = None
+            target_updated = True
+
+    if target_updated:
+        # We have changed some connection parameters. If the socket was already opened, reset it
+        statsd.close_socket()
+
     if statsd_namespace:
         statsd.namespace = text(statsd_namespace)
     if statsd_constant_tags:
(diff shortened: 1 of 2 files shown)


datadog/__init__.py
```diff
@@ -108,15 +108,32 @@ def initialize(
 
     # Statsd configuration
     # ...overrides the default `statsd` instance attributes
+    target_updated = False
```
sgnn7, line 111: @valerian-roche So I'm a bit curious why we would ever want to leave sockets open if someone is calling initialize. If someone is really reinitializing the configuration, I would assume it's a low frequency event and that we can't reliably predict that leaving the socket open will be handled properly. What are your thoughts on just calling `statsd.close_socket()` before we do anything?

PS: I'm still a bit new to this code so let me know if I'm off-base with my comment

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -147,6 +147,10 @@ def test_initialization(self):
         initialize(**options)
         self.assertEqual(statsd.host, "myhost")
         self.assertEqual(statsd.port, 1234)
+        self.assertIsNone(statsd.socket)
+        with mock.patch('socket.socket'):
+            statsd.get_socket()
+            statsd.socket.connect.assert_called_with(("myhost", 1234))
```
sgnn7, line 153: If we keep your current impl of `initialize`, I think these tests don't adequately cover the `target_updated = False` logic flow.

sgnn7 (changes requested):
@valerian-roche Left a couple of comments

datadog/__init__.py
valerian-roche: Our code does go through it potentially multiple times for different reasons, as each parts set a part of the setup (tags, sink, ...)
The only case I could see justifying closing the socket is when host/port or socket is updated, so I isolated the case

tests/unit/dogstatsd/test_statsd.py
valerian-roche: True, I will add some tests for those cases

PR #650, opened 2021-04-15
Release 0.41.0

zippolyte:
Also add release action and adjust RELEASING.md


.github/workflows/release.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+on:
+  pull_request:
+  release:
+    types:
+      - published
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
```
nmuesch, line 11: Where is this matrix?

zippolyte: I'm returning you the question hahaha

zippolyte: https://github.com/DataDog/datadog-api-client-python/blob/cdd884303b2c2d7959cff6890a7a36d1173bd0b6/.github/workflows/release.yml#L11

nmuesch (approved):

PR #667, opened 2021-06-23
[tests] Improve statsd performance benchmark on MacOS platforms

sgnn7:
### What does this PR do?

Few issues were noticed when running the statsd benchmark on MacOS that needed
fixing

### Description of the Change

- Receive buffer of the socket is now increased to reasonable levels
(32k from 4k)
- Packet count warning was moot since we may send multiple metrics per
packet
- If `statsd.flush` method is available, call it when the threaded
sender exits to ensure that we are accounting for all of the metrics in
the results.

### Verification Process

Run benchmark on MacOS with 10 threads via UDS and see the reduction in packet drops

### DIFF
--- a/tests/util/fake_statsd_server.py
+++ b/tests/util/fake_statsd_server.py
@@ -11,6 +11,7 @@ import socket
 import tempfile
 import threading
 import time
+import warnings
 from multiprocessing import Array, Event, Process, Value
 
 from datadog.util.compat import is_p3k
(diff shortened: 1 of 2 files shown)


tests/util/fake_statsd_server.py
```diff
@@ -64,6 +66,19 @@ def _run_server(self):
...
+
+            # Increase the receiving buffer size where needed (e.g. MacOS has 4k RX
+            # buffers which is half of the max packet size that the client will send.
+            if os.name != 'nt':
+                recv_buff_size = sock.getsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF)
+                if recv_buff_size <= self.MIN_RECV_BUFFER_SIZE:
+                    sock.setsockopt(
+                        socket.SOL_SOCKET,
+                        socket.SO_RCVBUF,
+                        self.MIN_RECV_BUFFER_SIZE,
```
hush-hush, line 78: should it not be `recv_buff_size` ?

sgnn7: Not here. `recv_buff_size` holds the current value of the buffer size and the `self.MIN_RECV_BUFFER_SIZE` is the target we want to set it to.

hush-hush (approved):

Issue #669, opened 2021-06-24
[statsd] Use a buffer by default


sgnn7:
**Is your feature request related to a problem? Please describe.**

`statsd` in its current implementation uses sync operations to transmit data
to the Agent's `statsd` server. This creates a big overhead in latency and CPU usage
for high-throughput environments since we're not batching our requests.

**Describe the solution you'd like**

`statsd` should:
- Perform batching-by-default
- Deprecate `open_buffer`/`close_buffer` functions
- Have ability to set/use old-style sync sending for backwards compatibility if needed

**Describe alternatives you've considered**

**Additional context**

Part of `agent-core` OKR6

PR #670, opened 2021-06-24
[statsd] Enable buffering by default for statsd

sgnn7:
### What does this PR do?

On a high-level, this PR:
- Enables buffering-by-default for `statsd` (300ms automatic flush cycle)
- Adds a `statsd.flush()` helper method
- Deprecates `statsd.open_buffer()` and `statsd.close_buffer()`
- Closes https://github.com/DataDog/datadogpy/issues/669

### Description of the Change

This change turns all sync handling of statsd data into buffered handling
with a thread that forces flushes at a specified interval (default:
300ms, configurable via `flush_interval` kwarg). Because we now do this
buffering by default, the functions `open_buffer()` and `close_buffer()`
are effectively low-level operations and have been deprecated.

To help backwards compatibility where a context manager (which flushes
automatically on exit of the context) is not used, we also now have a
`flush()` function that can be used to force the sending of the metrics
to the server and a `disable_buffering` kwarg in the statsd constructor.

### Possible Drawbacks

- Users of `open_buffer`/`close_buffer` will now start seeing deprecation warnings
- In specific cases where:
   - Context manager is not used
   - `statsd.flush()` is not used
   - Short scripts that exit the interpreter quickly after sending may not have the flush thread able to flush the data to the socket in time before the interpreter exits
 - `statsd` tests run a bit slower since we have to ensure that flushing works as expected in most of the test cases

### Verification Process

- Run the unit tests
- Ensure there are no failures
- (optional) Run the performance benchmark and ensure that you see `Received packets:  100.00%` in the final totals
:
```sh-session
BENCHMARK_TRANSPORT="UDP" \
    BENCHMARK_NUM_THREADS="1" \
    BENCHMARK_NUM_RUNS="10"  \
    BENCHMARK_NUM_DATAPOINTS="50000" \
    python3 -m unittest -vvv tests.performance.test_statsd_throughput 
```

#### Benchmark Data Summary

- ~20-35% reductions across the board in:
  - Overall benchmark duration
  - Sending latency
  - Benchmark CPU usage
- Increased memory usage by the benchmark of about 5-60kb
- Reduction in multithreaded packet drop by almost 100%

#### Detailed Benchmark Data

[Source data](https://gist.github.com/sgnn7/db10961ce8ac8459839548434a9161c6)

```
Python2 - 1 thread - UDP
Duration: -28%
latency: -28%
CPU: -27%
Mem: +16kb

 Python2 - 1 thread - UDS
 Duration: -26%
Latency: -24%
CPU: -22%
Mem: +60kb

Python3 - 1 thread - UDP
Duration: -32%
Latency: -37%
CPU: -36%
Mem: +7kb

Python3 - 1 thread - UDS
Duration: -30%
Latency: -31%
CPU: -29%
 Mem: +12kb
```

### Release Notes

- `statsd` now uses buffering by default (300ms automatic flush interval)
- `statsd.open_buffer()` and `statsd.close_buffer()` methods are superfluous and have been deprecated. Use via a context manager is highly recommended.
- `statsd.flush()` has been added for manual buffer flushing
- `disable_buffer` has been added as a named argument to `DogStatsd()` to revert to old functionality where data was sent synchronously.
- `flush_interval` has been added as a named argument to `DogStatsd()` to allow configuration of auto-flush interval
- `statsd`-created sockets now have a buffer lower limit of 32kb on posix platforms

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -312,14 +312,14 @@ class DogStatsd(object):
 
         # If buffering is disabled, we bypass the buffer function.
         self._send = self._send_to_buffer
-        if disable_buffering is True:
+        if disable_buffering:
             log.info("Statsd buffering is disabled")
             self._send = self._send_to_server
 
         # Start the flush thread if buffering is enabled and the interval is above
         # a reasonable range. This both prevents thrashing and allow us to use "0.0"
         # as a value for disabling the automatic flush timer as well.
-        if disable_buffering is False and flush_interval >= MIN_FLUSH_INTERVAL:
+        if not disable_buffering and flush_interval >= MIN_FLUSH_INTERVAL:
             self._register_flush_thread(flush_interval)
             log.debug(
                 "Statsd flush thread registered with period of %s",
@@ -442,7 +442,7 @@ class DogStatsd(object):
         WARNING: Deprecated method - all operations are now buffered by default.
         Open a buffer to send a batch of metrics.
 
-        To take dvantage of automatic flushing, you should use the context manager instead
+        To take advantage of automatic flushing, you should use the context manager instead
 
         >>> with DogStatsd() as batch:
         >>>     batch.gauge("users.online", 123)
@@ -480,7 +480,7 @@ class DogStatsd(object):
 
     def flush(self):
         """
-        Flush the metrics buffer by sending the data to the server
+        Flush the metrics buffer by sending the data to the server.
         """
         with self._buffer_lock:
             # Only send packets if there are packets to send
@@ -855,7 +855,7 @@ class DogStatsd(object):
 
         if len(string) > 8 * 1024:
             raise Exception(
-                u'Event "{}" payload is too big (>=8KB). Event discarded'.format(
+                u'Event "{0}" payload is too big (>=8KB). Event discarded'.format(
                     title
                 )
             )
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 
-# Unless explicitly stated otherwise all files in this repository are licensed under the BSD-3-Clause License.
-# This product includes software developed at Datadog (https://www.datadoghq.com/).
+# Unless explicitly stated otherwise all files in this repository are licensed under
+# the BSD-3-Clause License. This product includes software developed at Datadog
+# (https://www.datadoghq.com/).
 # Copyright 2015-Present Datadog, Inc
 """
 DogStatsd is a Python client for DogStatsd, a Statsd fork for Datadog.
@@ -12,8 +13,11 @@ import logging
 import os
 import socket
 import errno
+import threading
 import time
 from threading import Lock, RLock
+
+from typing import Optional, List, Text, Union
 
 # Datadog libraries
 from datadog.dogstatsd.context import (
@@ -22,9 +26,9 @@ from datadog.dogstatsd.context import (
 )
 from datadog.dogstatsd.route import get_default_route
 from datadog.util.compat import is_p3k, text
+from datadog.util.deprecation import deprecated
 from datadog.util.format import normalize_tags
 from datadog.version import __version__
-from typing import Optional, List, Text, Union
 
 # Logging
 log = logging.getLogger("datadog.dogstatsd")
@@ -33,10 +37,14 @@ log = logging.getLogger("datadog.dogstatsd")
 DEFAULT_HOST = "localhost"
 DEFAULT_PORT = 8125
 
+# Buffering-related values (in seconds)
+DEFAULT_FLUSH_INTERVAL = 0.3
+MIN_FLUSH_INTERVAL = 0.0001
+
 # Tag name of entity_id
 ENTITY_ID_TAG_NAME = "dd.internal.entity_id"
 
-# Default buffer settings
+# Default buffer settings based on socket type
 UDP_OPTIMAL_PAYLOAD_LENGTH = 1432
 UDS_OPTIMAL_PAYLOAD_LENGTH = 8192
 
@@ -67,26 +75,30 @@ TELEMETRY_FORMATTING_STR = "\n".join(
 )
 
 
+# pylint: disable=useless-object-inheritance,too-many-instance-attributes
+# pylint: disable=too-many-arguments,too-many-locals
 class DogStatsd(object):
     OK, WARNING, CRITICAL, UNKNOWN = (0, 1, 2, 3)
 
     def __init__(
         self,
-        host=DEFAULT_HOST,  # type: Text
-        port=DEFAULT_PORT,  # type: int
-        max_buffer_size=None,  # type: None
-        namespace=None,  # type: Optional[Text]
-        constant_tags=None,  # type: Optional[List[str]]
-        use_ms=False,  # type: bool
-        use_default_route=False,  # type: bool
-        socket_path=None,  # type: Optional[Text]
-        default_sample_rate=1,  # type: float
-        disable_telemetry=False,  # type: bool
+        host=DEFAULT_HOST,                      # type: Text
+        port=DEFAULT_PORT,                      # type: int
+        max_buffer_size=None,                   # type: None
+        flush_interval=DEFAULT_FLUSH_INTERVAL,  # type: float
+        disable_buffering=False,                # type: bool
+        namespace=None,                         # type: Optional[Text]
+        constant_tags=None,                     # type: Optional[List[str]]
+        use_ms=False,                           # type: bool
+        use_default_route=False,                # type: bool
+        socket_path=None,                       # type: Optional[Text]
+        default_sample_rate=1,                  # type: float
+        disable_telemetry=False,                # type: bool
         telemetry_min_flush_interval=(DEFAULT_TELEMETRY_MIN_FLUSH_INTERVAL),  # type: int
-        telemetry_host=None,  # type: Text
-        telemetry_port=None,  # type: Union[str, int]
-        telemetry_socket_path=None,  # type: Text
-        max_buffer_len=0,  # type: int
+        telemetry_host=None,                    # type: Text
+        telemetry_port=None,                    # type: Union[str, int]
+        telemetry_socket_path=None,             # type: Text
+        max_buffer_len=0,                       # type: int
     ):  # type: (...) -> None
         """
         Initialize a DogStatsd object.
@@ -138,6 +150,15 @@ class DogStatsd(object):
 
         :max_buffer_size: Deprecated option, do not use it anymore.
         :type max_buffer_type: None
+
+        :flush_interval: Amount of time in seconds that the flush thread will
+        wait before trying to flush the buffered metrics to the server. If set,
+        it overrides the default value.
+        :type flush_interval: float
+
+        :disable_buffering: If set, metrics are no longered buffered by the client and
+        all data is sent synchronously to the server
+        :type disable_buffering: bool
 
         :param namespace: Namespace to prefix all metric names
         :type namespace: string
@@ -246,7 +267,6 @@ class DogStatsd(object):
         # Socket
         self.socket = None
         self.telemetry_socket = None
-        self._send = self._send_to_server
         self.encoding = "utf-8"
 
         # Options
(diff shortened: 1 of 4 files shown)


datadog/dogstatsd/base.py
```diff
@@ -274,24 +297,66 @@ def __init__(
...
+
+        self._flush_thread = threading.Thread(
+            name="{}_flush_thread".format(self.__class__.__name__),
+            target=_flush_thread_loop,
+            args=(self, sleep_duration,),
+        )
+        self._flush_thread.daemon = True
+        self._flush_thread.start()
+
+    def _dedicated_telemetry_dest(self):
```
jirikuncar, line 351: I am not sure why do we want to rename the method.
```suggestion
    def _dedicated_telemetry_destination(self):
```

datadog/dogstatsd/base.py
```diff
@@ -316,42 +381,68 @@ def get_socket(self, telemetry=False):
         avoid bad thread race conditions.
         """
         with self._socket_lock:
-            if telemetry and self._dedicated_telemetry_destination():
+            if telemetry and self._dedicated_telemetry_dest():
```
jirikuncar, line 384:
```suggestion
            if telemetry and self._dedicated_telemetry_destination():
```

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -545,7 +756,7 @@ def func(arg1, arg2, kwarg1=1, kwarg2=1):
         self.assertEqual('tests.unit.dogstatsd.test_statsd.func', name)
         self.assert_almost_equal(0.5, float(value), 0.1)
 
-    @pytest.mark.skipif(not is_higher_py35(), reason="Coroutines are supported on Python 3.5 or higher.")
+    @unittest.skipIf(not is_higher_py35(), reason="Coroutines are supported on Python 3.5 or higher.")
```
jirikuncar, line 759: We are using `pytest` to run unit tests in `tox.ini`. Both decorators should be supported but probably we should not change it to the same PR.
```suggestion
    @pytest.mark.skipif(not is_higher_py35(), reason="Coroutines are supported on Python 3.5 or higher.")
```

ogaca-dd (approved):
LGTM

datadog/dogstatsd/base.py
```diff
@@ -274,24 +297,66 @@ def __init__(
...
+        self._send = self._send_to_buffer
+        if disable_buffering is True:
+            log.info("Statsd buffering is disabled")
+            self._send = self._send_to_server
+
+        # Start the flush thread if buffering is enabled and the interval is above
+        # a reasonable range. This both prevents thrashing and allow us to use "0.0"
+        # as a value for disabling the automatic flush timer as well.
+        if disable_buffering is False and flush_interval >= MIN_FLUSH_INTERVAL:
+            self._register_flush_thread(flush_interval)
```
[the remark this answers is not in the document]
sgnn7: If you disable the automatic flush thread, the assumption is that the user will be doing their own lifecycle management via `statsd_instance.flush()`

datadog/dogstatsd/base.py
```diff
@@ -773,14 +878,15 @@ def service_check(self, check_name, status, tags=None, timestamp=None, hostname=
 
         if self._telemetry:
             self.service_checks_count += 1
-        self._send(string)
+
+        self._send_to_buffer(string)
```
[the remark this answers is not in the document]
sgnn7: That's a good catch - I'll fix that up!

datadog/dogstatsd/base.py
sgnn7: It was a change to fix a [Pylint warning C0103](http://pylint-messages.wikidot.com/messages:c0103). It effectively was saying that the name was invalid because its length was >=30 characters long and trimming it a bit seemed like an easy fix. If you want the old name back, that's fine - I was just trying to remove the pylint noise during development.

tests/unit/dogstatsd/test_statsd.py
sgnn7: I wasn't using tox in my local dev env and the base `unittest` library does not recognize `@pytest` so I kept getting test failures. Since `unittest` is the lowest common denominator in testing and the `@unittest.skipIf` works fine with both frameworks I figured that it made sense not to use a third-party-specific one to support more development envs. I can split this out into a separate PR for sure but I think that overall there's no need to use a bespoke decorator if the built-in `unittest` module one works (especially if it covers more use cases).

datadog/dogstatsd/base.py
jirikuncar: I think we are fine with the old name for private method.

sgnn7: Reverted to old name

datadog/dogstatsd/base.py
sgnn7: Reverted to old name

datadog/dogstatsd/base.py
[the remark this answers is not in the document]
sgnn7: Fixed

datadog/dogstatsd/base.py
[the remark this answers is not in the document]
ogaca-dd: > If you disable the automatic flush thread, the assumption is that the user will be doing their own lifecycle management via statsd_instance.flush()

What do you think about documenting this behavior (If not already done) as `flush_interval=0` could also means "flush automatically for every metric"?

> Also, the packet size splitter logic will send the data out when the packet is full regardless of if you have a background thread or not.

Consider the following code:
```python
options = {
    'statsd_host':'127.0.0.1',
    'statsd_port':8125
}

initialize(**options)
statsd.increment('my_application.start', tags=["environment:dev"])
long_running_method_that_does_not_use_statsd()
statsd.flush()
```
In the previous code, the metric `'my_application.start'` can be sent few minutes after the start of the application and let the user believes that the metric is not sent. This kind of code is common when testing the first time statsd (I had a couple of support cards related to that).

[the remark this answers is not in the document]
sgnn7: > What do you think about documenting this behavior (If not already done) as flush_interval=0 could also means "flush automatically for every metric"?

Indeed - the documentation will be following this PR (AC-1040)

> In the previous code, the metric 'my_application.start' can be sent few minutes after the start of the application and let the user believes that the metric is not sent. This kind of code is common when testing the first time statsd (I had a couple of support cards related to that).

If I understand your comment I don't think this will be a problem here as `flush` is a new addition to the API and the delay would only be relevant if they set `disable_buffering=True` since the flush thread will send whatever partial-packet data we have automatically within the default interval (300ms) .

ogaca-dd (dismissed):

datadog/dogstatsd/base.py
```diff
@@ -319,39 +384,65 @@ def get_socket(self, telemetry=False):
...
         return sock
 
+    @deprecated("Statsd module now uses buffering by default.")
     def open_buffer(self, max_buffer_size=None):
         """
+        WARNING: Deprecated method - all operations are now buffered by default.
         Open a buffer to send a batch of metrics.
 
-        You can also use this as a context manager.
+        To take dvantage of automatic flushing, you should use the context manager instead
```
jirikuncar, line 445:
```suggestion
        To take advantage of automatic flushing, you should use the context manager instead
```

datadog/dogstatsd/base.py
```diff
@@ -360,33 +451,42 @@ def open_buffer(self, max_buffer_size=None):
...
+            self._manual_buffer_lock.release()
+
+    def _reset_buffer(self):
+        with self._buffer_lock:
+            self._current_buffer_total_size = 0
+            self._buffer = []
+
+    def flush(self):
+        """
+        Flush the metrics buffer by sending the data to the server
```
jirikuncar, line 483:
```suggestion
        Flush the metrics buffer by sending the data to the server.
```

datadog/dogstatsd/base.py
```diff
@@ -743,19 +854,32 @@ def event(
             string = "%s|#%s" % (string, ",".join(tags))
 
         if len(string) > 8 * 1024:
-            raise Exception(u'Event "%s" payload is too big (more than 8KB), ' "event discarded" % title)
+            raise Exception(
+                u'Event "{}" payload is too big (>=8KB). Event discarded'.format(
```
jirikuncar, line 858:
```suggestion
                u'Event "{0}" payload is too big (>=8KB). Event discarded'.format(
```

datadog/util/deprecation.py
```diff
@@ -0,0 +1,24 @@
...
+import warnings
+from functools import wraps
+
+
+def deprecated(message):
+    def deprecated_decorator(func):
+        @wraps(func)
+        def deprecated_func(*args, **kwargs):
+            warnings.warn(
+                "'{}' is a deprecated function. {}".format(func.__name__, message),
```
jirikuncar, line 14:
```suggestion
                "'{0}' is a deprecated function. {1}".format(func.__name__, message),
```

datadog/dogstatsd/base.py
sgnn7: @jirikuncar Hmm - isn't `{0}` implied by `{}`? I don't mind changing it - it just seems a bit superfluous.

datadog/util/deprecation.py
sgnn7: Fixed via https://github.com/DataDog/datadogpy/pull/670/commits/c0f08ca3f712c8134183dd709d5a387b62c018ce

datadog/dogstatsd/base.py
sgnn7: Fixed via https://github.com/DataDog/datadogpy/pull/670/commits/c0f08ca3f712c8134183dd709d5a387b62c018ce

datadog/dogstatsd/base.py
sgnn7: Fixed via https://github.com/DataDog/datadogpy/pull/670/commits/c0f08ca3f712c8134183dd709d5a387b62c018ce

jirikuncar (dismissed):

datadog/dogstatsd/base.py
```diff
@@ -274,24 +297,66 @@ def __init__(
...
+        self._reset_buffer()
+
+        # This lock is used for backwards compatibility to prevent concurrent
+        # changes to the buffer when the user is managing the buffer themselves
+        # via deprecated `open_buffer()` and `close_buffer()` functions.
+        self._manual_buffer_lock = RLock()
+
+        # If buffering is disabled, we bypass the buffer function.
+        self._send = self._send_to_buffer
+        if disable_buffering is True:
```
therve, line 315:
```suggestion
        if disable_buffering:
```

datadog/dogstatsd/base.py
```diff
@@ -274,24 +297,66 @@ def __init__(
...
+            log.info("Statsd buffering is disabled")
+            self._send = self._send_to_server
+
+        # Start the flush thread if buffering is enabled and the interval is above
+        # a reasonable range. This both prevents thrashing and allow us to use "0.0"
+        # as a value for disabling the automatic flush timer as well.
+        if disable_buffering is False and flush_interval >= MIN_FLUSH_INTERVAL:
```
therve, line 322:
```suggestion
        if not disable_buffering and flush_interval >= MIN_FLUSH_INTERVAL:
```

datadog/dogstatsd/base.py
```diff
@@ -274,24 +297,66 @@ def __init__(
...
+            self._register_flush_thread(flush_interval)
+            log.debug(
+                "Statsd flush thread registered with period of %s",
+                flush_interval,
+            )
+        else:
+            log.info("Statsd periodic buffer flush is disabled")
```
therve, line 329: Don't we get in that else if `disable_buffering is set` ? I think it would make more sense if it was only printed when `disable_buffering` is false.

sgnn7: The message gets shown for either:
- Buffering being explicitly disabled
- Or the flush interval is below the minimum allowed (explicit flush thread disabling)
So I think the message gets shown for the appropriate conditions

datadog/dogstatsd/base.py
sgnn7: Fixed in https://github.com/DataDog/datadogpy/pull/670/commits/1c83052e6b4990d0240c180b17e416937cfa8cad

datadog/dogstatsd/base.py
sgnn7: Fixed in https://github.com/DataDog/datadogpy/pull/670/commits/1c83052e6b4990d0240c180b17e416937cfa8cad

therve (approved):

sgnn7:
I'll try to kick off the integration tests again but it looks like an unrelated section of code and the test also passes locally:

### Failure

```
tests/integration/dogshell/test_dogshell.py::TestDogshell::test_screenboards FAILED 
```

### Local run
```
$ tox -e integration --  tests/integration/dogshell/*
...
tests/integration/dogshell/test_dogshell.py::TestDogshell::test_screenboards PASSED 
```

ogaca-dd (approved):

PR #678, opened 2021-07-12
[tests] Statsd fake server can now be run as a standalone script

sgnn7:
### What does this PR do?

Allows the `./tests/util/fake_statsd_server` module to be run directly from the shell

### Description of the Change

There may be times where having a quick fake standalone statsd server is
advantageous (e.g. writing/updating client docs). While this can be
achieved with some python scripts or the interpreter shell, it's much
easier for the user/developer just to invoke the script itself if
possible which is what this change enables.

To use this fake server you just invoke the script:

```
$ # UDS Server (default)
$ ./tests/util/fake_statsd_server.py
Listening via UDS on /tmp/dir/path/FakeServerabcd1234/fake_statsd_server_socket

$ # UDP Server
$ ./tests/util/fake_statsd_server.py udp
Listening via UDP on port 58445
```

Debugging is enabled by default which should print out the received packets to the
console when run.

### Verification Process

- Run `./tests/util/fake_statsd_server.py`
- Use `statsd` module to try to send data to this endpoint
- Ensure that the fake server prints the received packets

### Additional Notes

`is_p3k` use was changed to use stdlib to ensure we didn't need to import helpers from parent modules
which would have required `sys.path` mangling.

### Release Notes

N/A

### DIFF
--- a/tests/util/fake_statsd_server.py
+++ b/tests/util/fake_statsd_server.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 # Unless explicitly stated otherwise all files in this repository are licensed
 # under the BSD-3-Clause License. This product includes software developed at
 # Datadog (https://www.datadoghq.com/).
@@ -8,13 +9,11 @@ import ctypes
 import os
 import shutil
 import socket
+import sys
 import tempfile
 import threading
 import time
-import warnings
 from multiprocessing import Array, Event, Process, Value
-
-from datadog.util.compat import is_p3k
 
 # pylint: disable=too-many-instance-attributes,useless-object-inheritance
 class FakeServer(object):
@@ -28,7 +27,7 @@ class FakeServer(object):
     ALLOWED_TRANSPORTS = ["UDS", "UDP"]
     MIN_RECV_BUFFER_SIZE = 32 * 1024
 
-    def __init__(self, transport="UDS", debug=False):
+    def __init__(self, transport="UDS", ignore_timeouts=True, debug=False):
         if transport not in self.ALLOWED_TRANSPORTS:
             raise ValueError(
                 "Transport {} is not a valid transport type. Only {} are allowed!".format(
@@ -38,6 +37,7 @@ class FakeServer(object):
             )
 
         self.transport = transport
+        self.ignore_timeouts = ignore_timeouts
         self.debug = debug
 
         self.server_process = None
@@ -80,9 +80,12 @@ class FakeServer(object):
 
             sock.bind(socket_path)
 
+            if self.debug:
+                print("Listening via UDS on", socket_path)
+
             # We are using ctypes for shmem so we have to use a consistent
             # datatype across Python versions
-            if is_p3k():
+            if sys.version_info[0] > 2:
                 self._socket_path.value = socket_path.encode("utf-8")
             else:
                 self._socket_path.value = socket_path
@@ -93,6 +96,9 @@ class FakeServer(object):
             sock.bind(("", 0))
 
             _, self._port.value = sock.getsockname()
+
+            if self.debug:
+                print("Listening via UDP on port", self.port)
 
         # Run an async thread to update our shared-mem counters. We don't want to update
         # the shared memory values when we get data due to performance reasons.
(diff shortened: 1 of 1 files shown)


tests/util/fake_statsd_server.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
```
jirikuncar, line 1: Is Python3 required?

```suggestion
#!/usr/bin/env python
```

tests/util/fake_statsd_server.py
```diff
@@ -80,9 +80,12 @@ def _run_server(self):
 
             sock.bind(socket_path)
 
+            if self.debug:
+                print("Listening via UDS on", socket_path)
+
             # We are using ctypes for shmem so we have to use a consistent
             # datatype across Python versions
-            if is_p3k():
+            if sys.version_info[0] > 2:
```
jirikuncar, line 88: why don't we use the common utility function?

tests/util/fake_statsd_server.py
```diff
@@ -179,3 +194,17 @@ def __repr__(self):
         return "<FakeThreadedUDPServer(Packets RX: {}, Metrics RX: {}".format(
             self._payload_counter_shmem_var.value, self._metric_counter_shmem_var.value
         )
+
+
+if __name__ == '__main__':
+    options = {
+      'ignore_timeouts': False,
+      'debug': True,
```
jirikuncar, line 202: Should it be configurable?
```suggestion
      'debug': bool(sys.environ.get("FAKE_STATSD_SERVER_DEBUG", False)),
```

tests/util/fake_statsd_server.py
sgnn7: I mentioned it briefly in the PR description but we need to do some path mangling to do the import (`datadog.util.compat`) if I used the util since the script will run with the `sys.path` set to `datadogpy/tests/util` and not just `datadogpy` when run from CLI. It seemed like a simple conversion vs doing some gymnastics with the `sys.path.append` that would also be fragile.

tests/util/fake_statsd_server.py
sgnn7: Yeah that should work too - it's not required.

tests/util/fake_statsd_server.py
sgnn7: > Should it be configurable?

I think having a silent test server would not have a valid use case here since otherwise you wouldn't be able to confirm that the data arrived, that it matches expectations, nor would you see what port/path it's listening on. I think if there's a use case for it later, we can certainly have it as an option it's just that it seemed unnecessary here.

tests/util/fake_statsd_server.py
jirikuncar: Can it be that you are running the server with different python than the one used to run tests? The other possibility would be running the script as `python tests/util/fake_statsd_server.py` in the virtualenv where `datadog` is installed.

sgnn7: No, this is just a general issue of python importing. When you use the shebang line (`#!/...`), the initial "root" `sys.path` is set to the parent dir of the script itself (not the cwd). Normally this isn't a problem for third-party imports but if you're importing something from the same filesystem tree, you then have to change paths as `os.path.join(sys.path, "datadog")` doesn't exist from that starting dir. Things get even more complicated here as we need something from a path that has to go back up the tree out of the `tests` directory.

To work around this issue and to make it work on both regular tests and as a runnable script, I believe that there's only a few options that can be used:
- Remove dependencies from the parent paths (the approach I took here)
- Append parent path manually to `sys.path` (fragile)
- Mangle the `PYTHONPATH` variable (fragile, bad UX)
- Use a conditional `from .... import something` surrounded by `try/except ImportError` block to account for two different ways to run it (multiple logic paths, fragile, might not even work when run as a standalone script)

jirikuncar: The alternative would be registering the script in `console_scripts` (or custom one and load it with `pkg_resources`) possibly guarded with an optional `extras` or a condition.

We can more forward with current implementation as the "cure" for parent import issue is easy in this case.

sgnn7: > The alternative would be registering the script in console_scripts (or custom one and load it with pkg_resources) possibly guarded with an optional extras or a condition.

That's true for an installed package but wouldn't work if someone just checked out the repo and tried to run it I think.

> We can more forward with current implementation as the "cure" for parent import issue is easy in this case.

👍 Do you mind approving the PR or is something else in need of fixing?

jirikuncar (approved):

Issue #652, opened 2021-05-12
Using dogshell with keys supplied in env variables doesn't work for datadoghq.eu


janorn:
If you try to supply API and APP keys via env variables in a container and your Datadog site is datadoghq.**eu** it will fail.
This is due to dogshell lacking a site config parameter or environment variable. 

I have tried a workaround by supplying the site config in a dogrc file while the KEYs via env variables.
However due to this code the conf file will never be used if you supply keys via env variables.

https://github.com/DataDog/datadogpy/blob/ec571f6593c983c0228fd7b6019bced1c5e7fc93/datadog/dogshell/common.py#L49-56

Adding the keys to a config file in the container is not a viable solution.

janorn:
Just to clarify. It does not "fail" with an error. But the posted data ends up in datadoghq.com

aquiladayc:
@janorn Hey, would you try this? It worked for me.
`export DATADOG_HOST=https://api.datadoghq.eu/` and run dog commmand.

janorn:
Seems to work

```bash
(dogshell) root@laptop:~# env | grep DATADOG
DATADOG_API_KEY=XXXXXXXXXXXXXXXXXXXX
DATADOG_APP_KEY=YYYYYYYYYYYYYYYYYYYYYYYYYYYYY
DATADOG_HOST=https://api.datadoghq.eu/
(dogshell) root@laptop:~# dogshell search query test
metrics synthetics.test_runs
metrics datadog.estimated_usage.synthetics.api_test_runs
metrics synthetics.pl.worker.pulled_tests
metrics synthetics.browser.test_duration
metrics datadog.estimated_usage.synthetics.browser_test_runs
```

However it doesn't seem to work with --site.

```bash
(dogshell) root@laptop:~# env | grep DATADOG
(dogshell) root@laptop:~# dogshell --api-key XXXXXXXXXXXXXXXXXXXX--application-key YYYYYYYYYYYYYYYYYYYYYYYYYYYYY--site https://api.datadoghq.eu/ search query test
usage: dogshell [-h] [--config CONFIG] [--api-key API_KEY] [--application-key APP_KEY] [--pretty] [--raw] [--timeout TIMEOUT] [-v]  {comment,search,metric,tag,event,monitor,timeboard,dashboard,screenboard,dashboard_list,host,downtime,service_check,service_level_objective} ...
[...]

PR #691, opened 2021-10-08
Add an option to change Datadog site to call API

aquiladayc:
### What does this PR do?

Issue to fix
https://github.com/DataDog/datadogpy/issues/652

### Description of the Change

Adding `--site` option to change the site to call the API

### Verification Process

Tested with my sandbox account in EU and US3 site.

### DIFF
--- a/datadog/dogshell/__init__.py
+++ b/datadog/dogshell/__init__.py
@@ -71,6 +71,14 @@ def main():
         "-v", "--version", help="Dog API version", action="version", version="%(prog)s {0}".format(__version__)
     )
 
+    parser.add_argument(
+        "--site",
+        help="Datadog site to send data, us (datadoghq.com), eu (datadoghq.eu) or us3 (us3.datadoghq.com), default: us",
+        dest="site",
+        default=os.environ.get("DD_SITE", os.environ.get("DATADOG_HOST")),
+        choices=["datadoghq.com", "us", "datadoghq.eu", "eu", "us3.datadoghq.com", "us3"],
+    )
+
     config = DogshellConfig()
 
     # Set up subparsers for each service
@@ -93,7 +101,8 @@ def main():
     ServiceLevelObjectiveClient.setup_parser(subparsers)
 
     args = parser.parse_args()
-    config.load(args.config, args.api_key, args.app_key)
+
+    config.load(args.config, args.api_key, args.app_key, args.site)
 
     # Initialize datadog.api package
     initialize(**config)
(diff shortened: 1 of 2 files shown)


aquiladayc:
@DataDog/integrations-tools-and-libraries can anyone review please?

datadog/dogshell/__init__.py
```diff
@@ -71,6 +71,14 @@ def main():
         "-v", "--version", help="Dog API version", action="version", version="%(prog)s {0}".format(__version__)
     )
 
+    parser.add_argument(
+        "--site",
+        help="Datadog site to send data, us (datadoghq.com), eu (datadoghq.eu) or us3 (us3.datadoghq.com), default: us",
+        dest="site",
+        default=os.environ.get("DD_SITE", os.environ.get("DATADOG_HOST")),
```
therve, line 78: Site and host have different semantic, I don't think we should use the same env var for this.

datadog/dogshell/__init__.py
```diff
@@ -71,6 +71,14 @@ def main():
...
+        choices=["datadoghq.com", "us", "datadoghq.eu", "eu", "us3.datadoghq.com", "us3"],
```
therve, line 79: Are you sure we want support `us` and `eu` ? That could be confusing with other datacenters.

datadog/dogshell/common.py
```diff
@@ -43,8 +43,16 @@ def report_warnings(res):
...
+    def load(self, config_file, api_key, app_key, site):
         config = configparser.ConfigParser()
+        
+        if site is not None:
+            if site in ("us" or "datadoghq.com"):
+                self["api_host"] = "https://datadoghq.com"
+            elif site in ("datadoghq.eu", "eu"):
+                self["api_host"] = "https://datadoghq.eu"
+            elif site in ("us3.datadoghq.com", "us3"):
+                self["api_host"] = "https://us3.datadoghq.com"
```
therve, line 55: We should probably raise an error if the value doesn't match.

therve (changes requested):
Please fix the lint errors as well.

datadog/dogshell/__init__.py
aquiladayc: I thought we can be aligned with https://github.com/DataDog/datadogpy/blob/master/datadog/dogshell/wrap.py#L289
What do you think? Should we remove `us`, `eu`, `us3`?

aquiladayc:
Updated, can you review please?

janorn:
I know this has already been merged but just a tiny comment. The new option --api_host uses and under score where both other api options just uses a hyphen.
--api-key
--application-key

PR #692, opened 2021-10-13
[statsd] Disable statsd buffering by default

sgnn7:
### What does this PR do?

Due to impact on users of other clients where buffering was enabled by
default, especially in environments and frameworks where `fork()` is
used we will for the time being disable buffering by default until a
decision is made on the path forward. Buffering can still be turned on
with `disable_buffering = False` flag it's just that for now it is
defaulted to `True`.

### Description of the Change

- Use of `DogStatsd` via `statsd` module-level object defaults to synchronous (un-buffered) metric sending.
- Use of `DogStatsd` via the default constructor defaults to synchronous (un-buffered) metric sending.

### Alternate Designs

In discussion (TBD). We may revert this and use automated ways to detect environments
that are incompatible to thread-based buffering.

### Possible Drawbacks

Lower performance in high-throughput environments.

### Verification Process

Unit tests cover most of this:

```sh-session
python$(python --version 2>&1 | cut -c8-8)" -m unittest -vvv tests.unit.dogstatsd.test_statsd
```

### Additional Notes

This change may be reversed in the future but for now, the risk is too high
for user disruption.

### Release Notes

Disable statsd buffering by default. Buffering can be enabled with `disable_buffering = False` flag.

_Note: Since this is just a partial revert of https://github.com/DataDog/datadogpy/pull/670, removal of both release notes can be done as it's a net-0 change_

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -26,7 +26,6 @@ from datadog.dogstatsd.context import (
 )
 from datadog.dogstatsd.route import get_default_route
 from datadog.util.compat import is_p3k, text
-from datadog.util.deprecation import deprecated
 from datadog.util.format import normalize_tags
 from datadog.version import __version__
 
@@ -89,7 +88,7 @@ class DogStatsd(object):
         port=DEFAULT_PORT,                      # type: int
         max_buffer_size=None,                   # type: None
         flush_interval=DEFAULT_FLUSH_INTERVAL,  # type: float
-        disable_buffering=False,                # type: bool
+        disable_buffering=True,                 # type: bool
         namespace=None,                         # type: Optional[Text]
         constant_tags=None,                     # type: Optional[List[str]]
         use_ms=False,                           # type: bool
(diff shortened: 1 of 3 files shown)


gh123man (approved):

gh123man (approved):

datadog/dogstatsd/base.py
```diff
@@ -453,16 +450,16 @@ def open_buffer(self, max_buffer_size=None):
 
         self._manual_buffer_lock.acquire()
 
+        # XXX Remove if `disable_buffering` default is changed to False
+        self._send = self._send_to_buffer
```
truthbk, line 454: This stuff is not thread-safe. I presume it never really was and I assume that's beyond the expected use-cases.

sgnn7: @truthbk Generally things should be thread-safe due to various locks on the socket/buffer/context ops but you may get unexpectedly-buffered metrics if some threads are opening/closing buffers while another thread is just writing metrics without a context manager or `open_buffer`. Overall though, there should be no data loss or exceptions. As for the setter here, the lock + GIL makes this effectively an atomic operation. We do have a test around at least part of this [here](https://github.com/DataDog/datadogpy/blob/33f727cc51d637860a2a46677bf7151c8feb10fa/tests/unit/dogstatsd/test_statsd.py#L711). If I'm missing something though, I'll fix up the PR for sure.

> I presume it never really was and I assume that's beyond the expected use-cases.

Somewhat. The way it was coded originally was not really thread safe but the changes made subsequently over this year have closed up most of the glaring exception issues. The buffering-by-default fixed the `self._send =` GIL reliance but we're backing that part out with this PR for the time being 😢 .

therve (approved):

PR #700, opened 2021-11-10
[statsd] Add ability to toggle `statsd.disable_buffering` state during runtime

sgnn7:
### What does this PR do?

- Adds ability to set `statsd_instance.disable_buffering` flag after initialization
- Adds said flag as a argument to `initialize()` to be usable for the module-level `statsd` instance

### Description of the Change

While not likely to be useful currently, if/when we enable buffering on the `DogStatsd` clients
by default again we will want to have this feature already in place since the module level `statsd`
may not behave as expected in `fork()`-based environments. Because `datadog.initialize()` cannot
recreate the instance of the module-level `statsd`, we need a runtime setting for `disable_buffering`
instead of recreating the instance, making the implementation somewhat constrained in how it is
done.

As a side-effect, this change allow use of the module-level `statsd` instance with buffering enabled
which was previously not possible.

### Alternate Designs

None that would make sense with the limitations provided.

### Possible Drawbacks

Assumptions were made in the logic that `datadog.initialize()` would be only called once for
the application. While there are protections around concurrent invocations, there could be edge
cases not covered if multiple concurrent calls to `initialize are made to change this flag or if the
data is being added to the buffer while the buffering is being disabled.

### Verification Process

Tests covers the underlying functionality of this feature but it can be tested with:
- Using either:
  - `datadog.initialize()` and setting `statsd_disable_buffering` flag
  - Instantiating `DogStatsd`
- Verifying that data flows through

### Additional Notes

This feature will be gradually documented as it is expected to be
useful only in very niche high-throughput cases.

### Release Notes

- [statsd] Add ability to toggle `statsd.disable_buffering` state during runtime
- [datadog] Add ability to use `statsd_disable_buffering` as an argument to `datadog.initialize()`

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -37,6 +37,7 @@ def initialize(
     api_host=None,  # type: Optional[str]
     statsd_host=None,  # type: Optional[str]
     statsd_port=None,  # type: Optional[int]
+    statsd_disable_buffering=True,  # type: bool
     statsd_use_default_route=False,  # type: bool
     statsd_socket_path=None,  # type: Optional[str]
     statsd_namespace=None,  # type: Optional[str]
@@ -70,6 +71,10 @@ def initialize(
 
     :param statsd_port: Port of DogStatsd server or statsd daemon
     :type statsd_port: port
+
+    :param statsd_disable_buffering: Enable/disable statsd client buffering support
+                                     (default: True).
+    :type statsd_disable_buffering: boolean
 
     :param statsd_use_default_route: Dynamically set the statsd host to the default route
                                      (Useful when running the client in a container)
@@ -121,6 +126,8 @@ def initialize(
         statsd.namespace = text(statsd_namespace)
     if statsd_constant_tags:
         statsd.constant_tags += statsd_constant_tags
+
+    statsd.disable_buffering = statsd_disable_buffering
 
     api._return_raw_response = return_raw_response
 
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -304,28 +304,24 @@ class DogStatsd(object):
 
         self._reset_buffer()
 
-        # This lock is used for backwards compatibility to prevent concurrent
-        # changes to the buffer when the user is managing the buffer themselves
-        # via deprecated `open_buffer()` and `close_buffer()` functions.
-        self._manual_buffer_lock = RLock()
+        # This lock is used for all cases where buffering functionality is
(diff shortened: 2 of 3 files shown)


therve (approved):

djmitche (commented):
I don't think I have enough context to approve this, but it looks good to me as far as I understand.  I'll re-request agent-core review.

datadog/__init__.py
```diff
@@ -37,6 +37,7 @@ def initialize(
     api_host=None,  # type: Optional[str]
     statsd_host=None,  # type: Optional[str]
     statsd_port=None,  # type: Optional[int]
+    statsd_disable_buffering=True,  # type: bool
```
remeh, line 40: Is it named `disable_buffering` because the plan is to have the buffering enabled by default in the future? If that's the case, don't mind the following remark.
Having a parameter "disabling" something (and defaulting to True) is always weird to think about opposed to a parameter "enabling" a feature.

datadog/dogstatsd/base.py
```diff
@@ -304,59 +304,109 @@ def __init__(
...
+            return
+
+        try:
+            self.flush()
+        finally:
+            pass
+
+        self._flush_thread_stop.set()
+
+        self._flush_thread.join()
```
remeh, line 369: Not really a big deal I guess but if two threads are calling `_stop_flush_thread` simultaneously, there is a unlucky chance that `join()` could be called on one the thread while the `_flush_thread` has already be terminated by the others.

sgnn7: True that this method isn't thread-safe in isolation but [the comment](https://github.com/DataDog/datadogpy/pull/700/files/5534f60dec123e06273e8b258e8b2869a7b375e4#diff-c9e70dcb34e0c8aa6f036a6d228d459d9b57677a7bb41a78372b42265cc51c55R356) on the method mentions this as a warning and [the only caller is lock-protected](https://github.com/DataDog/datadogpy/pull/700/files/5534f60dec123e06273e8b258e8b2869a7b375e4#diff-c9e70dcb34e0c8aa6f036a6d228d459d9b57677a7bb41a78372b42265cc51c55R393-R404). I would have preferred to use the lock directly in this method like you imply but then the toggle to stop would needed 2 lock acquires (1 for toggle method and 1 for stop method; possibly a reentrant one) which seemed like an overkill and over-complication for a helper only called from a single spot.

remeh (approved):
I don't know the codebase extremely well but changes in this PR looks good to me!

PR #703, opened 2021-12-01
Add support for DD_TAGS environment variable

### LINKED ISSUE #702: Allow use of DD_TAGS environment variable for constant tags
`DD_TAGS` is the common environment variable used for constant tagging in other datadog services (such as APM),  and is what is referenced in most documentation.

This library uses `DATADOG_TAGS`, and does not support `DD_TAGS`.

Documentation for [assigning tags ](https://docs.datadoghq.com/getting_started/tagging/assigning_tags) references `DD_TAGS`, but does not mention `DATADOG_TAGS`.

**Expected behavior**
Constant tagging for metrics created should utilize the same environment values as other datadog services. 

We shouldn't have to specify multiple environment variables with the same values.

nrmitchi:
Falls back to reading global tags from the DD_TAGS env var if
DATADOG_TAGS is not present.

DD_TAGS used for other datadog products, such as APM.

### What does this PR do?
Adds support for using constant tags from the `DD_TAGS` environment variable, rather than the `DATADOG_TAGS` environment variable.

This PR closes #702 

### Description of the Change

If the `DATADOG_TAGS` environment variable is not present, fall back to the `DD_TAGS` environment variable for constant tags. This allows this library to match the behaviour of other datadog services (as well as the documentation).

By using `DD_TAGS` only if `DATADOG_TAGS` is not present, this change should be backwards compatible with existing usage.

### Alternate Designs

An alternative was a straight replacement of the environment variable, however a fallback was chosen to maintain backwards compatibility and avoid dropping of existing tags.

### Possible Drawbacks

It is possible that if a user is currently _not_ using the `DATADOG_TAGS` environment variable for constant tags, but is using the `DD_TAGS` environment variable for a different datadog service (such as APM), those tags will be added to custom metrics. This could lead to an unexpected increase in metric cardinality.

### Additional Notes

Currently duplicating all constant tags into both `DD_TAGS` and `DATADOG_TAGS` environment variables is prone to error.

### Release Notes

If you are currently utilizing `DD_TAGS` environment variable for other datadog functionality, and not utilizing `DATADOG_TAGS` for this library, your existing `DD_TAGS` will be added to existing metrics.


NouemanKHAL:
Can you update these docstrings accordingly ? 
https://github.com/DataDog/datadogpy/blob/33f727cc51d637860a2a46677bf7151c8feb10fa/datadog/threadstats/base.py#L55
https://github.com/DataDog/datadogpy/blob/6bf457859e23da8847291536e54b00e3b7023d5b/datadog/dogstatsd/base.py#L118

Thanks!

therve (approved):

hush-hush:
Hi @nrmitchi,

Thanks for opening this PR. Adding support for `DD_TAGS` across all our DogStatsD clients is something we're looking into.

The main issue is that when a env var `DD_TAGS` is set for the whole host and the agent/client are running locally it will be picked by both the Agent and the DogStatsD client so we have to be careful. This is especially important when using containers, VMs and k8s deployments. The last thing we want is for the clients or Agent to assign the wrong tags to a metric by default.

We're also looking into aligning the behavior between all clients around this, which is a bit all over the place right now.

In the meantime, clients that want to fetch tags from the env can easily do so using the library constructor. It's not the best but it's an easy workaround.

PR #713, opened 2022-02-17
Update package metadata

ofek:
### What does this PR do?

This implements [PEP 621](https://www.python.org/dev/peps/pep-0621/), obviating the need for both `setup.py` and `MANIFEST.in`. Support has not landed in `setuptools`, so builds will now use `hatchling`. We also did this for all integrations: https://github.com/DataDog/integrations-core

This also drops support for EOL Python 3.x versions since `hatchling` and other build backends like `setuptools` have already done so. This will not break anything because `pip` only looks for version candidates that satisfy the `requires-python` field. Here are stats for the curious: https://pypistats.org/packages/datadog

Note that the `setup.py` file will be removed when we drop Python 2 since new-style editable installations require versions of `pip` that are Python 3-only. Basically, this is just for the `py27` test `tox` environment.

### DIFF
--- a/.azure-pipelines/unit_tests.yml
+++ b/.azure-pipelines/unit_tests.yml
@@ -18,9 +18,6 @@ jobs:
       Py27:
         PYTHON_VERSION: "2.7"
         TOXENV: "py27"
-      Py36:
-        PYTHON_VERSION: "3.6"
-        TOXENV: "py36"
       Py37:
         PYTHON_VERSION: "3.7"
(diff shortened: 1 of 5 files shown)


ofek:
CI just looks flaky https://pypi.org/project/tomli/#history

ofek:
AZP has old PyPy 3.6

jirikuncar (approved):

skarimo (approved):

PR #729, opened 2022-07-11
[base] demote noisy log lines to debug level

yashsarma:
### What does this PR do?

We currently use datadogpy library v0.44.0 and noticed tons of log volume coming from the base class file especially in the flush thread flow, causing log volume spike and masking useful information that would otherwise be gleaned from our logs. This PR downgrades the log level to debug, these lines serve little purpose in my humble opinion and could easily be replaced with a counter.

What inspired you to submit this pull request?

This change isn't a feature or a bug fix, but follows a process of better log hygiene. 

### Description of the Change

please refer to the questions above. 

### Alternate Designs

Would be replace these log lines with simple statsd counter incr.

### Possible Drawbacks

None

### Verification Process

What process did you follow to verify that your change has the desired effects?

- How did you verify that all new functionality works as expected?
- How did you verify that all changed functionality works as expected?
- How did you verify that the change has not introduced any regressions?

Describe the actions you performed (including scripts, commands you ran, etc.), and describe the results you observed.

N/A

### Additional Notes

N/A

### Release Notes

Reduce log noise caused by repeated info logs in the statsd buffer flush flow. 

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -358,7 +358,7 @@ class DogStatsd(object):
         self._disable_buffering = disable_buffering
         if self._disable_buffering:
             self._send = self._send_to_server
-            log.info("Statsd buffering is disabled")
+            log.debug("Statsd buffering is disabled")
 
         # Start the flush thread if buffering is enabled and the interval is above
         # a reasonable range. This both prevents thrashing and allow us to use "0.0"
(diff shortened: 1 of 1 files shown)


tiwilliam:
I find these lines annoying. These were added when buffering was made default in #670, and was intended to signal when you had explicitly turned it off. Now buffering is disabled by default, and these will always be printed on import, even if you create your own instance and enabling buffering, making the log line confusing since you might actually be using a buffered instance.

sgnn7:
@tiwilliam

> was intended to signal when you had explicitly turned it off

It was added for a display on startup of the current mode (since it was inteded to be changed) and on any explicit toggling of the setting but I do see your point since the default was never switched in the final releases.

------

@tiwilliam  / @yashsarma: 

Given that this may have a bit bigger impact than it was anticipated originally or implied from the initial issue description we can merge this change for now. I think if/when buffering is enabled by default again as an option we may need to revisit this decision though.

sgnn7 (approved):

Issue #725, opened 2022-06-27
statsd.event should raise something more specific than Exception


tylertrussell:
**Is your feature request related to a problem? Please describe.**
The `statsd.event` function will `raise Exception` in the event the payload is invalid (i.e. too big).  This prevents users of the library from catching a specific exception, and could lead to other types of errors (e.g. `TypeError`) being "swallowed" when they should be allowed to propagate.

**Describe the solution you'd like**
Instead of `Exception`, the library should raise a `ValueError` (or some other more-specific `Exception` subclass).

**Describe alternatives you've considered**
Users of the library could catch an `Exception` and then try to compare the error message, which is not great.

Updating to remove stale tag; this is still reproducible.

tylertrussell:
I confirm that I can still reproduce this issue in the latest version of the project.

There is an open PR #730 that addresses this issue, it would be excellent to have that reviewed by the relevant parties at some point soon.

sgnn7:
@tylertrussell The [linked PR](https://github.com/DataDog/datadogpy/pull/730) has been merged so it should be out with the next release of this library (probably `0.45.x`). I'll mark this issue as closed but feel free to reopen if needed.

PR #730, opened 2022-07-13
[statsd] Raise ValueError instead of Exception when payload is too large

mlanicaputo:
### What does this PR do?
As noted in #725, statsd.event() raises a general Exception in the event that an event payload is too long. This can lead to error swallowing. This PR changes this error block to raise a ValueError instead.

### Description of the Change
The statsd.event() method throws a different Exception (ValueError) in the event that a payload is too large. A test was added to test_statsd.py in order to ensure that the proper error is raised when appropriate.

### Alternate Designs
Any exception could theoretically be raised, but ValueError seemed the most appropriate according to [Python's guidelines](https://docs.python.org/3/library/exceptions.html#TypeError). Alternatively, users of the datadog library could just catch the Exception but that is poor practice.

### Possible Drawbacks
Users of the library who currently have code in place to catch a general Exception may have to change their libraries to catch the more specific ValueError.

### Verification Process
A new test was added to test_statsd.py in order to ensure the viability of the error change. It tests that a payload that is too large raises the ValueError, and that a payload of allowable length does not raise a ValueError. This can be double checked by running the following from the datadogpy directory: `pytest tests/unit/dogstatsd/test_statsd.py -k test_event_payload_error`.

### Additional Notes
Thank you for being open source!

### DIFF
--- a/tests/unit/dogstatsd/test_statsd.py
+++ b/tests/unit/dogstatsd/test_statsd.py
@@ -488,6 +488,23 @@ class TestDogStatsd(unittest.TestCase):
             ),
         )
 
+    def test_event_payload_error(self):
+        def func(fail: bool):
+            # define an event payload that is > 8 * 1024
+            message = ["l" for i in range(8 * 1024)]
+            message = "".join(message)
+            payload = {"title": "title", "message": message}
+
+            self.statsd.event(**payload)
+
+        # check that the method fails when the payload is too large
+        with pytest.raises(TypeError):
+            func()
+
+        # check that the method does not fail with a small payload
+        self.statsd.event("title", "message")
+
+
     def test_service_check(self):
         now = int(time.time())
         self.statsd.service_check(
--- a/tests/unit/dogstatsd/test_statsd.py
+++ b/tests/unit/dogstatsd/test_statsd.py
@@ -489,7 +489,7 @@ class TestDogStatsd(unittest.TestCase):
         )
 
     def test_event_payload_error(self):
-        def func(fail: bool):
+        def func():
             # define an event payload that is > 8 * 1024
             message = ["l" for i in range(8 * 1024)]
             message = "".join(message)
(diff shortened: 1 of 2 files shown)


tests/unit/dogstatsd/test_statsd.py
```diff
@@ -488,6 +488,23 @@ def test_event_constant_tags(self):
...
+            # define an event payload that is > 8 * 1024
+            message = ["l" for i in range(8 * 1024)]
+            message = "".join(message)
+            payload = {"title": "title", "message": message}
+
+            self.statsd.event(**payload)
+
+        # check that the method fails when the payload is too large
+        with pytest.raises(ValueError):
+            func()
```
sgnn7, line 502: Why not have the `func()` body inside the `with` here rather than defining it separately from invoking it?

sgnn7 (changes requested):
Hey @mlanicaputo! I left a comment on the PR but it's a minor nit

tests/unit/dogstatsd/test_statsd.py
mlanicaputo: I used the `with` statement to match the convention with which `pytest.raises()` is invoked in this script already: on [line 917](https://github.com/DataDog/datadogpy/pull/730/files/80e670ead0cecf2359144bcb78e707fae62701fa#diff-57154b2052d2e9499c5f655cf922894cfa70ceffedbcdfcb326ec52b97cd0cf3L917) and [line 937](https://github.com/DataDog/datadogpy/pull/730/files/80e670ead0cecf2359144bcb78e707fae62701fa#diff-57154b2052d2e9499c5f655cf922894cfa70ceffedbcdfcb326ec52b97cd0cf3L937
).

sgnn7 (approved):

sgnn7:
@mlanicaputo Merged - thanks for your contribution!

PR #787, opened 2023-07-19
Add blocking socket mode and background sender

vickenty:
### What does this PR do?

Add options to improve dogstatsd packet deliverability.

### Description of the Change

Allow dogstatsd socket to be used in a blocking mode with a timeout. This improves our chances to send the packet if the socket queue is full (e.g. after a burst in packets).

Add a background thread to handle the writes, with a queue to act as a buffer.

Together, this allows users to reduce dropped packets to zero while maintaining low latency for the user application, at the cost of higher cpu and memory usage.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -110,6 +110,8 @@ class DogStatsd(object):
         max_buffer_len=0,                       # type: int
         container_id=None,                      # type: Optional[Text]
         origin_detection_enabled=True,          # type: bool
+        socket_timeout=0,                       # type: Optional[float]
+        telemetry_socket_timeout=0,             # type: Optional[float]
     ):  # type: (...) -> None
         """
         Initialize a DogStatsd object.
@@ -239,6 +241,17 @@ class DogStatsd(object):
         Default: True.
         More on this: https://docs.datadoghq.com/developers/dogstatsd/?tab=kubernetes#origin-detection-over-udp
         :type origin_detection_enabled: boolean
+
+        :param socket_timeout: Set timeout for socket operations, in seconds. Optional.
+        If sets to zero, never wait if operation can not be completed immediately. If set to None, wait forever.
+        This option does not affect hostname resolution when using UDP.
+        :type socket_timeout: float
+
+        :param telemetry_socket_timeout: Set timeout for the telemetry socket operations. Optional.
+        Effective only if either telemetry_host or telemetry_socket_path are set.
+        If sets to zero, never wait if operation can not be completed immediately. If set to None, wait forever.
+        This option does not affect hostname resolution when using UDP.
+        :type socket_timeout: float
         """
 
         self._socket_lock = Lock()
@@ -276,6 +289,7 @@ class DogStatsd(object):
 
         # Connection
         self._max_payload_size = max_buffer_len
+        self.socket_timeout = socket_timeout
         if socket_path is not None:
             self.socket_path = socket_path  # type: Optional[text]
             self.host = None
@@ -294,6 +308,7 @@ class DogStatsd(object):
         self.telemetry_socket_path = telemetry_socket_path
         self.telemetry_host = None
         self.telemetry_port = None
+        self.telemetry_socket_timeout = telemetry_socket_timeout
         if not telemetry_socket_path and telemetry_host:
             self.telemetry_socket_path = None
             self.telemetry_host = self.resolve_host(telemetry_host, use_default_route)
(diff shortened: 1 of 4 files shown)


therve (dismissed):

datadog/dogstatsd/base.py
```diff
@@ -239,6 +254,34 @@ def __init__(
...
+        :param telemetry_socket_timeout: Set timeout for the telemetry socket operations. Optional.
+        Effective only if either telemetry_host or telemetry_socket_path are set.
+        If sets to zero, never wait if operation can not be completed immediately. If set to None, wait forever.
+        This option does not affect hostname resolution when using UDP.
+        :type socket_timeout: float
+
+        :param disable_sender: Use a background thread to communicate with the dogstatsd server. Optional.
+        When enabled, a background thread will be used to send metric payloads to the Agent.
+        Default: True.
+        :type disable_sender: boolean
```
remeh, line 272: Would that make sense to have something like `disable_background_sender` or similar, explicitly displaying that the mentioned sender is running in the background/in a thread, not just the normal one?
This could also be thought the other way around: `enable_sender_thread` or similar.

It might be over-thinking it, please discard if you think so.

datadog/dogstatsd/base.py
```diff
@@ -239,6 +254,34 @@ def __init__(
...
+        :param socket_timeout: Set timeout for socket operations, in seconds. Optional.
+        If sets to zero, never wait if operation can not be completed immediately. If set to None, wait forever.
+        This option does not affect hostname resolution when using UDP.
+        :type socket_timeout: float
+
+        :param telemetry_socket_timeout: Set timeout for the telemetry socket operations. Optional.
+        Effective only if either telemetry_host or telemetry_socket_path are set.
+        If sets to zero, never wait if operation can not be completed immediately. If set to None, wait forever.
+        This option does not affect hostname resolution when using UDP.
+        :type socket_timeout: float
```
remeh, line 267: I think that's a typo.

```suggestion
        :type telemetry_socket_timeout: float
```

datadog/dogstatsd/base.py
```diff
@@ -239,6 +254,34 @@ def __init__(
...
+
+        :param disable_sender: Use a background thread to communicate with the dogstatsd server. Optional.
+        When enabled, a background thread will be used to send metric payloads to the Agent.
+        Default: True.
+        :type disable_sender: boolean
+
+        :param sender_queue_size: Set the maximum number of packets to queue for the sender. Optional
+        How may packets to queue before blocking or dropping the packet if the packet queue is already full.
+        Default: 0 (unlimited).
+        :type sender_queue_timeout: integer
```
remeh, line 277:
```suggestion
        :type sender_queue_size: integer
```

datadog/dogstatsd/base.py
```diff
@@ -1066,5 +1148,28 @@ def _set_container_id(self, container_id, origin_detection_enabled):
...
+        )
+        self._sender_thread.daemon = True
+        self._sender_thread.start()
+
+    def _sender_main_loop(self):
+        while True:
+            self._xmit_packet_with_telemetry(self._queue.get())
+            self._queue.task_done()
+
+    def wait_for_pending(self):
```
remeh, line 1165: Should we consider documenting the new need (when using the threaded sender) to use this method to make sure everything's flushed properly and no buffered metrics are lost on shutdown?

remeh (approved):
Made a few polishing comments but that's a neat implementation.

datadog/dogstatsd/base.py
```diff
@@ -834,7 +904,19 @@ def _is_telemetry_flush_time(self):
             self._last_flush_time + self._telemetry_flush_interval < time.time()
 
     def _send_to_server(self, packet):
-        self._xmit_packet(packet + '\n', False)
+        if self._queue is not None:
+            try:
+                self._queue.put(packet + '\n', self._queue_blocking, self._queue_timeout)
```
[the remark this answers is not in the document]
vickenty: `_send_to_server` is also used in a bunch of places indirectly via the `_send` alias when buffering is disabled.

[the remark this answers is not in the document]
remeh: Indeed, but it doesn't stop us from having the call to `_send_to_server` in the flush method to add the `\n` as shown above. No strong opinion here, feel free to discard 👍

datadog/dogstatsd/base.py
vickenty: Good idea, done

remeh (approved):

therve (approved):

PR #791, opened 2023-09-04
Add ipv6 support for dogstatsd

vickenty:
### What does this PR do?

Add IPv6 support for hostnames and literal addresses.

### Description of the Change

Call `getaddrinfo` to parse and resolve an internet address before creating the socket. Try in a loop until one address succeeds.

For backwards compatibility, prefer ipv4 addresses: ipv4-only agent should still work even if hostname (e.g. `localhost`) resolves to both ipv4 and ipv6 address and ipv6 is preferred by the OS. This also matches behavior of Golang.

### Possible Drawbacks

Address order specified in `gai.conf` is ignored, which may lead to inconsistencies with other applications.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -635,8 +635,8 @@ class DogStatsd(object):
         log.debug("Connecting to %s:%s", host, port)
         addrinfo = socket.getaddrinfo(host, port, type=socket.SOCK_DGRAM)
         # Override gai.conf order for backwrads compatibility: prefer
-        # v4, so that v4-only service on hosts with both addresses
-        # still work.
+        # v4, so that a v4-only service on hosts with both addresses
+        # still works.
         addrinfo.sort(key=lambda v: v[0] == socket.AF_INET, reverse=True)
         lastaddr = len(addrinfo) - 1
         for i, (af, ty, proto, _, addr) in enumerate(addrinfo):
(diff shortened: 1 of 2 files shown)


datadog/dogstatsd/base.py
```diff
@@ -625,13 +625,37 @@ def _get_uds_socket(cls, socket_path, timeout):
...
+        # Override gai.conf order for backwrads compatibility: prefer
+        # v4, so that a v4-only service on hosts with both addresses
+        # still works.
+        addrinfo.sort(key=lambda v: v[0] == socket.AF_INET, reverse=True)
+        lastaddr = len(addrinfo) - 1
+        for i, (af, ty, proto, _, addr) in enumerate(addrinfo):
+            sock = None
+            try:
+                sock = socket.socket(af, ty, proto)
+                sock.connect(addr)
```
remeh, line 646: Since your change, are you sure it's OK to call the `settimeout` after the call to `connect`? Could be for UDP but might be worth preserving the old behavior: https://docs.python.org/3/library/socket.html#timeouts-and-the-connect-method

remeh (dismissed):
Left a small remark, approved for a future merge as the rest looks good to me!

remeh (approved):

datadog/dogstatsd/base.py
vickenty: I'm pretty sure socket timeout doesn't make any difference for udp sockets, but just for symmetry I updated the code to have the order same between udp and uds.

PR #813, opened 2024-01-16
Support inode resolution mechanism for Origin Detection

wdhif:

### What does this PR do?

Implements the support of Origin Detection when the container ID is unavailable from inside the container by using the inode resolution mechanic. The original Go implementation is here: [https://github.com/DataDog/datadog-go/pull/291](https://github.com/DataDog/datadog-go/pull/291)

### Description of the Change

If we are not able to find the container ID from inside the container, we will send instead the cgroup inode to be able to resolve the container ID on the Agent side.

### Verification Process

Use the following deployment on a k8s cluster running `cgroupv2`:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: dummy-dsd-python-app
spec:
  replicas: 1
  selector:
    matchLabels:
      app: dummy-dsd-python-app
  template:
    metadata:
      labels:
        app: dummy-dsd-python-app
        admission.datadoghq.com/enabled: "false"
    spec:
      containers:
      - name: dummy-dsd-python-app
        image: wdhifdatadog/dummy-dsd-python-app
        imagePullPolicy: Always
        env:
        - name: DD_AGENT_HOST
          valueFrom:
            fieldRef:
              fieldPath: status.hostIP
```

You should then be able to find a `dummy.dsd.python.metric` with a `container_id` tag.

### DIFF
--- a/datadog/dogstatsd/container.py
+++ b/datadog/dogstatsd/container.py
@@ -94,12 +94,13 @@ class Cgroup(object):
             self.CGROUPV1_BASE_CONTROLLER,
             self.CGROUPV2_BASE_CONTROLLER,
         ]:
-            if controller, cpath in cgroup_controllers_paths.values():
+            if controller in cgroup_controllers_paths:
                 inode_path = os.path.join(
-                    self.DEFAULT_CGROUP_MOUNT_PATH,
+                    self.CGROUP_MOUNT_PATH,
                     controller,
-                    cpath if cpath != "/" else "",
+                    cgroup_controllers_paths[controller] if cgroup_controllers_paths[controller] != "/" else "",
                 )
-                return "in-{0}".format(os.stat(inode_path).st_ino)
+                inode = os.stat(inode_path).st_ino
+                return "in-{0}".format(inode) if int(inode) > 2 else None
 
         return None
--- a/datadog/dogstatsd/container.py
+++ b/datadog/dogstatsd/container.py
@@ -86,7 +86,7 @@ class Cgroup(object):
                 tokens = line.strip().split(":")
                 if len(tokens) != 3:
                     continue
-                if tokens[1] == self.CGROUPV1_BASE_CONTROLLER or tokens[1] == "":
+                if tokens[1] == self.CGROUPV1_BASE_CONTROLLER or tokens[1] == self.CGROUPV2_BASE_CONTROLLER:
                     cgroup_controllers_paths[tokens[1]] = tokens[2]
 
         # Retrieve the cgroup inode from "/sys/fs/cgroup + controller + cgroupNodePath"
--- a/datadog/dogstatsd/container.py
+++ b/datadog/dogstatsd/container.py
@@ -42,12 +42,12 @@ class Cgroup(object):
     CONTAINER_RE = re.compile(r"(?:.+)?({0}|{1}|{2})(?:\.scope)?$".format(UUID_SOURCE, CONTAINER_SOURCE, TASK_SOURCE))
 
     def __init__(self):
-        if self._is_cgroup_namespace():
+        if self._is_host_cgroup_namespace():
             self.container_id = self._read_cgroup_path()
             return
         self.container_id = self._get_cgroup_from_inode()
 
-    def _is_cgroup_namespace(self):
+    def _is_host_cgroup_namespace(self):
         """Check if the current process is in a host cgroup namespace."""
         return (
             os.stat(self.CGROUP_NS_PATH).st_ino == self.HOST_CGROUP_NAMESPACE_INODE
@@ -101,6 +101,7 @@ class Cgroup(object):
                     cgroup_controllers_paths[controller] if cgroup_controllers_paths[controller] != "/" else "",
                 )
                 inode = os.stat(inode_path).st_ino
+                # 0 is not a valid inode. 1 is a bad block inode and 2 is the root of a filesystem.
                 return "in-{0}".format(inode) if int(inode) > 2 else None
 
         return None
--- a/datadog/dogstatsd/container.py
+++ b/datadog/dogstatsd/container.py
@@ -49,11 +49,14 @@ class Cgroup(object):
 
     def _is_host_cgroup_namespace(self):
         """Check if the current process is in a host cgroup namespace."""
-        return (
-            os.stat(self.CGROUP_NS_PATH).st_ino == self.HOST_CGROUP_NAMESPACE_INODE
-            if os.path.exists(self.CGROUP_NS_PATH)
-            else False
-        )
+        try:
+            return (
+                os.stat(self.CGROUP_NS_PATH).st_ino == self.HOST_CGROUP_NAMESPACE_INODE
+                if os.path.exists(self.CGROUP_NS_PATH)
+                else False
+            )
+        except Exception:
+            return False
 
     def _read_cgroup_path(self):
         """Read the container ID from the cgroup file."""
--- a/datadog/dogstatsd/container.py
+++ b/datadog/dogstatsd/container.py
@@ -4,6 +4,7 @@
 # Copyright 2015-Present Datadog, Inc
 
 import errno
+import os
 import re
 
 
@@ -13,12 +14,14 @@ class UnresolvableContainerID(Exception):
     """
 
 
-class ContainerID(object):
+class Cgroup(object):
     """
-    A reader class that retrieves the current container ID parsed from a the cgroup file.
+    A reader class that retrieves either:
+    - The current container ID parsed from the cgroup file
+    - The cgroup controller inode.
 
     Returns:
-    object: ContainerID
+    object: Cgroup
 
     Raises:
         `NotImplementedError`: No proc filesystem is found (non-Linux systems)
@@ -26,6 +29,12 @@ class ContainerID(object):
     """
 
     CGROUP_PATH = "/proc/self/cgroup"
+    DEFAULT_CGROUP_MOUNT_PATH = "/sys/fs/cgroup"  # default cgroup mount path.
+    CGROUP_NS_PATH = "/proc/self/ns/cgroup"  # path to the cgroup namespace file.
+    CGROUPV1_BASE_CONTROLLER = "memory"  # controller used to identify the container-id in cgroup v1 (memory).
+    CGROUPV2_BASE_CONTROLLER = ""  # controller used to identify the container-id in cgroup v2.
+    HOST_CGROUP_NAMESPACE_INODE = "0xEFFFFFFBL"  # inode of the host cgroup namespace.
+
     UUID_SOURCE = r"[0-9a-f]{8}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{4}[-_][0-9a-f]{12}"
     CONTAINER_SOURCE = r"[0-9a-f]{64}"
     TASK_SOURCE = r"[0-9a-f]{32}-\d+"
@@ -33,11 +42,23 @@ class ContainerID(object):
     CONTAINER_RE = re.compile(r"(?:.+)?({0}|{1}|{2})(?:\.scope)?$".format(UUID_SOURCE, CONTAINER_SOURCE, TASK_SOURCE))
 
     def __init__(self):
-        self.container_id = self._read_container_id(self.CGROUP_PATH)
+        if self._is_cgroup_namespace():
+            self.container_id = self._read_cgroup_path()
+            return
+        self.container_id = self._get_cgroup_from_inode()
 
-    def _read_container_id(self, fpath):
+    def _is_cgroup_namespace(self):
+        """Check if the current process is in a host cgroup namespace."""
+        return (
+            os.stat(self.CGROUP_NS_PATH).st_ino == self.HOST_CGROUP_NAMESPACE_INODE
+            if os.path.exists(self.CGROUP_NS_PATH)
+            else False
+        )
+
+    def _read_cgroup_path(self):
+        """Read the container ID from the cgroup file."""
         try:
-            with open(fpath, mode="r") as fp:
+            with open(self.CGROUP_PATH, mode="r") as fp:
                 for line in fp:
                     line = line.strip()
                     match = self.LINE_RE.match(line)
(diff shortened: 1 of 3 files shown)


datadog/dogstatsd/container.py
```diff
@@ -55,3 +76,30 @@ def _read_container_id(self, fpath):
...
+    def _get_cgroup_from_inode(self):
+        """Read the container ID from the cgroup inode."""
+        # Parse /proc/self/cgroup and get a map of controller to its associated cgroup node path.
+        cgroup_controllers_paths = {}
+        with open(self.CGROUP_PATH, mode="r") as fp:
+            for line in fp:
+                tokens = line.strip().split(":")
+                if len(tokens) != 3:
+                    continue
+                if tokens[1] == self.CGROUPV1_BASE_CONTROLLER or tokens[1] == "":
```
vickenty, line 89:
```suggestion
                if tokens[1] == self.CGROUPV1_BASE_CONTROLLER or tokens[1] == self.CGROUPV2_BASE_CONTROLLER:
```

datadog/dogstatsd/container.py
```diff
@@ -55,3 +76,30 @@ def _read_container_id(self, fpath):
...
+            self.CGROUPV1_BASE_CONTROLLER,
+            self.CGROUPV2_BASE_CONTROLLER,
+        ]:
+            if controller in cgroup_controllers_paths:
+                inode_path = os.path.join(
+                    self.DEFAULT_CGROUP_MOUNT_PATH,
+                    controller,
+                    cgroup_controllers_paths[controller] if cgroup_controllers_paths[controller] != "/" else "",
+                )
+                return "in-{0}".format(os.stat(inode_path).st_ino)
```
vickenty, line 103: Does this need to check the inode number? https://github.com/DataDog/java-dogstatsd-client/pull/232/files#diff-79768f0f7e1c9370c006cfd086b4ae68cdc0995de5cf385a40d2d5312dddecbbR176

datadog/dogstatsd/container.py
```diff
@@ -55,3 +76,30 @@ def _read_container_id(self, fpath):
...
+                    continue
+                if tokens[1] == self.CGROUPV1_BASE_CONTROLLER or tokens[1] == "":
+                    cgroup_controllers_paths[tokens[1]] = tokens[2]
+
+        # Retrieve the cgroup inode from "/sys/fs/cgroup + controller + cgroupNodePath"
+        for controller in [
+            self.CGROUPV1_BASE_CONTROLLER,
+            self.CGROUPV2_BASE_CONTROLLER,
+        ]:
+            if controller in cgroup_controllers_paths:
```
vickenty, line 97:
```suggestion
            if controller, cpath in cgroup_controllers_paths.values():
```

datadog/dogstatsd/container.py
```diff
@@ -55,3 +76,30 @@ def _read_container_id(self, fpath):
...
+                inode_path = os.path.join(
+                    self.DEFAULT_CGROUP_MOUNT_PATH,
+                    controller,
+                    cgroup_controllers_paths[controller] if cgroup_controllers_paths[controller] != "/" else "",
```
vickenty, line 101:
```suggestion
                    cpath if cpath != "/" else "",
```

datadog/dogstatsd/container.py
```diff
@@ -13,31 +14,51 @@ class UnresolvableContainerID(Exception):
...
-    object: ContainerID
+    object: Cgroup
 
     Raises:
         `NotImplementedError`: No proc filesystem is found (non-Linux systems)
         `UnresolvableContainerID`: Unable to read the container ID
     """
 
     CGROUP_PATH = "/proc/self/cgroup"
+    DEFAULT_CGROUP_MOUNT_PATH = "/sys/fs/cgroup"  # default cgroup mount path.
```
vickenty, line 32: Since it isn't customizable, there is no point calling it the "default".

```suggestion
    CGROUP_MOUNT_PATH = "/sys/fs/cgroup"  # cgroup mount path.
```

datadog/dogstatsd/container.py
wdhif: Oops, my bad 👍 applied!

datadog/dogstatsd/container.py
wdhif: Actually wait, this is a `if`, not a `for`. This cannot work right?

datadog/dogstatsd/container.py
wdhif: Yes it would be better, I've added a commit for this 👍

datadog/dogstatsd/container.py
wdhif: I've removed this.

datadog/dogstatsd/container.py
wdhif: Same here

AliDatadog (approved):

datadog/dogstatsd/container.py
```diff
@@ -13,31 +14,51 @@ class UnresolvableContainerID(Exception):
...
     CONTAINER_RE = re.compile(r"(?:.+)?({0}|{1}|{2})(?:\.scope)?$".format(UUID_SOURCE, CONTAINER_SOURCE, TASK_SOURCE))
 
     def __init__(self):
-        self.container_id = self._read_container_id(self.CGROUP_PATH)
+        if self._is_cgroup_namespace():
+            self.container_id = self._read_cgroup_path()
+            return
+        self.container_id = self._get_cgroup_from_inode()
+
+    def _is_cgroup_namespace(self):
```
AliDatadog, line 50:
```suggestion
    def _is_host_cgroup_namespace(self):
```

datadog/dogstatsd/container.py
```diff
@@ -13,31 +14,51 @@ class UnresolvableContainerID(Exception):
...
+        """Check if the current process is in a host cgroup namespace."""
+        return (
+            os.stat(self.CGROUP_NS_PATH).st_ino == self.HOST_CGROUP_NAMESPACE_INODE
```
AliDatadog, line 53: Could there be a case where this raises an exception and we would want to pursue with getting the cgroup controller inode ?

wdhif: I don't think we have any specific exception to catch here to resolve it. If it does fail, it will be catched here: https://github.com/DataDog/datadogpy/blob/master/datadog/dogstatsd/base.py#L1290C17-L1290C17

Maybe we can get a confirmation about this from @vickenty?

datadog/dogstatsd/container.py
wdhif: Done 👍

datadog/dogstatsd/container.py
```diff
@@ -13,31 +14,51 @@ class UnresolvableContainerID(Exception):
...
         `NotImplementedError`: No proc filesystem is found (non-Linux systems)
         `UnresolvableContainerID`: Unable to read the container ID
     """
 
     CGROUP_PATH = "/proc/self/cgroup"
+    CGROUP_MOUNT_PATH = "/sys/fs/cgroup"  # cgroup mount path.
+    CGROUP_NS_PATH = "/proc/self/ns/cgroup"  # path to the cgroup namespace file.
+    CGROUPV1_BASE_CONTROLLER = "memory"  # controller used to identify the container-id in cgroup v1 (memory).
+    CGROUPV2_BASE_CONTROLLER = ""  # controller used to identify the container-id in cgroup v2.
+    HOST_CGROUP_NAMESPACE_INODE = "0xEFFFFFFBL"  # inode of the host cgroup namespace.
```
vickenty, line 36: Should this be a string? stat().st_ino is an integer.

datadog/dogstatsd/container.py
vickenty: We probably want the logic to be exactly the same between all clients, typically Go is the "reference" implementation. It seems Go treats errors in `isHostCgroupNamespace` as "false", and skips to inode code. Let's do the same here.

tests/unit/dogstatsd/test_container.py
```diff
@@ -125,12 +125,22 @@ def get_mock_open(read_data=None):
...
+        with mock.patch("os.stat", mock.MagicMock(return_value=mock.Mock(st_ino="0xEFFFFFFBL"))):
+            reader = Cgroup()
         assert expected_container_id == reader.container_id
 
         mock_open.assert_called_once_with("/proc/self/cgroup", mode="r")
+
+
+def test_container_id_inode():
+    """Test that the inode is returned when the container ID cannot be found."""
+    with mock.patch("datadog.dogstatsd.container.open", mock.mock_open(read_data="0::/")) as mock_open:
```
vickenty, line 142: Can we have a test for cgroupv1 and unified cgroup formats?

datadog/dogstatsd/container.py
wdhif: I've added a `try/except`
```diff
diff --git a/datadog/dogstatsd/container.py b/datadog/dogstatsd/container.py
index fc911cf..94b5be2 100644
--- a/datadog/dogstatsd/container.py
+++ b/datadog/dogstatsd/container.py
@@ -49,11 +49,14 @@ class Cgroup(object):

     def _is_host_cgroup_namespace(self):
         """Check if the current process is in a host cgroup namespace."""
-        return (
-            os.stat(self.CGROUP_NS_PATH).st_ino == self.HOST_CGROUP_NAMESPACE_INODE
-            if os.path.exists(self.CGROUP_NS_PATH)
-            else False
-        )
+        try:
+            return (
+                os.stat(self.CGROUP_NS_PATH).st_ino == self.HOST_CGROUP_NAMESPACE_INODE
+                if os.path.exists(self.CGROUP_NS_PATH)
+                else False
+            )
+        except Exception:
+            return False

     def _read_cgroup_path(self):
         """Read the container ID from the cgroup file."""
```
There is no logging here though.

tests/unit/dogstatsd/test_container.py
wdhif: Added 👍

datadog/dogstatsd/container.py
wdhif: Done 👍

tests/unit/dogstatsd/test_container.py
```diff
@@ -125,12 +127,57 @@ def get_mock_open(read_data=None):
...
+2:blkio:/
+1:name=systemd:/
+0::/
+"""
+
+    def inode_stat_mock(path):
+        global MEMORY_PRIORITY_VALID
+        if not MEMORY_PRIORITY_VALID:
+            # Priority is given to the cgroupv1 controller.
+            assert path == "/sys/fs/cgroup/memory/"
```
vickenty, line 170: I may be misreading things again, but wouldn't this always fail? The first stat is on `CGROUP_NS_PATH`.

wdhif: Yes I think it's because `pytest` actually just want an `assert` to be validated at some point in a run. I found another way to do this:
```
tests/unit/dogstatsd/test_container.py::test_container_id_from_cgroup[\n11:hugetlb:/ecs/55091c13-b8cf-4801-b527-f4601742204d/432624d2150diff --git a/tests/unit/dogstatsd/test_container.py b/tests/unit/dogstatsd/test_container.py
index cf259ae..f8da40e 100644
--- a/tests/unit/dogstatsd/test_container.py
+++ b/tests/unit/dogstatsd/test_container.py
@@ -12,7 +12,7 @@ import pytest

 from datadog.dogstatsd.container import Cgroup

-MEMORY_PRIORITY_VALID = False
+MEMORY_PRIORITY_VALID = []


 def get_mock_open(read_data=None):
@@ -165,10 +165,7 @@ def test_container_id_inode():

     def inode_stat_mock(path):
         global MEMORY_PRIORITY_VALID
-        if not MEMORY_PRIORITY_VALID:
-            # Priority is given to the cgroupv1 controller.
-            assert path == "/sys/fs/cgroup/memory/"
-            MEMORY_PRIORITY_VALID = True
+        MEMORY_PRIORITY_VALID.append(path)

         # The cgroupv1 controller is mounted on inode 0. This will cause a fallback to the cgroupv2 controller.
         if path == "/sys/fs/cgroup/memory/":
@@ -180,4 +177,9 @@ def test_container_id_inode():
         with mock.patch("os.stat", mock.MagicMock(side_effect=inode_stat_mock)):
             reader = Cgroup()
         assert reader.container_id == "in-1234"
+        global MEMORY_PRIORITY_VALID
+        assert MEMORY_PRIORITY_VALID[-2:] == [
+            "/sys/fs/cgroup/memory/",
+            "/sys/fs/cgroup/"
+        ]
         mock_open.assert_called_once_with("/proc/self/cgroup", mode="r")
```

Note that `MEMORY_PRIORITY_VALID` will contain at first two `CGROUP_NS_PATH`. One for the first `os.stat()`, the other one for the `os.path.exist()` since it's actually also using an `os.stat()`.

Good catch 👍

tests/unit/dogstatsd/test_container.py
```diff
@@ -125,12 +127,59 @@ def get_mock_open(read_data=None):
...
+
+    def inode_stat_mock(path):
+        global MEMORY_PRIORITY_VALID
+        MEMORY_PRIORITY_VALID.append(path)
+
+        # The cgroupv1 controller is mounted on inode 0. This will cause a fallback to the cgroupv2 controller.
+        if path == "/sys/fs/cgroup/memory/":
+            return mock.Mock(st_ino=0)
+        elif path == "/sys/fs/cgroup/":
+            return mock.Mock(st_ino=1234)
```
vickenty, line 174: Would it make sense to return a valid number for CGROUP_NS_PATH too?

tests/unit/dogstatsd/test_container.py
```diff
@@ -125,12 +127,59 @@ def get_mock_open(read_data=None):
...
+4:freezer:/
+3:perf_event:/
+2:blkio:/
+1:name=systemd:/
+0::/
+"""
+
+    def inode_stat_mock(path):
+        global MEMORY_PRIORITY_VALID
+        MEMORY_PRIORITY_VALID.append(path)
```
vickenty, line 168: If you use append, `MEMORY_PRIORITY_VALID` doesn't need to be global, it can be a regular variable in the test function. Also, could you please choose a more descriptive name for it, perhaps `paths_checked` or something like that.

tests/unit/dogstatsd/test_container.py
wdhif: No I don't think so since this case is in fact tested by `test_container_id_from_cgroup()`.

tests/unit/dogstatsd/test_container.py
wdhif: Sure, done 👍 
```diff
diff --git a/tests/unit/dogstatsd/test_container.py b/tests/unit/dogstatsd/test_container.py
index f8da40e..4824754 100644
--- a/tests/unit/dogstatsd/test_container.py
+++ b/tests/unit/dogstatsd/test_container.py
@@ -12,8 +12,6 @@ import pytest

 from datadog.dogstatsd.container import Cgroup

-MEMORY_PRIORITY_VALID = []
-

 def get_mock_open(read_data=None):
     mock_open = mock.mock_open(read_data=read_data)
@@ -163,9 +161,10 @@ def test_container_id_inode():
 0::/
 """

+    paths_checked = []
+
     def inode_stat_mock(path):
-        global MEMORY_PRIORITY_VALID
-        MEMORY_PRIORITY_VALID.append(path)
+        paths_checked.append(path)

         # The cgroupv1 controller is mounted on inode 0. This will cause a fallback to the cgroupv2 controller.
         if path == "/sys/fs/cgroup/memory/":
@@ -177,8 +176,7 @@ def test_container_id_inode():
         with mock.patch("os.stat", mock.MagicMock(side_effect=inode_stat_mock)):
             reader = Cgroup()
         assert reader.container_id == "in-1234"
-        global MEMORY_PRIORITY_VALID
-        assert MEMORY_PRIORITY_VALID[-2:] == [
+        assert paths_checked[-2:] == [
             "/sys/fs/cgroup/memory/",
             "/sys/fs/cgroup/"
```

vickenty (approved):

PR #831, opened 2024-05-03
Support passing Unix timestamps to dogstatsd

carlosroman:
### What does this PR do?

Following [v1.3 protocol](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics#dogstatsd-protocol-v13), allow passing timestamps along with the metric, which can mitigate the load on the agent when emitting a lot of metrics in a short time span.

### Description of the Change

Adds `gauge_with_timestamp` and `count_with_timestamp` to allow for passing metrics with timestamps. Also added `count` method which other clients have so that our clients all have similar methods.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -775,13 +775,32 @@ class DogStatsd(object):
         """
         return self._report(metric, "g", value, tags, sample_rate, timestamp)
 
+    # Minimum Datadog Agent version: 7.40.0
+    def gauge_with_timestamp(
+        self,
+        metric,  # type: Text
+        value,  # type: float
+        timestamp,  # type: int
+        tags=None,  # type: Optional[List[str]]
+        sample_rate=None,  # type: Optional[float]
+    ):  # type(...) -> None
+        """u
+        Record the value of a gauge with a Unix timestamp (in seconds),
+        optionally setting a list of tags and a sample rate.
+
+        Minimum Datadog Agent version: 7.40.0
+
+        >>> statsd.gauge("users.online", 123, 1713804588)
+        >>> statsd.gauge("active.connections", 1001, 1713804588, tags=["protocol:http"])
+        """
+        return self._report(metric, "g", value, tags, sample_rate, timestamp)
+
     def increment(
         self,
         metric,  # type: Text
         value=1,  # type: float
         tags=None,  # type: Optional[List[str]]
         sample_rate=None,  # type: Optional[float]
-        timestamp=0,  # type:int
     ):  # type: (...) -> None
         """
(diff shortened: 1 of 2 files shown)


datadog/dogstatsd/base.py
```diff
@@ -978,9 +1036,16 @@ def _report(self, metric, metric_type, value, tags, sample_rate):
         if sample_rate != 1 and random() > sample_rate:
             return
 
+        # timestamps (protocol v1.3) only allowed on gauges and counts
+        allows_timestamp = metric_type == "g" or metric_type == "c"
+        if not allows_timestamp or timestamp < 0:
+            timestamp = 0
```
remeh, line 1042: Do you think that would be worth a warning? It's a developer error at this point (somebody is calling `_report` with a wrong `metric_type`)

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -328,6 +336,26 @@ def test_counter(self):
...
+        self.statsd._reset_telemetry()
+        self.statsd.count_with_timestamp("page.views", 11, timestamp=2121)
+        self.statsd.flush()
+        self.assert_equal_telemetry("page.views:11|c|T2121\n", self.recv(2))
+
+    def test_count_with_invalid_ts_should_be_ignored(self):
+        self.statsd.count_with_timestamp("page.views", 1, timestamp=-1066)
+        self.statsd.flush()
+        self.assert_equal_telemetry("page.views:1|c\n", self.recv(2))
+
```
remeh, line 358: Would that be worth a direct call to `_report` with a timestamp and a wrong metric type to make sure it sets the timestamp to `0`? Not sure.

remeh (approved):
LGTM!

tests/unit/dogstatsd/test_statsd.py
carlosroman: Yeah, I feel there should be one more test to make sure that if statement that checks metric types with timestamps is covered correctly.

datadog/dogstatsd/base.py
abolibibelot: Yes worth a test (as said in the other comment) but maybe we don't want to flood stdout with logs?

carlosroman: I added a test but not sure about adding a warning. In other SDKs we don't log a warning if invalid timestamp used. Here if they are trying to send a metric that doesn't support timestamps then they're accessing methods they shouldn't be in the first place.

abolibibelot (approved):

gyandeeps (approved):

gh123man (approved):

HantingZhang2 (approved):

PR #837, opened 2024-06-26
add metric object type

andrewqian2001datadog:
### What does this PR do?

This PR is for adding metric objects that will make aggregation easier by allowing us to have one Metric object instantiation for each [unique metric context](https://docs.google.com/document/d/1AoveWKoMQc65hYbvZ27uE7pQ_I5nsGfvv25EcNVPLvQ/edit#heading=h.no4x6fjugsh) that we can add samples to which will make aggregation easier. Existing behavior already exists in the [client side aggregation for GO ](https://github.com/DataDog/datadog-go/pull/139/files#diff-768b1bb89feaa2e5fedd4162f84c59c801610960af421ebd18e744d3f0adaa94)

### Description of the Change

Create metric object type (`MetricAgggregator`) that other metric objects (`CountMetric`, `GaugeMetric`, etc.) implement that will enforce the objects to implement functions `aggregate` and **potentially** `flush_unsafe`. This should also allow us to group the different metric types together in one list to be sent when we [flush the metrics](https://github.com/DataDog/datadog-go/blob/0b67271aa15f5531a429d7d4598c705afefb4a55/statsd/aggregator.go#L70).

### Possible Drawbacks

Some future metrics may not have the necessary fields to be of type `MetricAggregator`

image

The existing code for flushing metrics does NOT use object oriented programming, as a result they have duplicate code for each metric type. 
image

Maybe i'm missing the potential drawbacks to using object oriented programming (specifically inheritance)? 

**Update:** 
image

### Verification Process

Unit tests that verify the new metric objects match the existing behavior in the [client side aggregator for Go](url).

### Additional Notes

The[ python dogstatsd API](https://github.com/DataDog/datadogpy/blob/master/datadog/dogstatsd/base.py) will need to be refactored in the future to use these metric objects.

### Release Notes

N/A, no new behavior until the aggregator class is implemented

### DIFF
--- a/datadog/dogstatsd/metrics.py
+++ b/datadog/dogstatsd/metrics.py
@@ -1,4 +1,5 @@
-from threading import Lock
+# TODO: potentially add locks
+# from threading import Lock
 
 
 class MetricAggregator(object):
@@ -10,8 +11,9 @@ class MetricAggregator(object):
     def aggregate(self, value):
         raise NotImplementedError("Subclasses should implement this method.")
 
-    def flush_unsafe(self):
-        raise NotImplementedError("Subclasses should implement this method.")
+    # TODO: potentially add this function
+    # def flush_unsafe(self):
+    #     raise NotImplementedError("Subclasses should implement this method.")
 
 
 class CountMetric(MetricAggregator):
@@ -22,14 +24,15 @@ class CountMetric(MetricAggregator):
     def aggregate(self, v):
         self.value += v
 
-    def flush_unsafe(self):
-        return {
-            "metric_type": "count",
-            "name": self.name,
-            "tags": self.tags,
-            "rate": self.rate,
-            "ivalue": self.value,
-        }
+    # TODO: potentially add this function
+    # def flush_unsafe(self):
+    #     return {
+    #         "metric_type": "count",
+    #         "name": self.name,
+    #         "tags": self.tags,
+    #         "rate": self.rate,
+    #         "ivalue": self.value,
+    #     }
 
 
 class GaugeMetric(MetricAggregator):
@@ -40,14 +43,15 @@ class GaugeMetric(MetricAggregator):
     def aggregate(self, v):
         self.value = v
 
-    def flush_unsafe(self):
-        return {
-            "metric_type": "gauge",
-            "name": self.name,
-            "tags": self.tags,
-            "rate": self.rate,
-            "fvalue": self.value,
-        }
+    # TODO: potentially add this function
+    # def flush_unsafe(self):
+    #     return {
+    #         "metric_type": "gauge",
+    #         "name": self.name,
+    #         "tags": self.tags,
+    #         "rate": self.rate,
+    #         "fvalue": self.value,
+    #     }
 
 
 class SetMetric(MetricAggregator):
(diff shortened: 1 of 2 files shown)


datadog/dogstatsd/metrics.py
```diff
@@ -0,0 +1,77 @@
...
+class GaugeMetric(MetricAggregator):
+    def __init__(self, name, value, tags, rate):
+        super(GaugeMetric, self).__init__(name, tags, rate)
+        self.value = value
+
+    def aggregate(self, v):
+        self.value = v
+
+    def flush_unsafe(self):
+        return {
```
rayz, line 44: Would it be better to create a new class for this object? Not too familiar with python conventions

andrewqian2001datadog: image

I was thinking that the purpose of having a Metric struct was so that they can be grouped into one list, since we can already do that in python I thought it wasn't necessary.

gh123man: Maybe we can omit the `flush` function for now since the current implementation of `flush_unsafe` makes assumptions about the structure of the data in regard to how it will be serialized and flushed before that part of the code is written. 

I'd consider making these functions `TODO` for now and introduce the flush data structure in the next PR. It's hard to argue which approach is better without seeing how it will ultimately be used. In fact - we may not even need these flush functions depending on how the metric classes are going to be used.

For example - I wonder if it makes sense to move metric serialization as a function on each metric class? So you can do something like:

```python
countMetric = CountMetric()
flush(countMetric.serialize())
```

datadog/dogstatsd/metrics.py
```diff
@@ -0,0 +1,77 @@
...
+            "fvalue": self.value,
+        }
+
+
+class SetMetric(MetricAggregator):
+    def __init__(self, name, value, tags, rate):
+        super(SetMetric, self).__init__(name, tags, rate)
+        self.data = set()
+        self.data.add(value)
+        self.lock = Lock()
```
gh123man, line 58: Similarly - maybe we should wait to add synchronization primitives until we actually have multithreaded code running. We can keep this PR focused to just introducing the metric classes and their associated properties.

andrewqian2001datadog: the serialization function needs values like namespace and container_id which is set in the base.py class. I think it would be more convenient to leave it as a function in base.py or move it to aggregator.py? Not sure.

This is what aggregator.py looks like so far https://github.com/DataDog/datadogpy/pull/838

Maybe we can move the serialization function there since it should have access to the DogStatsd client and its properties? It seems more intuitive that aggregator.py would handle serialization of a metric.

gh123man (dismissed):
LGTM!

tests/unit/dogstatsd/test_metrics.py
```diff
@@ -0,0 +1,80 @@
...
+        s = SetMetric("test", "value1", ["tag1", "tag2"], 1, 1713804588)
+        s.aggregate("value2")
+        s.aggregate("value2")
+        self.assertEqual(s.data, {"value1", "value2"})
+        self.assertEqual(s.name, "test")
+        self.assertEqual(s.tags, ["tag1", "tag2"])
+        self.assertEqual(s.rate, 1)
+        self.assertEqual(s.timestamp, 1713804588)
+
+if __name__ == '__main__':
```
gh123man, line 79: Our other tests classes don't have this. Not sure if required or not.

andrewqian2001datadog: The API in base.py has it so i figured it was necessary

image

andrewqian2001datadog: Oh you meant that you dont need to test if there is a timestamp or not. Oops, yeah i'm not sure if its necessary to check the field but I figured that I might as well just check all the cases.

skarimo (approved):

rayz (approved):

PR #841, opened 2024-07-03
add aggregator 

andrewqian2001datadog:
### What does this PR do?

This PR adds the aggregator functionality for client side aggregation, it is pushing to a feature branch. This has already been implemented for the [GO client](https://github.com/DataDog/datadog-go/pull/139/files#diff-0d2ff0ba6d3144d1b0aea4da6445ea870b0ae794a5a5a8955dfa6c1b9150539a).

[Issue](https://datadoghq.atlassian.net/jira/software/c/projects/AMLII/boards/5315?selectedIssue=AMLII-1856)
### Description of the Change

Adds a class that periodically aggregates (this is not implemented yet as it will rely on functionality from the base.py class) metrics for a given `flush_interval` and then sends the aggregated metrics to intake. A metric will be aggregated if there was another [matching context](https://docs.google.com/document/d/1AoveWKoMQc65hYbvZ27uE7pQ_I5nsGfvv25EcNVPLvQ/edit#heading=h.no4x6fjugsh) sent within the same flush interval.

### Possible Drawbacks

A possible drawback is that the current code relies on object oriented programming (inheritance) for the `add_metric` and other functions, it requires that the `MetricAggregator` objects will have the same fields

### Verification Process

Unit tests that verify the aggregator.py class matches the existing behavior in the client side aggregator for the [go client](https://github.com/DataDog/datadog-go/pull/139/files#diff-f554159d70c08124ea27e909de77e6b7162f782cbc12363a0cac8825834a2210).

### Release Notes

No behavioral changes until the `base.py` class is changed

### DIFF
--- a/datadog/dogstatsd/aggregator.py
+++ b/datadog/dogstatsd/aggregator.py
@@ -1,0 +1,86 @@
+import threading
+import time
+from enum import Enum
+
+from datadog.dogstatsd.metrics import MetricAggregator, CountMetric, GaugeMetric, SetMetric
+from datadog.dogstatsd import DogStatsd
+
+class MetricType(Enum):
+    COUNT = 'c'
+    GAUGE = 'g'
+    SET = 's'
+
+class Aggregator(object):
+    def __init__(self, client: DogStatsd):
+        self.client = client
+
+        self.metrics_map = {
+            MetricType.COUNT: {},
+            MetricType.GAUGE: {},
+            MetricType.SET: {}
+        }
+        self.locks = {
+            MetricType.COUNT: threading.RLock(),
+            MetricType.GAUGE: threading.RLock(),
+            MetricType.SET: threading.RLock()
+        }
+
+        self.closed = threading.Event()
+        self.exited = threading.Event()
+
+    def start(self, flush_interval):
+        self.flush_interval = flush_interval
+        self.ticker = threading.Timer(self.flush_interval, self.tick)
+        self.ticker.start()
+
+    def tick(self):
+        while not self.closed.is_set():
+            self.send_metrics()
+            time.sleep(self.flush_interval)
+        self.exited.set()
+
+    def send_metrics(self):
+        for metric in self.flush_metrics():
+            # TODO: change the _report function in base.py to handle new data
+            serialized_metric = self.client._serialize_metric(metric.name, metric.type, metric.value, metric.tags, metric)
+            self.client._report(serialized_metric)
+
+    def stop(self):
+        self.closed.set()
+        self.ticker.cancel()
+        self.exited.wait()
+        self.send_metrics()
+
+    def flush_metrics(self):
+        metrics = []
+
+        for metric_type in self.metrics_map.keys():
+            with self.locks[metric_type]:
+                current_metrics = self.metrics_map[metric_type]
+                self.metrics_map[metric_type] = {}
+
+            # TODO: the data type for unsafe_flush still needs to be decided.
+            for metric in current_metrics.values():
+                metrics.extend(metric if isinstance(metric, SetMetric) else [metric])
+
+        return metrics
+
+    def get_context(self, name, tags):
+        return "{}:{}".format(name, ",".join(tags))
+
+    def count(self, name, value, tags, rate, timestamp=0):
+        return self.add_metric(MetricType.COUNT, CountMetric, name, value, tags, rate, timestamp)
+
+    def gauge(self, name, value, tags, rate, timestamp=0):
+        return self.add_metric(MetricType.GAUGE, GaugeMetric, name, value, tags, rate, timestamp)
+
+    def set(self, name, value, tags, rate, timestamp=0):
+        return self.add_metric(MetricType.SET, SetMetric, name, value, tags, rate, timestamp)
+
+    def add_metric(self, metric_type: MetricType, metric_class: MetricAggregator, name, value, tags, rate, timestamp=0):
+        context = self.get_context(name, tags)
+        with self.locks[metric_type]:
+            if context in self.metrics_map[metric_type]:
+                self.metrics_map[metric_type][context].aggregate(value)
+            else:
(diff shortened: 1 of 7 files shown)


datadog/dogstatsd/metrics.py
```diff
@@ -34,20 +42,39 @@ def __init__(self, name, value, tags, rate, timestamp=0):
...
     def aggregate(self, v):
-        self.data.add(v)
+        with self.lock:
+            self.data.add(v)
 
-    # TODO: This may be implemented once flushing aggregated metrics is supported
     def unsafe_flush(self):
-        pass
+        with self.lock:
+            if not self.data:
```
rayz, line 69: not necessary i believe, the `else` code will return empty list

andrewqian2001datadog: 🫡

datadog/dogstatsd/metrics.py
```diff
@@ -34,20 +42,39 @@ def __init__(self, name, value, tags, rate, timestamp=0):
...
 
     def aggregate(self, v):
-        self.data.add(v)
+        with self.lock:
+            self.data.add(v)
 
-    # TODO: This may be implemented once flushing aggregated metrics is supported
     def unsafe_flush(self):
-        pass
+        with self.lock:
```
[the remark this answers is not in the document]
andrewqian2001datadog: 🫡

datadog/dogstatsd/aggregator.py
```diff
@@ -0,0 +1,77 @@
...
+            self.send_metrics()
+            time.sleep(self.flush_interval)
+        self.exited.set()
+
+    def send_metrics(self):
+        for metric in self.flush_metrics():
+            self.client._report(metric.name, metric.type, metric.value, metric.tags, 
+                timestamp=metric.timestamp if metric.timestamp else 0)
+
+    def stop(self):
```
rayz, line 40: can we add some tests to test the `start` and `stop` behavior?

datadog/dogstatsd/aggregator.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def start(self, flush_interval):
+        self.flush_interval = flush_interval
+        self.ticker = threading.Timer(self.flush_interval, self.tick)
+        self.ticker.start()
+
+    def tick(self):
+        while not self.closed.is_set():
+            self.send_metrics()
+            time.sleep(self.flush_interval)
```
gh123man, line 32: Do we need to sleep here if `tick` is being executed on a ticker every `flush_interval`?

datadog/dogstatsd/metrics.py
```diff
@@ -34,20 +42,36 @@ def __init__(self, name, value, tags, rate, timestamp=0):
...
 class SetMetric(MetricAggregator):
     def __init__(self, name, value, tags, rate, timestamp=0):
         super(SetMetric, self).__init__(name, tags, rate, timestamp)
         self.data = set()
         self.data.add(value)
+        self.lock = Lock()
 
     def aggregate(self, v):
-        self.data.add(v)
+        with self.lock:
```
gh123man, line 64: Do we need this lock? as far as I can tell, this is only ever called while the associated metric lock is held

datadog/dogstatsd/aggregator.py
andrewqian2001datadog: I don't think that the timer repeats every flush interval.

image

It seems like it is just called once.

Is this a problem? Do we want a thread for every flush interval?

gh123man: Oh my bad I misunderstood how timers work. This looks fine :)

We already maintain a [flush loop for buffered metrics in](https://github.com/DataDog/datadogpy/blob/f20e449db05e55aff389289b55f9a57b536b37db/datadog/dogstatsd/base.py#L517) `base.py`

I wonder if this can be re-used or refactored to be shared with the aggregator since it's quite similar in usage.

andrewqian2001datadog:
For this error
image

I think I need to add a dependency?
https://pypi.org/project/enum34/#downloads

image

datadog/dogstatsd/aggregator.py
```diff
@@ -0,0 +1,75 @@
...
+
+class Aggregator(object):
+    def __init__(self, client):
+        self.client = client
+        self.metrics_map = {
+            MetricType.COUNT: {},
+            MetricType.GAUGE: {},
+            MetricType.SET: {},
+        }
+        self.locks = {
```
rayz, line 18:
```suggestion
self._locks = {
```

datadog/dogstatsd/aggregator.py
```diff
@@ -0,0 +1,75 @@
...
+        }
+        self._locks = {
+            MetricType.COUNT: threading.RLock(),
+            MetricType.GAUGE: threading.RLock(),
+            MetricType.SET: threading.RLock(),
+        }
+
+    def start(self, flush_interval):
+        self.flush_interval = flush_interval
+        self.client._start_flush_thread(flush_interval, self.send_metrics)
```
gh123man, line 26: This is looking good. One more design suggestion that may further simplify things:

Currently `Aggregator` takes an instance of the client. Presumably, the aggregator instance will be created by the client during startup (correct me if I am wrong) - this will create a dependency cycle. (which is not a "problem" in python, but could make testing because the components are tightly coupled). 

It may make the code simpler if the client made calls to the aggregator and the aggregator was not aware of the client at all. For example:

- you can remove the `start` and `stop` methods here since the flush thread is managed by the client. 
- `send_metrics` could instead return the list of metrics to flush and `client._report` can be executed from within the client. 

This would also simply writing tests since the aggregator simply takes metrics in, and aggregated metrics come out - and the threading/lifecycle is completely controlled by the client. 

Let me know what you think or if you have any questions. I think this should be a relatively small code change considering the state this change is in right now :)

andrewqian2001datadog: Yeah this makes sense, thanks!

gh123man (approved):
🚀

PR #844, opened 2024-07-17
Add client side aggregation

andrewqian2001datadog:

### What does this PR do?

[Project description
](https://docs.google.com/document/d/11fPSNMnp5tMLRK16VgXa6FJ7fTpzggKevpiO6MsQ0H0/edit?usp=sharing)
This PR adds simple client side aggregation (aggregating metrics before they are received by the agent) for the python client. This PR combines two previous PRs together ([1](https://github.com/DataDog/datadogpy/pull/841), [2](https://github.com/DataDog/datadogpy/pull/837))

Client side aggregation has already been implemented in [Go](https://github.com/DataDog/datadog-go/pull/139), Java, and the C# dogstatsd clients

### Description of the Change

Periodically aggregates metrics (Count, Gauge, Set) for a given `statsd_aggregation_flush_interval` and then sends the aggregated metrics to intake. A metric will be aggregated if `statsd_disable_aggregating` is `False` and there was another [matching context](https://docs.google.com/document/d/1AoveWKoMQc65hYbvZ27uE7pQ_I5nsGfvv25EcNVPLvQ/edit#heading=h.no4x6fjugsh) sent within the same flush interval.

### Verification Process

Unit tests that verify the aggregator.py class matches the existing behavior in the client side aggregator for the [go client](https://github.com/DataDog/datadog-go/pull/139/files#diff-f554159d70c08124ea27e909de77e6b7162f782cbc12363a0cac8825834a2210).

Unit tests that verify the new metric objects (merics.py) match the existing behavior in the [client side aggregator for Go](https://github.com/DataDog/datadogpy/pull/url).

**Local testing**:
[Confluence doc](https://datadoghq.atlassian.net/wiki/spaces/~71202058eaddc626774b6382b0a98b3e9f4a84/pages/3978855545/Local+testing+for+python+client+side+aggregation+project)
[More Docs](https://dddev.datadoghq.com/notebook/9246140/python-client-side-aggregation-testing)

Bench Mark tests (TODO as a stretch goal)

### Additional Notes

We currently do not allow buffering and aggregation at the same time, if both enabled, aggregation will occur instead of buffering.

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -83,7 +83,7 @@ def initialize(
     :type statsd_disable_aggregating: boolean
 
     :param statsd_aggregation_flush_interval: Sets the flush interval for aggregation
-                                     (default: 3 seconds)
+                                     (default: 2 seconds)
     :type statsd_aggregation_flush_interval: float
 
     :param statsd_use_default_route: Dynamically set the statsd host to the default route
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -39,7 +39,7 @@ def initialize(
     statsd_port=None,  # type: Optional[int]
     statsd_disable_aggregating=False,  # type: bool
     statsd_disable_buffering=True,  # type: bool
-    statsd_aggregation_flush_interval=3,  # type: float
+    statsd_aggregation_flush_interval=2,  # type: float
     statsd_use_default_route=False,  # type: bool
     statsd_socket_path=None,  # type: Optional[str]
     statsd_namespace=None,  # type: Optional[str]
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -37,7 +37,7 @@ def initialize(
     api_host=None,  # type: Optional[str]
     statsd_host=None,  # type: Optional[str]
     statsd_port=None,  # type: Optional[int]
-    statsd_disable_aggregating=True,  # type: bool
+    statsd_disable_aggregator=True,  # type: bool
     statsd_disable_buffering=True,  # type: bool
     statsd_aggregation_flush_interval=2,  # type: float
     statsd_use_default_route=False,  # type: bool
@@ -78,9 +78,9 @@ def initialize(
                                      (default: True).
     :type statsd_disable_buffering: boolean
 
-    :param statsd_disable_aggregating: Enable/disable statsd client aggregation support
+    :param statsd_disable_aggregator: Enable/disable statsd client aggregation support
                                      (default: True).
-    :type statsd_disable_aggregating: boolean
+    :type statsd_disable_aggregator: boolean
 
     :param statsd_aggregation_flush_interval: Sets the flush interval for aggregation
                                      (default: 2 seconds)
@@ -138,7 +138,7 @@ def initialize(
     if statsd_constant_tags:
         statsd.constant_tags += statsd_constant_tags
 
-    if statsd_disable_aggregating:
+    if statsd_disable_aggregator:
         statsd.disable_aggregation()
     else:
         statsd.enable_aggregation(statsd_aggregation_flush_interval)
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -139,8 +139,8 @@ def initialize(
         statsd.constant_tags += statsd_constant_tags
 
     statsd._aggregation_flush_interval = statsd_aggregation_flush_interval
-    statsd.disable_buffering = statsd_disable_buffering
     statsd.disable_aggregating = statsd_disable_aggregating
+    statsd.disable_buffering = statsd_disable_buffering
     api._return_raw_response = return_raw_response
 
     # HTTP client and API options
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -38,7 +38,7 @@ def initialize(
     statsd_host=None,  # type: Optional[str]
     statsd_port=None,  # type: Optional[int]
     statsd_disable_buffering=True,  # type: bool
-    statsd_disable_aggregating=True,  # type: bool
+    statsd_disable_aggregating=False,  # type: bool
     statsd_aggregation_flush_interval=3,  # type: float
     statsd_use_default_route=False,  # type: bool
     statsd_socket_path=None,  # type: Optional[str]
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -39,7 +39,7 @@ def initialize(
     statsd_port=None,  # type: Optional[int]
     statsd_disable_buffering=True,  # type: bool
     statsd_disable_aggregating=True,  # type: bool
-    statsd_aggregation_flush_interval=3, # type: float
+    statsd_aggregation_flush_interval=3,  # type: float
     statsd_use_default_route=False,  # type: bool
     statsd_socket_path=None,  # type: Optional[str]
     statsd_namespace=None,  # type: Optional[str]
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -37,8 +37,8 @@ def initialize(
     api_host=None,  # type: Optional[str]
     statsd_host=None,  # type: Optional[str]
     statsd_port=None,  # type: Optional[int]
-    statsd_disable_buffering=True,  # type: bool
     statsd_disable_aggregating=False,  # type: bool
+    statsd_disable_buffering=True,  # type: bool
     statsd_aggregation_flush_interval=3,  # type: float
     statsd_use_default_route=False,  # type: bool
     statsd_socket_path=None,  # type: Optional[str]
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -39,6 +39,7 @@ def initialize(
     statsd_port=None,  # type: Optional[int]
     statsd_disable_buffering=True,  # type: bool
     statsd_disable_aggregating=True,  # type: bool
+    statsd_aggregation_flush_interval=3, # type: float
     statsd_use_default_route=False,  # type: bool
     statsd_socket_path=None,  # type: Optional[str]
     statsd_namespace=None,  # type: Optional[str]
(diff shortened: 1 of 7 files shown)


datadog/__init__.py
```diff
@@ -37,7 +37,9 @@ def initialize(
     api_host=None,  # type: Optional[str]
     statsd_host=None,  # type: Optional[str]
     statsd_port=None,  # type: Optional[int]
+    statsd_disable_aggregating=False,  # type: bool
```
gh123man, line 40: I think at least for now we should default this to disabled. Once we do more testing and wider deployments on more workloads we can consider making it the default :)

datadog/dogstatsd/base.py
```diff
@@ -24,9 +24,13 @@
     # pypy has the same module, but capitalized.
     import Queue as queue  # type: ignore[no-redef]
 
+# pylint: disable=unused-import
```
gh123man, line 27: NIT can we resolve/remove this? If an import is unused, I think we can omit it?

datadog/dogstatsd/base.py
```diff
@@ -45,9 +49,12 @@
...
 
 # Buffering-related values (in seconds)
-DEFAULT_FLUSH_INTERVAL = 0.3
-MIN_FLUSH_INTERVAL = 0.0001
+DEFAULT_BUFFERING_FLUSH_INTERVAL = 0.3
+MIN_BUFFERING_FLUSH_INTERVAL = 0.0001
 
+# Aggregation-related values (in seconds)
+DEFAULT_AGGREGATION_FLUSH_INTERVAL = 2
+MIN_AGGREGATION_FLUSH_INTERVAL = 0.01
```
gh123man, line 57: Maybe this can be the same as `MIN_BUFFERING_FLUSH_INTERVAL`? (we could just keep `MIN_FLUSH_INTERVAL` as it was and use it for both?) 

Unless this new value has some significance. I don't see why the minimum couldn't be much smaller even if impractical.

datadog/dogstatsd/base.py
```diff
@@ -128,31 +138,35 @@ class DogStatsd(object):
...
-        max_buffer_len=0,                       # type: int
-        container_id=None,                      # type: Optional[Text]
-        origin_detection_enabled=True,          # type: bool
-        socket_timeout=0,                       # type: Optional[float]
-        telemetry_socket_timeout=0,             # type: Optional[float]
-        disable_background_sender=True,         # type: bool
-        sender_queue_size=0,                    # type: int
-        sender_queue_timeout=0,                 # type: Optional[float]
-        track_instance=True,                    # type: bool
+        host=DEFAULT_HOST,  # type: Text
```
gh123man, line 141: it looks like the formatting got changed here. 
I'm not sure if a linter was flagging this or if a formatter was misconfigured, but the previous (column aligned) formatting is better IMO. Maybe we can revert this?

datadog/dogstatsd/base.py
```diff
@@ -328,8 +345,9 @@ def __init__(
 
         # Check for deprecated option
         if max_buffer_size is not None:
-            log.warning("The parameter max_buffer_size is now deprecated and is not used anymore")
-
+            log.warning(
```
gh123man, line 348: I think your linter may have [ignored the config](https://github.com/DataDog/datadogpy/blob/master/tox.ini#L49). it looks like this got line truncated, but our line limit is 120 chars. 

Can you re-run the linter suggestions with the config applied? It will remove several unnecessary changes in this diff.

datadog/dogstatsd/base.py
```diff
@@ -423,26 +441,51 @@ def __init__(
...
-        self._send = self._send_to_buffer
         self._disable_buffering = disable_buffering
-        if self._disable_buffering:
-            self._send = self._send_to_server
-            log.debug("Statsd buffering is disabled")
+        self._disable_aggregating = disable_aggregating
 
+        self._flush_interval = flush_interval
+        self._aggregation_flush_interval = aggregation_flush_interval
+        # We make the _buffering_flush_thread and _aggregation_flush_thread a list so that it is a mutable object,
```
gh123man, line 453: is there a reason we cannot continue using `self._flush_thread` in favor of the list thread containers? 
Since aggregation and buffering won't be enabled at the same time (at this point) `self._flush_thread` will only be occupied by one or the other routine.

gh123man (changes requested):
Did a first pass. Left a few comments/questions. 

I think the first thing to address is getting the linter to not change a bunch of unnecessary lines since it makes it hard to determine which lines are functional changes, and which is just formatting. 

Ill do another review once these are addressed :)

datadog/dogstatsd/base.py
andrewqian2001datadog: Its used in the comments and they give me a lint issue when I removed them

datadog/dogstatsd/base.py
andrewqian2001datadog: No I think that works as well, I just thought designing it like this would be more flexible and at the time I wrote the code I dont think I knew whether or not they would be allowed at the same time.

gh123man: Makes sense. In that case - let's do the simple thing for now and reuse `self. _flush_thread`.

andrewqian2001datadog: Sounds good

gh123man (changes requested):

datadog/dogstatsd/base.py
```diff
@@ -541,27 +585,27 @@ def _flush_thread_loop(self, flush_interval):
...
         )
 
     # Note: Invocations of this method should be thread-safe
     def _stop_flush_thread(self):
         if not self._flush_thread:
             return
-
         try:
-            self.flush()
+            if self.disable_aggregation:
```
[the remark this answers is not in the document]
andrewqian2001datadog: Oh this was mb it was supposed to be self._disable_aggregating, thanks!

datadog/dogstatsd/base.py
```diff
@@ -598,7 +642,35 @@ def disable_buffering(self, is_disabled):
...
+                return
+
+            self._disable_aggregating = True
+
+            # If aggregation has been disabled, flush and kill the background thread
+            # otherwise start up the flushing thread and enable aggregation.
+            self._stop_flush_thread()
+            log.debug("Statsd aggregation is disabled")
+
+    def enable_aggregation(self, aggregation_flush_interval=DEFAULT_AGGREGATION_FLUSH_INTERVAL):
```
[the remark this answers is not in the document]
andrewqian2001datadog: the __init__ function is sometimes not called with any parameters set and just has the default parameters as we don't require the user to call initialize before sending metrics. So I believe we want a way to start aggregation.

This was a convo I had with Vikentiy 

"That’s because we create the global statsd instance with all default parameters ([here](https://github.com/DataDog/datadogpy/blob/add-client-side-aggregation/datadog/dogstatsd/base.py#L1630)) - we do not require user to call initialize before sending metrics. Then, if the user calls configuration routine, we change the state of the object as needed. We would need at least a way to enable aggregation on an object that already exists, ..."

tests/unit/dogstatsd/test_metrics.py
```diff
@@ -52,29 +52,22 @@ def test_gauge_metric_aggregate(self):
         self.assertEqual(g.timestamp, 1713804588)
 
     def test_new_set_metric(self):
-        s = SetMetric("test", "value1", ["tag1", "tag2"], 1, 1713804588)
+        s = SetMetric("test", "value1", ["tag1", "tag2"], 1)
         self.assertEqual(s.data, {"value1"})
         self.assertEqual(s.name, "test")
         self.assertEqual(s.tags, ["tag1", "tag2"])
         self.assertEqual(s.rate, 1)
-        self.assertEqual(s.timestamp, 1713804588)
```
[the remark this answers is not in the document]
andrewqian2001datadog: Yeah set metric doesn't have timestamp, felt kind of misleading to instantiate it with one.

tests/unit/dogstatsd/test_statsd.py
```diff
@@ -803,7 +803,7 @@ def func(arg1, arg2, kwarg1=1, kwarg2=1):
         self.assertEqual('distributed.test', name)
         self.assert_almost_equal(500, float(value), 100)
 
-    def test_timed(self):
```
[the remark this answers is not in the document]
andrewqian2001datadog: Yeah mb, thanks for the catch

datadog/dogstatsd/base.py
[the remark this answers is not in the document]
gh123man: Oh awesome. Ill give this another look 👍

datadog/dogstatsd/base.py
gh123man: Spoke offline. seems like the linter wasn't running on this file before.

gh123man (dismissed):
LGTM
Nice work on this 🚀

rayz (dismissed):
looks good

datadog/__init__.py
```diff
@@ -128,8 +138,11 @@ def initialize(
     if statsd_constant_tags:
         statsd.constant_tags += statsd_constant_tags
 
+    if statsd_disable_aggregating:
+        statsd.disable_aggregation()
+    else:
+        statsd.enable_aggregation(statsd_aggregation_flush_interval)
```
skarimo, line 144: Do you need to make sure buffering is disabled here?

skarimo (approved):

datadog/__init__.py
andrewqian2001datadog: I think it should already be handled [here](https://github.com/DataDog/datadogpy/blob/add-client-side-aggregation/datadog/dogstatsd/base.py#L549) since we wont have a flush thread for buffering if aggregation is enabled

skarimo (dismissed):

rayz (dismissed):

rayz (approved):

PR #851, opened 2024-08-21
enable the ability for buffering and aggregation to work at the same

andrewqian2001datadog:
### What does this PR do?

[Issue](https://datadoghq.atlassian.net/jira/software/c/projects/AMLII/boards/5315?selectedIssue=AMLII-1982)

enable the ability for buffering and aggregation to work at the same

### Description of the Change

Refactors code to enable aggregation/buffering at the same time. The aggregator will add the aggregated metrics into the buffer at the end of every flush interval before the buffer sends them to the agent. Aggregation/buffering share the same flush thread and flush interval.

### Possible Drawbacks

aggregation and buffering seems tightly coupled, for [example](https://github.com/DataDog/datadogpy/blob/enable-buffering-aggregation-simultaneously/datadog/dogstatsd/base.py#L625)

Instead of killing the flush thread when buffering is disabled, (I assume) it needs to check if both aggregation and buffering is disabled before killing the flush thread. May be a good idea to add a if statement in [stop_flush_thread](https://github.com/DataDog/datadogpy/blob/enable-buffering-aggregation-simultaneously/datadog/dogstatsd/base.py#L579) to not kill the flush thread if either aggregation or buffering is enabled?
### Verification Process

Added unit tests to test aggregation and buffering at the same time.

For Local testing:
Follow these [steps](https://datadoghq.atlassian.net/wiki/spaces/~71202058eaddc626774b6382b0a98b3e9f4a84/pages/3978855545/Local+testing+for+python+client+side+aggregation+project) for local testing and in the main.py file, enable aggregation and buffering at the same time

main.py
```
from datadog import initialize, statsd
import time

options = {
    "statsd_host": "127.0.0.1",
    "statsd_port": 8125,
    "statsd_disable_buffering" : False,
    "statsd_disable_aggregation" : False,
    "statsd_aggregation_flush_interval" : 10
}

initialize(**options)

while(1):
  print("-------------------------------------------------")
  print("metric added")
  statsd.count('testandrew123.count', 1, tags=["environment:dev"], sample_rate=1)
  time.sleep(1)
```
You can add a print statement [here](https://github.com/DataDog/datadogpy/blob/enable-buffering-aggregation-simultaneously/datadog/dogstatsd/base.py#L814) to ensure that the agggregated metrics are added to buffer, we would expect that the count metrics are aggregated (single metric with value of 10) inside the buffer

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -82,7 +82,7 @@ def initialize(
                                      (default: True).
     :type statsd_disable_aggregator: boolean
 
-    :param statsd_aggregation_flush_interval: If aggregation is enabled, set the flush interval for 
+    :param statsd_aggregation_flush_interval: If aggregation is enabled, set the flush interval for
                     aggregation/buffering
                                      (default: 0.3 seconds)
     :type statsd_aggregation_flush_interval: float
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -37,7 +37,7 @@ def initialize(
     api_host=None,  # type: Optional[str]
     statsd_host=None,  # type: Optional[str]
     statsd_port=None,  # type: Optional[int]
-    statsd_disable_aggregator=True,  # type: bool
+    statsd_disable_aggregation=True,  # type: bool
     statsd_disable_buffering=True,  # type: bool
     statsd_aggregation_flush_interval=0.3,  # type: float
     statsd_use_default_route=False,  # type: bool
@@ -78,9 +78,9 @@ def initialize(
                                      (default: True).
     :type statsd_disable_buffering: boolean
 
-    :param statsd_disable_aggregator: Enable/disable statsd client aggregation support
+    :param statsd_disable_aggregation: Enable/disable statsd client aggregation support
                                      (default: True).
-    :type statsd_disable_aggregator: boolean
+    :type statsd_disable_aggregation: boolean
 
     :param statsd_aggregation_flush_interval: If aggregation is enabled, set the flush interval for
                     aggregation/buffering
@@ -139,7 +139,7 @@ def initialize(
     if statsd_constant_tags:
         statsd.constant_tags += statsd_constant_tags
 
-    if statsd_disable_aggregator:
+    if statsd_disable_aggregation:
         statsd.disable_aggregation()
     else:
         statsd.enable_aggregation(statsd_aggregation_flush_interval)
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -82,7 +82,8 @@ def initialize(
                                      (default: True).
     :type statsd_disable_aggregator: boolean
 
-    :param statsd_aggregation_flush_interval: If aggregation is enabled, set the flush interval for aggregation/buffering
+    :param statsd_aggregation_flush_interval: If aggregation is enabled, set the flush interval for 
+                    aggregation/buffering
                                      (default: 0.3 seconds)
     :type statsd_aggregation_flush_interval: float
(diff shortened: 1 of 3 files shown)


datadog/dogstatsd/base.py
```diff
@@ -147,7 +145,7 @@ def __init__(
         host=DEFAULT_HOST,                      # type: Text
         port=DEFAULT_PORT,                      # type: int
         max_buffer_size=None,                   # type: None
-        flush_interval=DEFAULT_BUFFERING_FLUSH_INTERVAL,  # type: float
+        flush_interval=DEFAULT_FLUSH_INTERVAL,  # type: float
         disable_aggregating=True,               # type: bool
```
gh123man, line 149: I missed this in the last PR, but should we call this `disable_aggregation`?
aggregation seems to be the term we use throughout.

datadog/__init__.py
```diff
@@ -39,7 +39,7 @@ def initialize(
     statsd_port=None,  # type: Optional[int]
     statsd_disable_aggregator=True,  # type: bool
```
gh123man, line 40: Similar to my other comment, but we should be consistent. maybe rename this to `statsd_disable_aggregation`?

gh123man (approved):
Code changes look good to me. Just a minor suggestion to align the config naming. 
Lets be sure we manually test/verify: 
- aggregation + buffering
- buffering only
- aggregation only
(just for completeness).

datadog/dogstatsd/base.py
andrewqian2001datadog: We have a function called [disable_aggregation](https://github.com/DataDog/datadogpy/blob/enable-buffering-aggregation-simultaneously/datadog/dogstatsd/base.py#L632) so the variable name is a bit different

datadog/__init__.py
andrewqian2001datadog: Yeah this sounds good to me, I just changed the name to one that someone suggested

datadog/dogstatsd/base.py
gh123man: we seem to have the same conflict with `disable_buffering` so I think it should be fine. Though I will admit it's a bit confusing having multiple re-definitions of the same name. (I think this confused me in a previous PR comment)

rayz:
would there be any potential problems with having the default aggregation flush interval that low? old `2s` -> new `.3s`

datadog/dogstatsd/base.py
andrewqian2001datadog: Sounds good, I can make the change

gh123man (approved):

datadog/__init__.py
```diff
@@ -78,12 +78,13 @@ def initialize(
                                      (default: True).
     :type statsd_disable_buffering: boolean
 
-    :param statsd_disable_aggregator: Enable/disable statsd client aggregation support
+    :param statsd_disable_aggregation: Enable/disable statsd client aggregation support
```
carlosroman, line 81: What the reason for changing the name? Just curious as if anyone using this feature today will have to update to using name now.

gh123man: I asked him to change this (and the other setting `disable_aggregating` to be consistent with the rest of our naming. (Since this feature is not documented yet, I expect it to be ok. - Would like to document this for the next release)

HantingZhang2 (approved):

carlosroman:
@gorangasic  have created the release PR #856. Hopefully should all get approved really soon.

PR #853, opened 2024-09-04
add buffered_metrics object type

andrewqian2001datadog:
### What does this PR do?

Adds BufferedMetric object to allow for extended aggregation (Distribution, Histogram, Timing metrics)

Based on off existing [Go code](https://github.com/DataDog/datadog-go/blob/master/statsd/metrics.go#L124)

### Description of the Change

Adds a `buffered_metrics.py` class that has all the properties that Distribution, Histogram and Timing metrics require

### Verification Process

Added unit tests matching the [existing tests](https://github.com/DataDog/datadog-go/blob/master/statsd/metrics_test.go#L134C6-L134C28) 

### Additional Notes

1. This PR will merge onto a feature branch.

### DIFF
--- a/datadog/dogstatsd/buffered_metrics.py
+++ b/datadog/dogstatsd/buffered_metrics.py
@@ -3,13 +3,13 @@ from datadog.dogstatsd.metric_types import MetricType
 
 
 class BufferedMetric(object):
-    def __init__(self, name, tags, metric_type, max_metric_samples=0, specified_rate=1.0):
+    def __init__(self, name, value, tags, metric_type, max_metric_samples=0, specified_rate=1.0):
         self.name = name
         self.tags = tags
         self.metric_type = metric_type
         self.max_metric_samples = max_metric_samples
         self.specified_rate = specified_rate
-        self.data = []
+        self.data = [value]
         self.stored_metric_samples = 1
         self.total_metric_samples = 1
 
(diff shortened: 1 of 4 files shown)


datadog/dogstatsd/buffered_metrics.py
```diff
@@ -0,0 +1,70 @@
...
+        self.total_metrics += 1
+
+    def maybe_add_metric(self, value):
+        if self.max_metrics > 0:
+            if self.stored_metrics >= self.max_metrics:
+                i = random.randint(0, self.total_metrics - 1)
+                if i < self.max_metrics:
+                    self.data[i] = value
+            else:
+                self.data.append(value)
```
rayz, line 28: Re: your 2nd question, they store the value like [this ](https://github.com/DataDog/datadog-go/blob/master/statsd/metrics.go#L176) since they have `self.data` presized [here](https://github.com/DataDog/datadog-go/blob/20af2dbfabbbe6bd0347780cd57ed931f903f223/statsd/metrics.go#L264C3-L264C37)

rayz (changes requested):

datadog/dogstatsd/buffered_metrics.py
```diff
@@ -0,0 +1,70 @@
...
+
+class BufferedMetric(object):
+    def __init__(self, name, tags, metric_type, max_metrics=0, specified_rate=1.0):
+        self.name = name
+        self.tags = tags
+        self.metric_type = metric_type
+        self.max_metrics = max_metrics
+        self.specified_rate = specified_rate
+        self.data = []
+        self.stored_metrics = 0
```
rayz, line 13: Should these be set to `1`?

datadog/dogstatsd/buffered_metrics.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def flush(self):
+        total_metrics = self.total_metrics
+        if self.specified_rate != 1.0:
+            rate = self.specified_rate
+        else:
+            if total_metrics != 0:
+                rate = self.stored_metrics / total_metrics
+            else:
+                rate = 1.0
```
rayz, line 45: `total_samples` would never be 0 i believe since if an instance of the object exists, theres at least 1 sample (?)

datadog/dogstatsd/buffered_metrics.py
andrewqian2001datadog: The object will be instantiated immediately in the aggregator w/o a metric being added yet so i think 0 is fine

datadog/dogstatsd/buffered_metrics.py
```diff
@@ -0,0 +1,70 @@
+import random
+from datadog.dogstatsd.metric_types import MetricType
+
+
+class BufferedMetric(object):
```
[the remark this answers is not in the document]
andrewqian2001datadog: for naming, I'm open to either sample or add_metric but I think we should keep it consistanted with the existing code in this repo, will follow up with Brian later to see if he has any thoughts.

For locks, when we added the other metric object types, [we waited to add the locks](https://github.com/DataDog/datadogpy/pull/837#discussion_r1658904919) and just kept the PR to the metric properties. I was thinking of doing the same here.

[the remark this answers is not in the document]
gh123man: Not a strong preference here. 

Slight preference for `sample` given the context in this case. Since each metric type inherits `BufferedMetric`, `sample` is an operation on the metric unlike the simple aggregator where `add_metric` adds a metric of a specific type. 

(could always go with `metric_sample` if you prefer 🤷 best of both)

rayz (approved):

PR #863, opened 2024-10-28
[AMLII-2019] Max samples per context for Histogram, Distribution and Timing metrics (Experimental Feature)

andrewqian2001datadog:

### What does this PR do?

This experimental feature allows the user to limit the number of samples per context for histogram, distribution, and timing metrics.

This can be enabled with the statsd_max_samples_per_context flag. When enabled up to n samples will be kept in per context for Histogram, Distribution and Timing metrics when Aggregation is enabled. The default value is 0 which means no limit.

This is already implemented for the go client. [Go Client Docs](https://github.com/DataDog/datadog-go/tree/master?tab=readme-ov-file#maximum-samples-per-context) 

### Verification Process

For local testing, follow steps [here](https://github.com/DataDog/datadogpy/blob/8391471502d84c7c97d38cf6f8ba427a631e480e/doc/dev/testing.md) to set up local testing for the python client

Replace `testapp/main.py` with 
```
from datadog import initialize, statsd
import time


options = {
    "statsd_host": "127.0.0.1",
    "statsd_port": 8125,
    "statsd_disable_buffering" : True,
    "statsd_disable_aggregation" : False,
    "statsd_max_samples_per_context" : 1,
    "statsd_aggregation_flush_interval" : 15
}

initialize(**options)

x = 0
name = "andrew_q_maxsample2"
sleepTime = 3
while(1):
  print("-------------------------------------------------")
  print("running :)", x)
  statsd.histogram(name, 1)
  time.sleep(sleepTime)
  x += 1
```

### DIFF
--- a/datadog/dogstatsd/aggregator.py
+++ b/datadog/dogstatsd/aggregator.py
@@ -40,6 +40,7 @@ class Aggregator(object):
                 self.metrics_map[metric_type] = {}
             for metric in current_metrics.values():
                 metrics.extend(metric.get_data() if isinstance(metric, SetMetric) else [metric])
+
         return metrics
 
     def set_max_samples_per_context(self, max_samples_per_context=0):
--- a/datadog/dogstatsd/aggregator.py
+++ b/datadog/dogstatsd/aggregator.py
@@ -40,7 +40,7 @@ class Aggregator(object):
             for metric in current_metrics.values():
                 metrics.extend(metric.get_data() if isinstance(metric, SetMetric) else [metric])
         return metrics
-    
+
     def set_max_samples_per_context(self, max_samples_per_context=0):
         self.max_samples_per_context = max_samples_per_context
 
--- a/datadog/dogstatsd/aggregator.py
+++ b/datadog/dogstatsd/aggregator.py
@@ -14,16 +14,16 @@ from datadog.dogstatsd.max_sample_metric_context import MaxSampleMetricContexts
 
 
 class Aggregator(object):
-    def __init__(self, max_metric_samples=0):
+    def __init__(self):
         self.metrics_map = {
             MetricType.COUNT: {},
             MetricType.GAUGE: {},
             MetricType.SET: {},
         }
         self.max_sample_metric_map = {
-            MetricType.HISTOGRAM: MaxSampleMetricContexts(HistogramMetric, max_metric_samples),
-            MetricType.DISTRIBUTION: MaxSampleMetricContexts(DistributionMetric, max_metric_samples),
-            MetricType.TIMING: MaxSampleMetricContexts(TimingMetric, max_metric_samples)
+            MetricType.HISTOGRAM: MaxSampleMetricContexts(HistogramMetric),
+            MetricType.DISTRIBUTION: MaxSampleMetricContexts(DistributionMetric),
+            MetricType.TIMING: MaxSampleMetricContexts(TimingMetric)
         }
         self._locks = {
             MetricType.COUNT: threading.RLock(),
@@ -40,6 +40,9 @@ class Aggregator(object):
             for metric in current_metrics.values():
                 metrics.extend(metric.get_data() if isinstance(metric, SetMetric) else [metric])
         return metrics
+    
+    def set_max_samples_per_context(self, max_samples_per_context=0):
+        self.max_samples_per_context = max_samples_per_context
 
     def flush_aggregated_sampled_metrics(self):
         metrics = []
@@ -100,6 +103,7 @@ class Aggregator(object):
     ):
         if rate is None:
             rate = 1
+        print("inside the add sample function", self.max_samples_per_context)
         context_key = self.get_context(name, tags)
         metric_context = self.max_sample_metric_map[metric_type]
-        return metric_context.sample(name, value, tags, rate, context_key)
+        return metric_context.sample(name, value, tags, rate, context_key, self.max_samples_per_context)
--- a/datadog/dogstatsd/aggregator.py
+++ b/datadog/dogstatsd/aggregator.py
@@ -4,13 +4,13 @@ from datadog.dogstatsd.metrics import (
     GaugeMetric,
     SetMetric,
 )
-from datadog.dogstatsd.buffered_metrics import (
+from datadog.dogstatsd.max_sample_metric import (
     HistogramMetric,
     DistributionMetric,
     TimingMetric
 )
 from datadog.dogstatsd.metric_types import MetricType
-from datadog.dogstatsd.buffered_metrics_context import BufferedMetricContexts
+from datadog.dogstatsd.max_sample_metric_context import MaxSampleMetricContexts
 
 
 class Aggregator(object):
@@ -20,10 +20,10 @@ class Aggregator(object):
             MetricType.GAUGE: {},
             MetricType.SET: {},
         }
-        self.buffered_metrics_map = {
-            MetricType.HISTOGRAM: BufferedMetricContexts(HistogramMetric, max_metric_samples),
-            MetricType.DISTRIBUTION: BufferedMetricContexts(DistributionMetric, max_metric_samples),
-            MetricType.TIMING: BufferedMetricContexts(TimingMetric, max_metric_samples)
+        self.max_sample_metric_map = {
+            MetricType.HISTOGRAM: MaxSampleMetricContexts(HistogramMetric, max_metric_samples),
+            MetricType.DISTRIBUTION: MaxSampleMetricContexts(DistributionMetric, max_metric_samples),
+            MetricType.TIMING: MaxSampleMetricContexts(TimingMetric, max_metric_samples)
         }
         self._locks = {
             MetricType.COUNT: threading.RLock(),
@@ -41,10 +41,10 @@ class Aggregator(object):
                 metrics.extend(metric.get_data() if isinstance(metric, SetMetric) else [metric])
         return metrics
 
-    def flush_aggregated_buffered_metrics(self):
+    def flush_aggregated_sampled_metrics(self):
         metrics = []
-        for metric_type in self.buffered_metrics_map.keys():
-            metric_context = self.buffered_metrics_map[metric_type]
+        for metric_type in self.max_sample_metric_map.keys():
+            metric_context = self.max_sample_metric_map[metric_type]
             for metricList in metric_context.flush():
                 metrics.extend(metricList)
         return metrics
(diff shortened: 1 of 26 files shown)


gh123man:
Do we need to merge `master` into this branch? I think that may fix the failing tests.

datadog/dogstatsd/buffered_metrics.py
```diff
@@ -0,0 +1,63 @@
+import random
+from datadog.dogstatsd.metric_types import MetricType
+from datadog.dogstatsd.metrics import MetricAggregator
+
+
+class BufferedMetric(object):
+    def __init__(self, name, tags, metric_type, specified_rate=1.0, max_metric_samples=0):
```
gh123man, line 7: `max_metric_samples` appears to always be hardcoded to `0`. 
In the go client, I believe this is user configurable. Should we do the same in the python client?

datadog/dogstatsd/buffered_metrics.py
```diff
@@ -0,0 +1,63 @@
...
+        self.data = []
+        self.stored_metric_samples = 0
+        self.total_metric_samples = 0
+
+    def sample(self, value):
+        self.data.append(value)
+        self.stored_metric_samples += 1
+        self.total_metric_samples += 1
+
+    def maybe_keep_sample(self, value):
```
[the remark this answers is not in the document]
gh123man: [Also this comment thread is interesting](https://github.com/DataDog/datadog-go/pull/283/files#r1352035189) regarding how sampling interacts with aggregation/extended aggregation. I think we would want to ensure we match this behavior.

datadog/dogstatsd/base.py
```diff
@@ -826,6 +828,13 @@ def flush_aggregated_metrics(self):
         for m in metrics:
             self._report(m.name, m.metric_type, m.value, m.tags, m.rate, m.timestamp)
 
+        buffered_metrics = self.aggregator.flush_aggregated_buffered_metrics()
+        send_method = self._send
+        self._send = self._send_to_buffer
```
gh123man, line 833: Why do we need to change the `_send` method when flushing aggregated metrics? 

The way I read this seems like we are buffering twice ex: histogram/distribution/timing in the buffered metric class, and then again by writing to the sender buffer via `_send_to_buffer`. 

Also - with aggregation enabled, `_start_flush_thread` is given the `flush_aggregated_metrics` method, so in this case will `flush_buffered_metrics` ever be called by the flush thread? 

Please correct me if I am wrong/misunderstanding.

gh123man (changes requested):
Left some comments/questions.

In the go client, extended aggregation is behind a config flag, and disabled by default. 
I think we should do the same here. 

However first let's confirm the buffering behavior. I wonder if it would be simpler and have similar results if we enabled buffering and aggregation to work together (Or open question: Is the old buffering implementation still relevant if we keep this one instead?)

datadog/dogstatsd/aggregator.py
```diff
@@ -30,6 +44,15 @@ def flush_aggregated_metrics(self):
                 metrics.extend(metric.get_data() if isinstance(metric, SetMetric) else [metric])
         return metrics
 
+    def flush_aggregated_buffered_metrics(self):
+        metrics = []
+        for metric_type in self.buffered_metrics_map.keys():
+            with self._locks[metric_type]:
```
ddrthall, line 50: Nit: is this lock necessary? It looks like the buffered_metrics_map is more or less constant, with changes performed in submethods (like flush) that lock themselves

datadog/dogstatsd/base.py
```diff
@@ -826,6 +828,13 @@ def flush_aggregated_metrics(self):
...
+        for m in buffered_metrics:
+            self._report(m.name, m.metric_type, m.value, m.tags, m.rate, m.timestamp)
```
ddrthall, line 835: _report will perform its own sampling, so putting buffered metrics in post-sampling looks like it will cause them to be sampled twice

datadog/dogstatsd/base.py
```diff
@@ -268,6 +274,10 @@ def __init__(
         depending on the connection type.
         :type max_buffer_len: integer
 
+        :param max_metric_samples: Maximum number of metric samples for buffered
+        metrics (Histogram, Distribution, Timing)
+        :type max_metric_samples: integer
```
ddrthall, line 279: This param seems to be an older version of the per_context one, I don't think it's actually defined

datadog/dogstatsd/base.py
```diff
@@ -826,6 +837,17 @@ def flush_aggregated_metrics(self):
         for m in metrics:
             self._report(m.name, m.metric_type, m.value, m.tags, m.rate, m.timestamp)
 
+        sampled_metrics = self.aggregator.flush_aggregated_sampled_metrics()
+        if not self._enabled:
+            return
+        for m in sampled_metrics:
```
ddrthall, line 843: Nit: While this is definitely fully functional, I would be a bit more comfortable if _report was modified to allow the option to not internally sample. That way we only have one location to worry about changes to "are we disabled", "how do we handle constant tags", "how do we handle telemetry", etc.

datadog/dogstatsd/max_sample_metric.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class MaxSampleMetric(object):
+    def __init__(self, name, tags, metric_type, specified_rate=1.0, max_metric_samples=0):
+        self.name = name
+        self.tags = tags
+        self.metric_type = metric_type
+        self.max_metric_samples = max_metric_samples
+        self.specified_rate = specified_rate
+        self.data = []
```
ddrthall, line 13: Likely makes sense to pre-allocate data in the case of non-zero max_metric_samples

datadog/dogstatsd/max_sample_metric.py
```diff
@@ -0,0 +1,63 @@
...
+        self.stored_metric_samples = 0
+        self.total_metric_samples = 0
+
+    def sample(self, value):
+        self.data.append(value)
+        self.stored_metric_samples += 1
+        self.total_metric_samples += 1
+
+    def maybe_keep_sample(self, value):
+        if self.max_metric_samples > 0:
```
ddrthall, line 23: All or most of this function will need lock protections, a number of internal variables like stored metrics and the data array are potentially being updated by multiple threads.

datadog/dogstatsd/max_sample_metric.py
```diff
@@ -0,0 +1,63 @@
...
+            if self.stored_metric_samples >= self.max_metric_samples:
+                i = random.randint(0, self.total_metric_samples - 1)
```
ddrthall, line 25: This looks like an off-by-one - if max_metric_samples is five, then the sixth sample should have a chance to not be persisted to the data array. It looks like we're persisting it 100% of the time here.

datadog/dogstatsd/max_sample_metric.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def skip_sample(self):
+        self.total_metric_samples += 1
+
+    def flush(self):
+        total_metric_samples = self.total_metric_samples
+        if self.specified_rate != 1.0:
+            rate = self.specified_rate
+        else:
+            rate = self.stored_metric_samples / total_metric_samples
```
ddrthall, line 43: With the current structure it is highly likely that we can always use this value, pending confirmation that we can safely increment all skips behind a lock without too heavy a performance hit.

datadog/dogstatsd/max_sample_metric_context.py
```diff
@@ -0,0 +1,42 @@
...
+        self.nb_context = 0
+        self.lock = Lock()
+        self.values = {}
+        self.max_sample_metric_type = max_sample_metric_type
+
+    def flush(self):
+        metrics = []
+        """Flush the metrics and reset the stored values."""
+        with self.lock:
+            copiedValues = self.values.copy()
```
ddrthall, line 16: It doesn't look to me like the copy and the clear are necessary - the subsequent assignment of values to an empty dict should be generally equivalent to the clear for our use case, and without a clear call there's no need to shallow copy the dict.

datadog/dogstatsd/max_sample_metric_context.py
```diff
@@ -0,0 +1,42 @@
...
+            self.values.clear()
+        self.values = {}
+        for _, metric in copiedValues.items():
+            metrics.append(metric.flush())
+
+        self.nb_context += len(copiedValues)
```
ddrthall, line 22: nb_context doesn't seem to be used

datadog/dogstatsd/max_sample_metric_context.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def sample(self, name, value, tags, rate, context_key, max_samples_per_context):
+        """Sample a metric and store it if it meets the criteria."""
+        keeping_sample = self.should_sample(rate)
+        with self.lock:
+            if context_key not in self.values:
+                # Create a new metric if it doesn't exist
+                self.values[context_key] = self.max_sample_metric_type(name, tags, rate, max_samples_per_context)
+            metric = self.values[context_key]
+        if keeping_sample:
```
ddrthall, line 33: It looks like we're likely racing the flush to persist this correctly. Operating flow: 
Thread in sample pulls a metric from the values dict and gets pre-empted after dropping the metric_context's lock. 
Separate thread enters flush, pulls the metric from values and removes further reference to it, then flushes the metric.
Original thread wakes back up and continues operating on the metric object that it had pulled, unaware that the metric had already been flushed and any additional samples added will be ignored.

datadog/dogstatsd/max_sample_metric.py
```diff
@@ -0,0 +1,63 @@
...
+            self.sample(value)
+
+    def skip_sample(self):
+        self.total_metric_samples += 1
+
+    def flush(self):
+        with self.lock:
+            values = [None] * self.stored_metric_samples
+            for i in range(self.stored_metric_samples):
+                values[i] = MetricAggregator(self.name, self.tags, self.specified_rate, self.metric_type, self.data[i])
```
ddrthall, line 47: I fairly confident that we want the `rate = self.stored_metric_samples / total_metric_samples` value here, since that is the actual real rate. Specified rate is used in the go variant as a workaround to not reliably being able to determine the real rate (go code has issues guaranteeing updates to its own version of total_metric_samples)

datadog/dogstatsd/max_sample_metric.py
```diff
@@ -0,0 +1,63 @@
...
+                    self.data[i] = value
+        else:
+            self.sample(value)
+
+    def skip_sample(self):
+        self.total_metric_samples += 1
+
+    def flush(self):
+        with self.lock:
+            values = [None] * self.stored_metric_samples
```
ddrthall, line 45: Nit: Python list comprehension, e.g. some_list = [ x for i in y], is actually faster than a pre-allocate, which itself is faster than an empty initialization and append loop. I would be rather surprised if any of the variants actually have a performance impact for a flush operation, but wanted to let you know that what you originally had here was actually optimal.

datadog/dogstatsd/max_sample_metric.py
ddrthall: Updated functionality looks safe to me.
Addendum nit: since this method requires a lock but relies on the caller to actually utilize said lock, it is common practice to append the work unsafe to the method name.

datadog/dogstatsd/base.py
```diff
@@ -960,7 +971,12 @@ def histogram(
         >>> statsd.histogram("uploaded.file.size", 1445)
         >>> statsd.histogram("album.photo.count", 26, tags=["gender:female"])
         """
-        self._report(metric, "h", value, tags, sample_rate)
+        if not self._disable_aggregation and self.aggregator.max_samples_per_context != 0:
+            print("Aggregated histogram")
```
ddrthall, line 975: Reminder to remove the prints here and below

ddrthall (approved):

ddrthall (approved):

skarimo (approved):

gh123man (approved):

PR #883, opened 2025-01-14
Add Cardinality common field

wdhif:
### What does this PR do?

Implement the `card:` common field to DogStatsD Datagram spec. This new field is used to override the Agent cardinality.

### Description of the Change

Adds a new cardinality setting that can be either:

- Defined globally as an option
- Override on a per-metrics basis

This settings is then send as the `card:` common field.

### Possible Drawbacks

None

### Verification Process

Unit tests have been implemented to test the feature. QA has been done manually and can be found in https://github.com/DataDog/datadog-agent/pull/32917.

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -47,6 +47,7 @@ def initialize(
     statsd_constant_tags=None,  # type: Optional[List[str]]
     return_raw_response=False,  # type: bool
     hostname_from_config=True,  # type: bool
+    cardinality=None,  # type: Optional[str]
     **kwargs  # type: Any
 ):
     # type: (...) -> None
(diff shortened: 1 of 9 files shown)


tobz:
What happens when this is used with an Agent version that doesn't have support for the structured `card` field? 🤔

datadog/__init__.py
```diff
@@ -117,6 +118,12 @@ def initialize(
 
     :param hostname_from_config: Set the hostname from the Datadog agent config (agent 5). Will be deprecated
     :type hostname_from_config: boolean
+
+    :param cardinality: Set the global cardinality for all metrics. \
+        Possible values are "none", "low", "orchestrator" and "high".
+        Can also be set via the DATADOG_CARDINALITY or DD_CARDINALITY environment variables.
```
carlosroman, line 124: Should we link to possible values to the [tag cardinality section](https://docs.datadoghq.com/getting_started/tagging/assigning_tags/?tab=containerizedenvironments#environment-variables) in our docs with something like to "to see all possible values check our docs"

wdhif: Discussed together as it would be better to also include documentation in case of value changes in the future.

wdhif: Sadly adding the link would break the linter.

carlosroman (approved):
Works as expected.

PR #915, opened 2026-03-06
refactor: cache constant tags

KowalskiThomas:
# What is this PR?

This PR updates the way we handle constant tags to improve the overall performance of reporting metrics.
Previously, we would compute the string of user tags + constant tags at every `report` call; now we instead
cache the constant tags part and only compute the string for user tags and the sum of both.

When the `DogStatsD` instance has even a few constant tags, this can make a significant difference in how
many allocations and processing we have to do.

## In practice

- `constant_tags` is now a `@property` with a setter that pre-computes `_constant_tags_str` (the normalised string)
- `_report` does not call `_add_constant_tags` anymore for metrics.
- `_add_constant_tags` is kept for `event` and `service_check` paths, but updated to use `self._constant_tags`.

## The gains

- `_add_constant_tags` list concatenation — previously allocated a new `tags + self.constant_tags` list on every call; this is gone.
- `normalize_tags` ran regex substitution over every tag on every call.  Now constant tags are normalised once at init.
- `",".join(...)` ran over all tags every call. Now constant tags are pre-joined; only user tags need joining at call time.

In order to maintain existing behaviour when using `statsd.constant_tags.<mutating_method>` (e.g. `clear` or `append`), I had to add a `TagList` class that is a very basic subclass of `list` and that updates the cached value when it is mutated.   
The alternative would have been to make it immutable (e.g. return a tuple) but this would have broken potential existing usages, which doesn't sound worth it.

## Benchmark results

**Setup:** 5M `statsd.increment()` calls per case (with buffering enabled).  
The baseline is the code before my changes, optimised adds the constant tags cache.

| Case | Baseline (ops/sec) | Optimised (ops/sec) | Delta |                                                                                                      
|---|---|---|---|
| 5 const + 3 user tags | 1,840k | 2,504k | **+36.1%** |                                                                                                         
| 5 const, no user tags | 1,905k | 3,985k | **+109.2%** |                                                                                                      
| 3 user, no const tags | 2,433k | 2,593k | **+6.6%** |
| No tags | 3,831k | 4,603k | **+20.1%** |
| 10 const + 2 user tags | 1,700k | 2,629k | **+54.6%** |

The win scales with the number of constant tags and to a lower extent user tags.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -19,7 +19,9 @@ import threading
 import time
 from threading import Lock, RLock
 import weakref
-from typing import TYPE_CHECKING
+
+if sys.version_info[:2] >= (3, 5):
+    from typing import TYPE_CHECKING  # noqa: F401
 
 try:
     import queue
@@ -29,7 +31,13 @@ except ImportError:
 
 
 # pylint: disable=unused-import
-from typing import Any, Optional, List, Text, Type, Union
+if sys.version_info[:2] >= (3, 5):
+    from typing import Any, Optional, List, Text, Type, Union, Iterable, Callable, overload  # noqa: F401
+
+try:
+    from typing import SupportsIndex
+except ImportError:
+    SupportsIndex = int  # type: ignore[assignment,misc]
 # pylint: enable=unused-import
 
 # Datadog libraries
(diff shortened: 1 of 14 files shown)


datadog/dogstatsd/base.py
```diff
@@ -1199,16 +1261,26 @@ def close_socket(self):
     def _serialize_metric(
         self, metric, metric_type, value, tags, sample_rate=1, timestamp=0, cardinality=None
     ):
+        # Build the combined tag string from user tags + cached constant tags
+        if tags:
+            user_tags_str = ",".join(normalize_tags(tags))
+            if self._constant_tags_str:
+                tag_str = user_tags_str + "," + self._constant_tags_str
```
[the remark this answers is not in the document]
KowalskiThomas: Great suggestion, this was actually even better than what I had initially proposed. I updated the code (and the PR description).

datadog/dogstatsd/base.py
```diff
@@ -1206,9 +1278,15 @@ def _serialize_metric(
             parts.append("|@")
             parts.append(text(sample_rate))
 
-        if tags:
+        if tags or self._constant_tags_str:
```
StephenWakely: This needs to lock `self._config_lock`.

StephenWakely (changes requested):
There's a couple of places where constant tags isn't thread safe. I don't think this was introduced by this PR, but I think it would be worth fixing. Would it be possible to do? Does it affect the benchmarks in any way?

datadog/dogstatsd/base.py
```diff
@@ -1565,12 +1641,33 @@ def service_check(
...
+
+    @constant_tags.setter
+    def constant_tags(self, value):
+        with self._config_lock:
+            self._constant_tags = TagList(value or [], on_change=self._rebuild_constant_tags_str)
+            self._rebuild_constant_tags_str()
+
     def _add_constant_tags(self, tags):
-        if self.constant_tags:
+        if self._constant_tags:
```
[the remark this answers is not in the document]
KowalskiThomas: Added this -- I don't think this should have any impact on the benchmark results as this is not on the hot path.

datadog/dogstatsd/base.py
KowalskiThomas: I don't think it does if we capture it first?

Like instead of doing `if tags or self._constant_tags_str`, we do

````py
constant_tags_str = self._constant_tags_str
if tags or constant_tags_str:
  # use constant_tags_str, not self._constant_tags_str
````

What do you think? Running benchmark shows acquiring the lock here is somewhat costly.  
The assignment I do in the proposed code is thread-safe in Python so while it may differ slightly from the-same-code-but-with-a-lock, but at least doing the capture instead of re-reading the attribute avoids the TOCTOU race.

StephenWakely: Ok, yes good call. The lock isn't needed. I hadn't realised that these reads are atomic in CPython. I don't think capturing it first is necessary.

StephenWakely (approved):

datadog/dogstatsd/base.py
KowalskiThomas: I do believe we need capture if we want to be consistent between the conditions and usage, right?

StephenWakely: Oh, yes true, it's used afterwards.

walidcavelius (approved):

PR #916, opened 2026-03-06
fix: pass missing parameter in `MaxSampleMetricContexts`

KowalskiThomas:
### What does this PR do?

When creating a new metric context, the `rate` parameter is currently not passed to the metric constructor. I believe that means sampling rates for histogram/distribution/timing metrics are ignored. It also means that the parameters we did pass to the constructor were shifted by one position, making them do _not what we wanted_.

### Description of the Change

This adds forwarding of the `rate` parameter to the constructor we call, and adds a unit test to ensure forwarding is done properly. 

### Possible Drawbacks

_I don't see any._

### Reproducer

This is as simple as I could make it -- if we pass parameters properly, the "max samples per context" limiting should work.

```py
from unittest.mock import patch
from datadog.dogstatsd.max_sample_metric import HistogramMetric
from datadog.dogstatsd.max_sample_metric_context import MaxSampleMetricContexts

MAX_SAMPLES = 10

with patch("datadog.dogstatsd.max_sample_metric_context.random.random", return_value=0.0):
    ctx = MaxSampleMetricContexts(HistogramMetric)
    for _ in range(1000):
        ctx.sample("m", 1.0, ["t:v"], 0.5, "key", max_samples_per_context=MAX_SAMPLES)

metric = ctx.values["key"]
print(f"buffer size: {len(metric.data)} (expected: {MAX_SAMPLES})")
assert len(metric.data) == MAX_SAMPLES, f"BUG: buffer grew to {len(metric.data)}"
```

### Additional Notes

❓  I think this problem is one that would typically be detected and prevented by even basic static typing / static correctness checking. Has anyone ever looked into this, and would you be opposed to (me) adding some?  Not necessarily a full blown typing of everything, but at least some basics.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ concurrency:
 
 jobs:
   lint:
-    name: Lint datadopy files
+    name: Lint datadogpy files
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0


StephenWakely (approved):
Good catch! Thanks.

datadog/dogstatsd/max_sample_metric_context.py
```diff
@@ -25,7 +25,7 @@ def sample(self, name, value, tags, rate, context_key, max_samples_per_context,
             if context_key not in self.values:
                 # Create a new metric if it doesn't exist
                 self.values[context_key] = self.max_sample_metric_type(
-                    name, tags, max_samples_per_context, cardinality=cardinality
+                    name, tags, rate, max_samples_per_context, cardinality=cardinality
                 )
```
damfleu: Could we use keyword args here? It would eliminate this class of positional off-by-one bugs.

damfleu (dismissed):

PR #917, opened 2026-03-06
fix: restore `--pretty`/`--raw` support in `dogshell`

KowalskiThomas:
### What does this PR do?

This PR fixes the `format` check `"pretty"` in various `dogshell` functions. Previously, they didn't work because `format` wasn't set anywhere, so it just compared the builtin `format` function to the `"pretty"` string which was  never `True`, so it was dead code.  It now uses `args.format` which other functions already did.

### DIFF
--- a/datadog/dogshell/timeboard.py
+++ b/datadog/dogshell/timeboard.py
@@ -244,10 +244,11 @@ class TimeboardClient(object):
             report_warnings(res)
             report_errors(res)
 
-            print(json.dumps(res))
-
             if args.format == "pretty":
+                print(pretty_json(res))
                 print("Uploaded file {0} (dashboard {1})".format(f.name, dash_obj["id"]))
+            else:
+                print(json.dumps(res))
 
     @classmethod
     def _post(cls, args):


datadog/dogshell/search.py
```diff
@@ -27,17 +27,4 @@ def _query(cls, args):
...
-                        print("%s\t%s" % (facet, result))
-                    else:
-                        print("%s\t%s" % (" " * len(facet), result))
-        elif format == "raw":
-            print(json.dumps(res))
-        else:
-            for facet, results in list(res["results"].items()):
-                for result in results:
-                    print("%s\t%s" % (facet, result))
+        print(json.dumps(res))
```
damfleu: I think this one needs a bit more care. The `else` branch here isn't dead code, since `format` was never `pretty` or `raw`, the `else` branch is always executed and doesn't dump JSON, it prints the output in some tabular format. Your change actually changes the default output format.

The fix here is again to just set `format = args.format` at the top of `_query` and keep all three branches intact?

datadog/dogshell/monitor.py
```diff
@@ -424,20 +424,14 @@ def _unmute(cls, args):
         res = api.Monitor.unmute(args.monitor_id, scope=args.scope, all_scopes=args.all_scopes)
         report_warnings(res)
         report_errors(res)
-        if format == "pretty":
-            print(pretty_json(res))
-        else:
-            print(json.dumps(res))
+        print(json.dumps(res))
```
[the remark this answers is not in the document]
KowalskiThomas: Okay, sure, I can do that!

datadog/dogshell/search.py
KowalskiThomas: Should be fixed now!

PR #918, opened 2026-03-11
typing: fix existing typing issues

KowalskiThomas:
### What does this PR do?

This PR
- Updates `mypy` to use the latest version
- Adds `mypy` configuration to report untyped code by default, and adds exclusions to all existing folders (overall a no-op, but will allow folder-wise fixes in the future)
- Fixes errors reported by `mypy` after upgrading
  - Implicit optionals are forbidden
  - Properly handles typing for try-to-import missing modules; type is `Optional[ModuleType]`, defaults to `None` and we `try` to set it using `__import__`. Functionally it's the same as before, but it's correctly typed now.

**Note** we need `noqa: F401`'s because we don't have actual typing, we just have typing comments so technically those names are unused.

### DIFF
--- a/datadog/api/http_client.py
+++ b/datadog/api/http_client.py
@@ -32,10 +32,13 @@ try:
 except ImportError:
     pass
 
+urlfetch = None  # type: Optional[types.ModuleType]
+urlfetch_errors = None  # type: Optional[types.ModuleType]
 try:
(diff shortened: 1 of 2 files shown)


MintsInc (commented):
Just a few questions inlines.

PR #925, opened 2026-03-19
performance: ruff-based improvements

KowalskiThomas:
## What is this PR?

This PR makes a few code improvements based on `ruff check` reports.

### DIFF
--- a/datadog/threadstats/base.py
+++ b/datadog/threadstats/base.py
@@ -165,14 +165,14 @@ class ThreadStats(object):
 
     def stop(self):
         if not self._is_auto_flushing:
-            return True
+            return
 
         if self._flush_thread:
             self._flush_thread.end()
             self._is_auto_flushing = False
-            return True
+            return
 
-        return True
+        return
 
     def event(
         self,
--- a/datadog/threadstats/base.py
+++ b/datadog/threadstats/base.py
@@ -166,10 +166,13 @@ class ThreadStats(object):
     def stop(self):
         if not self._is_auto_flushing:
             return True
+
         if self._flush_thread:
             self._flush_thread.end()
             self._is_auto_flushing = False
             return True
+
+        return True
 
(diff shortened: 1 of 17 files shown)


datadog/threadstats/base.py
```diff
@@ -166,11 +166,14 @@ def start(
...
         if not self._is_auto_flushing:
             return True
+
         if self._flush_thread:
             self._flush_thread.end()
             self._is_auto_flushing = False
             return True
 
+        return True
+
```
damfleu, line 176: Every branch of this function now return `True`. I am not sure if callers check it, but that's a behavior change since it implicitly returned `None` before and that makes me wonder about the intent.

I'm not really sure what this return value brings anything at all here, and if it was supposed to mean something, we have probably lost that. Should we just return `None` in all cases?

atanzu: I believe it would be cleaner to not to return at all.

KowalskiThomas: Good point; updated now.

datadog/util/hostname.py
```diff
@@ -26,14 +26,12 @@
...
-            "localhost6.localdomain6",
-            "ip6-localhost",
-        ]
-    ):
+    if hostname.lower() in {
+        "localhost",
+        "localhost.localdomain",
+        "localhost6.localdomain6",
+        "ip6-localhost",
+    }:
```
atanzu, line 34: nit: since we are iterating over possible values, a simple list might suffice.

atanzu (approved):

datadog/util/hostname.py
KowalskiThomas: I think a `set` was chosen here because `in` for `set` is O(1) whereas for `list` it is O(n).

damfleu (approved):

PR #932, opened 2026-04-03
chore(dogstatsd): improvements

KowalskiThomas:
### What does this PR do?

Small code quality improvements in `dogstatsd`.

### DIFF
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -925,15 +925,15 @@ class DogStatsd(object):
     @classmethod
     def _ensure_min_send_buffer_size(cls, sock, min_size=MIN_SEND_BUFFER_SIZE):
         # type: (_Socket, int) -> None
-        # Increase the receiving buffer size where needed (e.g. MacOS has 4k RX
+        # Increase the send buffer size where needed (e.g. MacOS has 4k TX
         # buffers which is half of the max packet size that the client will send.
         if os.name == 'posix':
             try:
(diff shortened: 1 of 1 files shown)


datadog/dogstatsd/base.py
```diff
@@ -925,15 +925,15 @@ def set_socket_timeout(self, timeout):
...
         if os.name == 'posix':
             try:
-                recv_buff_size = sock.getsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF)
-                if recv_buff_size <= min_size:
+                send_buff_size = sock.getsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF)
+                if send_buff_size <= min_size:
                     sock.setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, min_size)
                     log.debug("Socket send buffer increased to %dkb", min_size / 1024)
-            finally:
+            except Exception:
```
vickenty: This now silently eats errors from setsockopt. This may have been the intent all along, but we should at least log something. However, given that exceptions raised here didn't bother anyone all these years, I'd just remove the try block altogether and keep the old raising behavior.

KowalskiThomas: Thank you, updated.

KowalskiThomas: Well actually _not updated_ because I lost my push permissions on the repo 🤔  Looking into this.

KowalskiThomas: Okay actually updated now! 🎉

Kaycell (approved):

PR #964, opened 2026-07-07
Retry DogStatsD UDS connect failures

rayz:
### What does this PR do?

Adds configurable retry handling when the DogStatsD client connects to a Unix Domain Socket and encounters a transient connection error.

The client retries UDS `connect()` failures for:

- `ENOENT`: the socket path does not exist yet, such as during Agent cold start.
- `ECONNREFUSED`: the socket exists but the Agent is not accepting connections, such as during an Agent restart.

Retries use exponential backoff and are bounded by the new `socket_connect_timeout` option. The timeout applies to the entire connection operation, including fallback between datagram and stream socket types.

`socket_timeout` remains responsible for normal socket operations after the connection succeeds.

`socket_connect_timeout` defaults to `0`, so retries are disabled by default and existing client behavior is preserved.

Inspired by the [datadog-go equivalent PR](https://github.com/DataDog/datadog-go/pull/323) 


datadog/dogstatsd/base.py
```diff
@@ -951,21 +954,43 @@ def _get_uds_socket(cls, socket_path, timeout):
...
+                    if (
+                        deadline is not None
+                        and getattr(e, "errno", None) in UDS_TRANSIENT_CONNECT_ERRORS
+                    ):
+                        remaining_time = max(0, deadline - time.time())
+                        sleep_time = min(backoff, remaining_time)
+                        if sleep_time > 0:
+                            time.sleep(sleep_time)
+                            backoff = min(backoff * 2, UDS_CONNECT_RETRY_MAX_BACKOFF)
+                            continue
```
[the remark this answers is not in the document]
rayz: Fixed in [fc0939d](https://github.com/DataDog/datadogpy/pull/964/commits/fc0939d195ff9a43257299bd326f24ed09cfd5b9)

datadog/dogstatsd/base.py
```diff
@@ -951,21 +954,47 @@ def _get_uds_socket(cls, socket_path, timeout):
...
+                deadline = time.time() + timeout
+
+            backoff = UDS_CONNECT_RETRY_INITIAL_BACKOFF
+
+            while deadline is None or time.time() < deadline:
+                sock = None
+                try:
+                    sock = socket.socket(socket.AF_UNIX, socket_kind)
+                    if deadline is not None:
+                        sock.settimeout(max(0, deadline - time.time()))
```
StephenWakely: Here we risk accidentally setting a timeout of 0 which would turn the socket non blocking. We then potentially cause errors to be raised the the client is not expecting because they have specified a timeout.

datadog/dogstatsd/base.py
```diff
@@ -951,21 +954,47 @@ def _get_uds_socket(cls, socket_path, timeout):
...
-                    last_error = e
-                    continue
-                raise e
+            deadline = None
+            if timeout and timeout > 0:
+                deadline = time.time() + timeout
+
+            backoff = UDS_CONNECT_RETRY_INITIAL_BACKOFF
+
+            while deadline is None or time.time() < deadline:
```
StephenWakely, line 1040: We can never attempt reconnecting beyond the time specified by the socket timeout. 

Would it make more sense to expose a new configuration option for connection timeout? Potentially in the face of agent restarts this could be 1 or more minutes before we can reconnect - but we probably wouldn't want to have that timeout for send errors.

StephenWakely (changes requested):
I think it would make more sense to set up a separate config option for this timeout, reusing the socket timeout for the connection timeout

datadog/dogstatsd/base.py
rayz: Added a new option in commit [78797a3](https://github.com/DataDog/datadogpy/pull/964/commits/78797a36d8c205bdbf0fae3485ccad1ed22f4f87)

datadog/dogstatsd/base.py
rayz: Good catch, fix in [0535140](https://github.com/DataDog/datadogpy/pull/964/commits/05351404635297081f88aa1951371c1f3807cd0e)

datadog/dogstatsd/base.py
```diff
@@ -946,26 +962,55 @@ def _get_uds_socket(cls, socket_path, timeout):
...
-                if sock is not None:
-                    sock.close()
-                log.debug("Failed to connect to %s with kind %s: %s", socket_path, sk_name, e)
-                if getattr(e, "errno", None) == errno.EPROTOTYPE:
-                    last_error = e
-                    continue
-                raise e
+            deadline = None
+            if connect_timeout and connect_timeout > 0:
+                deadline = time.time() + connect_timeout
```
StephenWakely: We should calculate this before the `for socket_kind` loop. Currently if the server crashes we will wait 2x the timeout as we try both socket kinds.

StephenWakely (commented):
Sorry, spotted one more change needed..

StephenWakely (dismissed):
Beautiful.

armcburney (dismissed):

StephenWakely (approved):

PR #870, opened 2024-11-18
Add DD_DOGSTATSD_URL support for unix and udp urls

ddrthall:
Adds support for the unix:// and udp:// variants of DD_DOGSTATSD_URL.

### Description of the Change
During dogstatsd creation, utilize the DD_DOGSTATSD_URL environment variable if the provided host and port are their default values and no socket_path has been provided. unix:// and udp:// paths are supported, but \\\\.\pipe\ paths (windows named pipes) are not yet implemented. 

### Verification Process
Form a Dogstatsd object with different values of DD_DOGSTATSD_URL. The following cases are important to verify:
* A non-default host, port or socket_path provided to the Dogstatsd constructor should take precedence over DD_DOGSTATS_URL.
* unix:// prefixed DOGSTATD_URLS should result in a socket_path and a None host/port pair
* udp:// prefixed DOGSTATSD_URLS should result in a valid host/port pair, with appropriate type safety around casting the port component from a string to an integer. 
* \\\\.\pipe\ prefixed DOGSTATD_URLS should provide an informative debug message and fallback to other connection identifications. 
* An unrecognized prefix for DOGSTATD_URL should fallback to other connection identifications

During fallback, DD_AGENT_HOST and DD_DOGSTATSD_PORT should be utilized if present. Otherwise, udp is selected with DEFAULT_HOST:DEFAULT_PORT. 

### Additional Notes
This PR is reliant on the functionality in [!869](https://github.com/DataDog/datadogpy/pull/869), with a comparison available [here](https://github.com/DataDog/datadogpy/compare/ryan.hall/add_uds_stream_support...ryan.hall/add_dogstatsd_url_support?expand=1)

### Review checklist (to be filled by reviewers)

- [ ] Feature or bug fix MUST have appropriate tests (unit, integration, etc...)
- [ ] PR title must be written as a CHANGELOG entry [(see why)](https://github.com/DataDog/datadogpy/blob/master/CONTRIBUTING.md#pull-request-title)
- [ ] Files changes must correspond to the primary purpose of the PR as described in the title (small unrelated changes should have their own PR)
- [ ] PR must have one `changelog/` label attached. If applicable it should have the `backward-incompatible` label attached.
- [ ] PR should not have `do-not-merge/` label attached.
- [ ] If Applicable, issue must have `kind/` and `severity/` labels attached at least.

PR #968, opened 2026-07-14
Add DD_DOGSTATSD_URL support for unix and udp urls

### LINKED ISSUE #805: Support the standard DD_DOGSTATSD_URL environment variable
**Note:** 

**Is your feature request related to a problem? Please describe.**

Currently the only way to configure a connection to a unix domain socket requires doing this in code for every application.

Some of the other Datadog dogstatsd libraries support the `DD_DOGSTATSD_URL` environment variable for setting either host/port or a unix domain socket. For example https://github.com/DataDog/dogstatsd-ruby gem does this. Also the Datadog Kubernetes operator also automatically adds the `DD_DOGSTATSD_URL` environment variable to a mounted socket into all pods.

**Describe the solution you'd like**

This library should support reading the `DD_DOGSTATSD_URL` environment variable to configure either host/port or unix domain sockets. This will allow configuration outside of code and be consistent with other dogstatsd libraries.

**Describe alternatives you've considered**

Continue setting this in code and therefore being inconsistent with other Datadog libraries.

**Additional context**

sortega:
### Description of the Change

Reads the standard `DD_DOGSTATSD_URL` environment variable to configure the DogStatsd connection, matching the other Datadog dogstatsd libraries and the Datadog Kubernetes operator (which mounts a UDS and sets `DD_DOGSTATSD_URL=unix:///var/run/datadog/dsd.socket`).

Supported schemes:
- `udp://host:port` → resolves to host/port
- `unix:///path/to/socket` → resolves to `socket_path` (the full url is stored; `_get_uds_socket` already strips the scheme, so `unixgram://`/`unixstream://` distinctions are preserved). Windows named pipes are not supported.

`host`/`port` are now `Optional` (`None` sentinels). This makes an explicit connection argument (e.g. `DogStatsd(host="localhost")`) distinguishable from the default, so it is never silently overridden by an environment variable. Precedence: **explicit arg > `DD_DOGSTATSD_URL` > `DD_AGENT_HOST`/`DD_DOGSTATSD_PORT` > hard-coded default.**

This revives #870 with the review feedback applied: reuses the existing `datadog.util.compat.urlparse` (py2/py3) shim instead of a duplicate import, drops the unused Windows-named-pipe constant, upgrades the unsupported-scheme / invalid-port messages to `warning` with the offending value, and uses `None` sentinels to fix the explicit-default-precedence concern.

### Verification Process

New unit tests in `tests/unit/dogstatsd/test_statsd.py` cover udp and unix urls, explicit-arg precedence (including the `DogStatsd(host="localhost")` regression case), unsupported schemes, a portless udp url, and url-vs-legacy-env precedence. `pytest tests/unit/dogstatsd/test_statsd.py`, `flake8`, and `mypy` all pass.

### Release Notes

`DogStatsd` now honors the standard `DD_DOGSTATSD_URL` environment variable (`udp://` and `unix://`).

Fixes #805

### DIFF
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -21,6 +21,7 @@ if sys.version_info[0] >= 3:
 # datadog
 from datadog import api
 from datadog.dogstatsd import DogStatsd, statsd  # noqa
+from datadog.dogstatsd.base import DEFAULT_HOST, DEFAULT_PORT
 from datadog.threadstats import ThreadStats, datadog_lambda_wrapper, lambda_metric  # noqa
 from datadog.util.compat import iteritems, NullHandler, text
 from datadog.util.hostname import get_hostname
@@ -152,6 +153,12 @@ def initialize(
         # UDS and the manual host/port override is silently ignored.
         if statsd_host or statsd_use_default_route or statsd_port:
             statsd.socket_path = None
+            # The client may have been UDS-configured (host/port None). Backfill any
+            # side not supplied here so the UDP socket has a valid host and port.
+            if statsd.host is None:
+                statsd.host = DEFAULT_HOST
+            if statsd.port is None:
+                statsd.port = DEFAULT_PORT
     statsd.close_socket()
     if statsd_namespace:
         statsd.namespace = text(statsd_namespace)
--- a/datadog/__init__.py
+++ b/datadog/__init__.py
@@ -147,6 +147,11 @@ def initialize(
             statsd.host = statsd.resolve_host(statsd_host, statsd_use_default_route)
         if statsd_port:
             statsd.port = int(statsd_port)
+        # Selecting a UDP destination must clear any socket path (e.g. one inherited
+        # from DD_DOGSTATSD_URL=unix://...), otherwise get_socket() keeps using the
+        # UDS and the manual host/port override is silently ignored.
+        if statsd_host or statsd_use_default_route or statsd_port:
+            statsd.socket_path = None
     statsd.close_socket()
     if statsd_namespace:
         statsd.namespace = text(statsd_namespace)
--- a/datadog/dogstatsd/base.py
+++ b/datadog/dogstatsd/base.py
@@ -32,7 +32,7 @@ except ImportError:
 
 # pylint: disable=unused-import
 if sys.version_info[:2] >= (3, 5):
-    from typing import Any, Optional, List, Text, Type, Union, Iterable, Callable, overload  # noqa: F401
+    from typing import Any, Optional, List, Text, Tuple, Type, Union, Iterable, Callable, overload  # noqa: F401
 
 try:
     from typing import SupportsIndex
@@ -49,7 +49,7 @@ from datadog.dogstatsd.context import (
 )
 from datadog.dogstatsd.route import get_default_route
 from datadog.dogstatsd.container import Cgroup
-from datadog.util.compat import text
+from datadog.util.compat import text, urlparse
 from datadog.util.format import normalize_tags, validate_cardinality
 from datadog.version import __version__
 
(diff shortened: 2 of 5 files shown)


datadog/dogstatsd/base.py
```diff
@@ -730,6 +733,65 @@ def enable_telemetry(self):
...
+        without overriding values already provided explicitly to the constructor.
+
+        DD_DOGSTATSD_URL takes precedence over DD_AGENT_HOST / DD_DOGSTATSD_PORT and
+        is only honored when no connection information was passed explicitly. It
+        supports 'udp://host:port' and 'unix:///path/to/socket' URLs.
+        """
+        dogstatsd_url = os.environ.get("DD_DOGSTATSD_URL")
+        if host is None and port is None and socket_path is None and dogstatsd_url is not None:
+            parsed = urlparse(dogstatsd_url)
+            if parsed.scheme == "unix":
```
StephenWakely: We should also check for `unixstream` and `unixgram` schemes here.

datadog/dogstatsd/base.py
```diff
@@ -730,6 +733,65 @@ def enable_telemetry(self):
...
+                # The scheme is stripped later by _get_uds_socket, so the full URL
+                # is stored as the socket path (this also preserves unixgram/unixstream).
+                log.debug("DD_DOGSTATSD_URL %r resolved to a unix socket path.", dogstatsd_url)
+                return None, None, dogstatsd_url
```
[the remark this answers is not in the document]
StephenWakely: Looks like this is legit, can we fix this too.

StephenWakely (changes requested):

datadog/dogstatsd/base.py
sortega: Done in b915de1 — now matches `unix`, `unixstream`, and `unixgram` (and the unsupported-scheme warning lists all four). Added test coverage for the two new schemes.

datadog/dogstatsd/base.py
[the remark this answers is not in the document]
sortega: Fixed in b915de1. `initialize()` now clears `statsd.socket_path` whenever a UDP destination (`statsd_host` / `statsd_port` / `statsd_use_default_route`) is selected, so a `socket_path` inherited from `DD_DOGSTATSD_URL=unix://...` no longer wins in `get_socket()` and the manual host/port override takes effect. Added `test_initialization_udp_clears_socket_path` to cover it.

StephenWakely (dismissed):

datadog/__init__.py
```diff
@@ -147,6 +147,11 @@ def initialize(
             statsd.host = statsd.resolve_host(statsd_host, statsd_use_default_route)
         if statsd_port:
             statsd.port = int(statsd_port)
+        # Selecting a UDP destination must clear any socket path (e.g. one inherited
+        # from DD_DOGSTATSD_URL=unix://...), otherwise get_socket() keeps using the
+        # UDS and the manual host/port override is silently ignored.
+        if statsd_host or statsd_use_default_route or statsd_port:
+            statsd.socket_path = None
```
damfleu, line 155: Question about a possible edge case here: if a process starts with `DD_DOGSTATSD_URL=<some_url>`, the module-level `statsd` singleton gets constructed once with `host=None, port=None, socket_path=<some_url>`, right? If the code later calls `initialize(statsd_host="some_host")` **without** passing `statsd_port`, this code here will clear `socket_path` and set `host` but `port` will be left untouched, so I think it would stay `None`? How is that handled in the downstream code?

Should we reset the `host` and `port` to their `DEFAULT_` value here if they're not set?

sortega: Yes, we should. Good catch.

StephenWakely (approved):

damfleu (approved):

PR #979, opened 2026-08-10
Retry the connection if a socket error is because the server went away.

StephenWakely:
### What does this PR do?

Retries reconnecting and resending a payload on send errors socket error.

### Description of the Change

On transient send errors  - `errno.ECONNREFUSED`, `errno.ECONNRESET`, `errno.ENOTCONN`, `errno.EPIPE`, `errno.ENOENT`, close the socket and attempt  to resend. We only attempt resending once to avoid getting stuck in a loop if the agent doesn't come back beyond the `socket_connect_timeout`. 

### Review checklist (to be filled by reviewers)

- [ ] Feature or bug fix MUST have appropriate tests (unit, integration, etc...)
- [ ] PR title must be written as a CHANGELOG entry [(see why)](https://github.com/DataDog/datadogpy/blob/master/CONTRIBUTING.md#pull-request-title)
- [ ] Files changes must correspond to the primary purpose of the PR as described in the title (small unrelated changes should have their own PR)
- [ ] PR must have one `changelog/` label attached. If applicable it should have the `backward-incompatible` label attached.
- [ ] PR should not have `do-not-merge/` label attached.
- [ ] If Applicable, issue must have `kind/` and `severity/` labels attached at least.

PR #982, opened 2026-08-26
Release 0.54.0

StephenWakely:
### What does this PR do?

Prepares the `0.54.0` release: adds the CHANGELOG entry and bumps `datadog/version.py`.

### Description of the Change

Follows [RELEASING.md](https://github.com/DataDog/datadogpy/blob/master/RELEASING.md):

* `CHANGELOG.md` — new `v0.54.0` section.
* `datadog/version.py` — `0.53.0` -> `0.54.0`.

One PR has merged to `master` since `v0.53.0`:

* [#979](https://github.com/DataDog/datadogpy/pull/979) — retry sending DogStatsD packets over UDS when the connection drops (`changelog/Changed`).

Minor bump rather than a patch because #979 is labelled `changelog/Changed`: it widens the meaning of the existing `socket_connect_timeout` option, which now also bounds send-retry rather than connect only. The behaviour is opt-in — `socket_connect_timeout` defaults to `0`, which disables retries — so users who do not set it are unaffected.

### Verification Process

* `CHANGELOG.md` entry follows the existing format and links to #979.
* `python setup.py --version` reports `0.54.0`; `from datadog.version import __version__` imports cleanly.
* No other file references the old version string.
* Note: `ddev` was not available locally, so the changelog entry was written by hand in the established format rather than generated via `ddev release changelog`.

### Additional Notes

Not to be merged until reviewed. Per RELEASING.md, after this merges: create the GitHub release/tag, confirm the PyPI publish action, then open a follow-up PR bumping `datadog/version.py` to a dev version (e.g. `0.54.1.dev`) — note that step appears to have been skipped after `v0.53.0`, since `master` was left at a flat `0.53.0`.

### Release Notes

* [Changed] Retry sending DogStatsD packets over UDS when the connection drops (opt-in via `socket_connect_timeout`). See [#979](https://github.com/DataDog/datadogpy/pull/979).


StephenWakely:
There is an issue with this release, with a long `connect_timeout` and a large queue size, on shutting down the library if the agent isn't running it will wait for `connect_timeout` seconds for every item in the queue - which could be a very long time! We'll need to fix this before creating a new release.

PR #985, opened 2026-09-03
Migrate integration tests from static API key secret to dd-sts

chouetz:
https://datadoghq.atlassian.net/browse/ACIX-1817
## Summary

- Replaces `DD_TEST_CLIENT_API_KEY` and `DD_TEST_CLIENT_APP_KEY` secrets with short-lived credentials fetched via [`dd-sts-action`](https://github.com/DataDog/dd-sts-action)
- Adds `id-token: write` permission required for OIDC token federation
- Switches `DD_TEST_CLIENT_USER` from a secret to a repo variable (`vars.DD_TEST_CLIENT_USER`)

This depends on the dd-source policy `datadogpy-integration-tests` being merged and deployed before this workflow is enabled.

## Test plan

- [x] Confirm dd-source policy PR is merged and deployed see https://github.com/ddoghq/dd-source/pull/62067
- [x] Set `DD_TEST_CLIENT_USER` as a repo variable in GitHub settings
- [ ] Remove `DD_TEST_CLIENT_API_KEY` and `DD_TEST_CLIENT_APP_KEY` secrets from repo settings once confirmed working
- [x] Trigger the workflow by adding the `ci/integrations` label to a PR and verify credentials are fetched successfully

replaces #981

### DIFF
--- a/.github/workflows/test_integration.yml
+++ b/.github/workflows/test_integration.yml
@@ -2,6 +2,7 @@ name: Run Integration Tests
 
 permissions:
   contents: read
+  id-token: write # Needed to federate tokens with dd-sts
 
 on: # yamllint disable-line rule:truthy
   pull_request:
@@ -25,6 +26,12 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
+      - name: Get Datadog credentials
+        id: dd-sts
+        uses: DataDog/dd-sts-action@2e8187910199bd93129520183c093e19aa585c75 # v1.0.0
+        with:
+          policy: datadogpy-integration-tests
+
       - name: Set up Python 3.9
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test_integration.yml
```diff
@@ -25,6 +26,12 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
+      - name: Get Datadog credentials
+        id: dd-sts
+        uses: DataDog/dd-sts-action@2e8187910199bd93129520183c093e19aa585c75 # v1.0.0
+        with:
+          policy: datadogpy-integration-tests
```
skarimo, line 33: These tests run against our custom org `321813`. Please don't merge this PR as the integration tests can be quite destructive at times and we don't want to inadvertently corrupt resources org 2.

skarimo (changes requested):

.github/workflows/test_integration.yml
chouetz: @skarimo currently the token has not any write permissions. I suppose we can change the `org` target on the chainguard

skarimo: IIUC, the new dynamic token is generated for the org 2 in us1. The existing keys `secrets.DD_TEST_CLIENT_API_KEY` and `secrets.DD_TEST_CLIENT_APP_KEY` are for the org `321813`. This is what I was referring to. We don't want to execute the integration test against org 2

chouetz: An update [here](https://github.com/ddoghq/dd-source/blob/main/domains/seceng/sit/apps/apis/dd-sts/config/policies/us1.ddbuild.io/datadogpy-integration-tests.yaml#L24) should do the trick, correct?

skarimo: Assuming dd-sts supports a non-standard monitoring org such as 321813, then yes. But from what I recall, they do not support custom org yet.

Issue #834, opened 2024-06-14
Possible bug with dogstatsd deadlock


mazad01:
**Describe the bug**
Code is hanging on datadog's post_fork() function. Possibly https://github.com/DataDog/datadogpy/blob/master/datadog/dogstatsd/base.py#L976 ?

**Expected behavior**
Dogstatsd lib references not deadlocking

**Environment and Versions (please complete the following information):**
dogstatsd 0.49.1

**Additional context**
From our developers: the parent code is trying to spawn a subprocess, but it hangs. Output of pyspy:

```
Thread 1733153 (idle): "Thread-52 (process_items)"
    close_socket (datadog/dogstatsd/base.py:933)
    post_fork (datadog/dogstatsd/base.py:1369)
    post_fork (datadog/dogstatsd/base.py:120)
    _execute_child (subprocess.py:1796)
    __init__ (subprocess.py:971)
...
```

Output of strace

```
strace: Process 1733153 attached
futex(0x5644f988f040, FUTEX_WAIT_BITSET_PRIVATE|FUTEX_CLOCK_REALTIME, 0, NULL, FUTEX_BITSET_MATCH_ANY
```

mazad01:
Not stale, still waiting for resolution. Thanks

anupamdialpad:
What are the side effects of setting `DD_DOGSTATSD_DISABLE_FORK_SUPPORT`?

PR #989, opened 2026-09-21
Suppress logging from post_fork_child/post_fork_parent

seanmuth:
### What does this PR do?

Related to #834, a different specific manifestation of the same underlying class — a thread that hangs somewhere reachable from `post_fork_child`/`post_fork_parent`, the callbacks registered via `os.register_at_fork()`.

`_start_flush_thread` and `_start_sender_thread` both call `log.debug(...)` on every branch they can take, and both run from `post_fork_child`/`post_fork_parent`. `logging.Logger.debug()` is not safe to call there: it can block acquiring a `StreamHandler`'s own lock, and if some other thread held that lock at the exact instant of `fork()`, it's left permanently locked in the child — no thread survives fork to release it. Confirmed live: a process using this client hung permanently inside `post_fork_child` → `_start_flush_thread` → `log.debug`, even with buffering, aggregation, and the background sender all disabled — the "everything is disabled" branch still logs unconditionally today, so there was no config that avoided this path entirely.

### Description of the Change

Add a `quiet=False` parameter to both `_start_flush_thread` and `_start_sender_thread`, gating every `log.debug(...)` call inside each behind `if not quiet:`. `post_fork_child`/`post_fork_parent` now call both with `quiet=True`. Every other caller is unaffected — same logging as before.

### Alternate Designs

Could have just deleted the log lines outright — #817 removed a different post-fork log call for being "misleading" for a similar reason. Went with `quiet` instead so normal (non-fork) construction/config-change logging is untouched; only the specific atfork-reachable path goes quiet.

### Possible Drawbacks

Anyone relying on these specific debug-level log lines firing after a fork loses that signal. These aren't documented as a stable interface and are debug-level only, so this seems low-risk.

### Verification Process

- Added `test_post_fork_does_not_log` (parametrized across the same buffering/sender config combinations as the existing fork tests) asserting `log.debug` is never called during `post_fork_child`/`post_fork_parent`. Confirmed it fails against the pre-fix code with the exact two log calls this PR removes from that path, and passes with the fix.
- Full existing suite: `tests/unit/dogstatsd/test_statsd.py` (134 passed, 1 skipped, unrelated), `tests/integration/dogstatsd/test_statsd_fork.py` (16 passed, including the new test), `tests/integration/dogstatsd/test_statsd_sender.py` (79 passed) — all green, no regressions.
- `mypy datadog/dogstatsd/base.py` clean.

### Additional Notes

Generated with AI assistance (Claude Sonnet 5); I reviewed and tested the change before opening this PR.


StephenWakely (approved):

nogates (approved):

StephenWakely (changes requested):
Can you fix the [copilot suggestion](https://github.com/DataDog/datadogpy/pull/989#discussion_r4094757788) as well. We'll need to pass `quiet` to `close_socket` as well..

